PR #1, opened 2019-07-24
Redis features, Logger context fields

netsi:
- Incr IncrBy
    - Decr DecrBy
    - Exists
    - Expire


netsi:
It is possible now to store Fields in the context that will be used by the logger

PR #10, opened 2019-08-27
Add handler for multiple bindings;

rchernanko:


pkg/apiserver/handler.go
```diff
@@ -110,6 +115,10 @@ func CreateJsonHandler(handler HandlerWithInput) gin.HandlerFunc {
 	return handleWithInput(handler, binding.JSON, defaultErrorHandler)
 }
 
+func CreateJsonAndQueryHandler(handler HandlerWithMultipleBindings) gin.HandlerFunc {
```
applike-ss: since you are just wrapping `handleWithChainedInputs`, i would suggest to name the function `CreateChainedHandler`

pkg/apiserver/handler.go
```diff
@@ -165,6 +174,25 @@ func handleWithInput(handler HandlerWithInput, binding binding.Binding, errHandl
 	}
 }
 
+func handleWithChainedInputs(handler HandlerWithMultipleBindings, errHandler ErrorHandler) gin.HandlerFunc {
```
applike-ss: does it make more sense to call this `handleWithMultipleBindings`?

applike-ss:
looks good, please rebase and squash your commits so we can merge it as one commit :+1:

PR #12, opened 2019-08-29
Authenticate by query param key;

rchernanko:


pkg/apiserver/auth/config_key.go
```diff
@@ -91,10 +89,20 @@ func (a *configKeyAuthenticator) IsValid(ginCtx *gin.Context) (bool, error) {
 	return true, nil
 }
 
-func (a *configKeyAuthenticator) getApiKeyFromSource(ginCtx *gin.Context) string {
-	if a.source == SourceQueryParams {
-		return ginCtx.Query(ByApiKey)
+func QueryProvider(queryName string) ApiKeyProvider {
```
applike-ss: Can we find a better naming than `QueryProvider`? I'm not sure myself what speaks best for what it does, but this is too generic i think.

pkg/apiserver/auth/config_key.go
```diff
@@ -91,10 +89,20 @@ func (a *configKeyAuthenticator) IsValid(ginCtx *gin.Context) (bool, error) {
...
+	return func(ginCtx *gin.Context) string {
+		return ginCtx.Query(queryName)
+	}
+}
+
+func HeaderProvider(headerName string) ApiKeyProvider {
```
applike-ss: same as above

pkg/apiserver/auth/config_key.go
```diff
@@ -91,10 +89,20 @@ func (a *configKeyAuthenticator) IsValid(ginCtx *gin.Context) (bool, error) {
...
+	return func(ginCtx *gin.Context) string {
+		return ginCtx.GetHeader(headerName)
 	}
+}
 
-	return ginCtx.GetHeader(headerApiKey)
+func ParamProvider(paramName string) ApiKeyProvider {
```
applike-ss: same as above

PR #21, opened 2019-09-12
add stages to pipeline, let nested config be overwritten by env variable

Teaxtor:
So far, a pipeline only allows a single processing step. When in need of any kind of pre- or post-processing, this approach fails to result in clean code. Thus, I added the possibility to have multiple processing stages.

Furthermore, during testing the configurable input for pipelines, I noticed that the config loading from env variables is broken for nested structures. I applied a patch for that one.


pkg/sqs/service_test.go
```diff
@@ -92,3 +95,63 @@ func TestService_CreateQueue(t *testing.T) {
...
+	client := new(mocks.SQSAPI)
+	manager := new(cloudMocks.ResourceManager)
+
+	manager.On("GetResources", mock.AnythingOfType("cloud.ResourceFilter")).Return([]string{
+		"arn:applike-test-gosoline-sqsq-my-queue",
+	}, nil)
+
+	expected := sqs.Properties{
+		Name: "applike-test-gosoline-sqsq-my-queue",
+		Url:  "url://applike-test-gosoline-sqsq-my-queue",
```
applike-ss: maybe at least something like `http://localstack:4576/queue/applike-test-gosoline-sqsq-my-queue`?

pkg/sqs/service_test.go
```diff
@@ -92,3 +95,63 @@ func TestService_CreateQueue(t *testing.T) {
...
+		Attributes: map[string]*string{
+			"QueueArn": aws.String(expected.Arn),
+		},
+	}, nil)
+
+	srv := sqs.NewServiceWithInterfaces(logger, client, manager, &sqs.ServiceSettings{
+		AutoCreate: true,
+	})
+
+	input := map[string][]string{}
```
applike-ss: should we make `map[string][]string{}` a type?

pkg/sqs/service_test.go
```diff
@@ -92,3 +95,63 @@ func TestService_CreateQueue(t *testing.T) {
...
+	client.AssertExpectations(t)
+	manager.AssertExpectations(t)
+}
+
+func TestService_Purge(t *testing.T) {
+	logger := monMocks.NewLoggerMockedAll()
+	client := new(mocks.SQSAPI)
+	manager := new(cloudMocks.ResourceManager)
+
+	name := "url://applike-test-gosoline-sqsq-my-queue"
```
applike-ss: same as at the top, please use proper values

pkg/stream/pipeline.go
```diff
@@ -149,11 +151,17 @@ func (p *Pipeline) process(ctx context.Context, force bool) {
 	}()
 
 	p.ticker.Stop()
-	messages, err := p.callback.Process(ctx, p.batch)
 
-	if err != nil {
-		p.logger.Error(err, "could not process the batch")
-		return
+	var err error
```
applike-ss: why do you define this here? does it make a difference if you just deleted this and make it := in line 158 then?

pkg/stream/pipeline.go
```diff
@@ -149,11 +151,17 @@ func (p *Pipeline) process(ctx context.Context, force bool) {
...
+	var messages = p.batch
```
applike-ss: why do you write it that way? Don't we usually do it like this: `messages := p.batch` ?

pkg/stream/pipeline.go
```diff
@@ -149,11 +151,17 @@ func (p *Pipeline) process(ctx context.Context, force bool) {
...
+
+	for _, stage := range p.stages {
+		messages, err = stage.Process(ctx, messages)
+
+		if err != nil {
+			p.logger.Error(err, "could not process the batch")
```
applike-ss: should we note in which state it failed as well? Like a name for the stage or so?

pkg/stream/pipeline.go
```diff
@@ -149,11 +151,17 @@ func (p *Pipeline) process(ctx context.Context, force bool) {
...
+			return
```
applike-ss: as just discussed, the stage should remove invalid messages that should not be processed by the next stage. Due to that the return should also be removed

pkg/sqs/service.go
```diff
@@ -19,31 +21,45 @@ type ServiceSettings struct {
...
+	CreateQueue(settings Settings) (*Properties, error)
+	QueueExists(name string) (bool, error)
+	GetTaggedQueues(tagsFilter map[string][]string) ([]Properties, error)
+	GetProperties(name string) (*Properties, error)
+	GetUrl(name string) (string, error)
+	GetArn(name string) (string, error)
+	Purge(name string) error
+}
+
 type service struct {
```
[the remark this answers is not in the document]
Teaxtor: That's good a thought. But iIn this case, I wanted to prevent exporting the actual struct, so only the interface is used outside of the package.
Furthmore, the package itself is already named `sqs`, thus I did not include it in the service name :)

pkg/stream/pipeline.go
Teaxtor: Maybe we should add that later on. For now, I would assume, that the error specifies enough information.

pkg/cfg/config.go
```diff
@@ -101,6 +103,10 @@ func (c *config) keyCheck(key string) {
 	}
 }
 
+func (c *config) SetEnvReplacer(replacer *strings.Replacer) {
```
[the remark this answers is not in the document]
Teaxtor: Removed it, as it should be an internal call only

pkg/resources/resources.go
```diff
@@ -0,0 +1,75 @@
...
+	ResourceFilter []string
+	TagFilter      map[string][]string
+}
+
+type resourceManager struct {
+	client resourcegroupstaggingapiiface.ResourceGroupsTaggingAPIAPI
+	logger mon.Logger
+}
+
+func NewService(config cfg.Config, logger mon.Logger) Service {
```
applike-ss: Why not just New?

pkg/sqs/service.go
```diff
@@ -19,14 +20,25 @@ type ServiceSettings struct {
 	AutoCreate bool
 }
 
+//go:generate mockery -name Service
+type Service interface {
+	CreateQueue(settings Settings) (*Properties, error)
+	QueueExists(name string) (bool, error)
+	GetPropertiesByName(name string) (*Properties, error)
```
applike-ss: please remove variable names from interface

pkg/sqs/service.go
```diff
@@ -183,7 +141,24 @@ func (s service) QueueExists(name string) (bool, error) {
 	return false, nil
 }
 
-func (s service) GetProperties(name string) (*Properties, error) {
+func (s *service) GetPropertiesByArn(arn string) (*Properties, error) {
```
applike-ss: that looks a bit hacky, can we not instead get the queue url via its arn?

Teaxtor: Unluckily, that's not possible, without scanning all attributes for all existing queues.

pkg/resources/resources.go
Teaxtor: Because we're creating a service.

PR #32, opened 2019-09-25
apiserver: add duplication check

LordMaarten:


pkg/apiserver/crud/create.go
```diff
@@ -33,6 +33,12 @@ func (ch createHandler) Handle(ctx context.Context, request *apiserver.Request)
 	repo := ch.transformer.GetRepository()
 	err = repo.Create(ctx, model)
 
+	exists := IsDuplicateEntryError(err)
+
+	if exists {
+		return apiserver.NewStatusResponse(409), nil
```
applike-ss: please use the status codes from the go http package here (`net/http`)

pkg/apiserver/crud/handler.go
```diff
@@ -55,3 +56,15 @@ func getApiViewFromHeader(reqHeaders http.Header) string {
 
 	return DefaultApiView
 }
+
+const mysqlDuplicateEntryError = 1062
```
applike-ss: constants should be in the top part of the file

pkg/apiserver/crud/handler.go
```diff
@@ -55,3 +56,15 @@ func getApiViewFromHeader(reqHeaders http.Header) string {
...
+
+func IsDuplicateEntryError(err error) bool {
+	if err != nil {
+		mysqlErr, ok := err.(*mysql.MySQLError)
```
applike-ss: did you see the error type in your IDE? i couldn't quickly find out which type it actually is, but wouldn't have expected it to be a pointer type

pkg/apiserver/crud/handler.go
```diff
@@ -55,3 +56,15 @@ func getApiViewFromHeader(reqHeaders http.Header) string {
 
 	return DefaultApiView
 }
+
+const mysqlDuplicateEntryError = 1062
+
+func IsDuplicateEntryError(err error) bool {
```
applike-ss: I am not sure if this really belongs in the apiserver/crud namespace. I would rather see it in the db namespace.

pkg/apiserver/crud/update.go
```diff
@@ -46,6 +46,12 @@ func (uh updateHandler) Handle(ctx context.Context, request *apiserver.Request)
 
 	err = repo.Update(ctx, model)
 
+	exists := IsDuplicateEntryError(err)
+
+	if exists {
+		return apiserver.NewStatusResponse(409), nil
```
applike-ss: same as is create.go

pkg/db/client.go
```diff
@@ -13,7 +13,10 @@ import (
 	"strconv"
 )
 
-const FormatDateTime = "2006-01-02 15:04:05"
+const (
+	FormatDateTime           = "2006-01-02 15:04:05"
+	mysqlDuplicateEntryError = 1062
```
applike-ss: instead of `mysqlDuplicateEntryError`, should we call it `mysqlDuplicateEntryErrorNumber`?

PR #33, opened 2019-09-26
redis: retry on connection error;

floelm:


pkg/redis/redis_client.go
```diff
@@ -166,8 +186,29 @@ func (c *redisClient) preventOOMByBackoff(wrappedCmd func() (interface{}, error)
...
 
-	// No further error handling as the wrapped redis error is handled elsewhere
 	err = backoff.RetryNotify(operation, backoffConfig, notify)
 
-	return res
+	return res, err
+}
+
+func (c *redisClient) waitForConnection() bool {
+	backoffConfig := backoff.NewExponentialBackOff()
```
Teaxtor: We should use the same backof config as for the OOM case, to have it in the same place (and configurable eventually). The important one here is also `MaxElapsedTime = 0`, otherwise the client does not completely block

PR #37, opened 2019-10-01
currency: add conversion to any supported currency

applike-ss:


pkg/currency/service.go
```diff
@@ -55,3 +56,29 @@ func (service *CurrencyService) ToEur(value float64, from string) (float64, erro
 
 	return value / exchangeRate, nil
 }
+
+func (service *CurrencyService) ToCurrency(to string, value float64, from string) (float64, error) {
+	if from == to {
+		return value, nil
+	}
+
+	exchangeRateString, err := service.redis.HGet(ExchangeRateDataKey, to)
```
Teaxtor: I feel like getting the exchange rate as float64 could be unified with the one in `ToEur`

pkg/currency/service.go
```diff
@@ -55,3 +56,29 @@ func (service *CurrencyService) ToEur(value float64, from string) (float64, erro
 
 	return value / exchangeRate, nil
 }
+
+func (service *CurrencyService) ToCurrency(to string, value float64, from string) (float64, error) {
```
Teaxtor: Does it make sense to have a shorthand for `ToUsd`, as this is one of our most used currencies?

PR #39, opened 2019-10-02
sub: improve resilience;

j4k4:
handle panics for subscriptions


pkg/stream/input_sqs.go
```diff
@@ -97,6 +99,10 @@ func (i *sqsInput) doRun() error {
 				continue
 			}
 
+			if msg.Attributes == nil {
+				msg.Attributes = make(map[string]interface{})
```
Teaxtor: You could at least allocate one element here, as you know that you're going to put the ReceiptHandle in it

PR #40, opened 2019-10-04
db: removed connection module;

j4k4:
remove db connection module as the connection pool is handled by the sql package internally


pkg/db/client.go
```diff
@@ -39,14 +40,24 @@ type ClientSqlx struct {
...
-	return NewClient(logger, DefaultConnection)
+	if err != nil {
+		logger.Fatal(errors.New("db not booted yet"), "can not connect to sql database")
+	}
+
+	return NewClientWithInterfaces(logger, db)
 }
 
-func NewClient(logger mon.Logger, db *sqlx.DB) Client {
+//func NewClientWithDefault(logger mon.Logger) Client {
```
Teaxtor: Please remove comments :)

pkg/db/connection.go
```diff
@@ -32,103 +28,79 @@ type Settings struct {
...
 
-var barrier = sync.WaitGroup{}
-var DefaultConnection *sqlx.DB = nil
-
-func NewConnection() *Connection {
-	barrier.Add(1)
-
-	connection := Connection{
-		stop: make(chan struct{}),
+	if defaultConnection.err != nil {
```
Teaxtor: Caching the err upon first connection prevents any kind of retry logic. I suggest not to do so.

PR #41, opened 2019-10-05
test: separated localstack services from each other;

ss89:


applike-ss:
i would still like to push this @j4k4

PR #42, opened 2019-10-11
stream: allow custom metrics;

selm0:
I'd like to have some custom metrics available for the pipeline.


selm0:
Currently no need anymore, close it

PR #49, opened 2019-10-21
currency, redis: added currency functions and additional redis functions

applike-ss:


pkg/currency/service.go
```diff
@@ -17,13 +18,40 @@ const ExchangeRateDateFormat = time.RFC3339
...
+		return nil, errors.WithMessage(err, "CurrencyService: error getting currency keys")
+	}
+
+	service.currencies = currencies
+
+	return service.currencies, nil
+}
+
+func (service *CurrencyService) HasCurrency(currency string) (bool, error) {
+	return service.redis.HExists(ExchangeRateDataKey, currency)
```
LordMaarten: moving the new functions beneath the new() function would be better I would assume

PR #57, opened 2019-10-25
stream: add metrics and runner to batch consumer;

selm0:
Need batch consumers to increase throughput for sqs


applike-ss:
please rebase onto master

pkg/stream/batch_consumer.go
```diff
@@ -12,53 +11,101 @@ import (
...
 
+const MetricNameConsumerReceivedCount = "ConsumerReceivedCount"
+const MetricNameConsumerProcessedCount = "ConsumerProcessedCount"
+
 //go:generate mockery -name=BatchConsumerCallback
 type BatchConsumerCallback interface {
-	Boot(config cfg.Config, logger mon.Logger)
-	Consume(ctx context.Context, msg []*Message) error
+	Boot(cfg.Config, mon.Logger) error
+	Consume(context.Context, []*Message) ([]*Message, error)
```
applike-ss: @j4k4 suggested to keep parameter names in interfaces because they make it easier to know which parameter is for what.

pkg/stream/batch_consumer.go
```diff
@@ -12,53 +11,101 @@ import (
...
+	c.metric = metricWriter
+	c.settings = settings
+	c.ticker = time.NewTicker(settings.IdleTimeout)
+	c.batch = make([]*Message, 0, c.settings.BatchSize)
+}
+
+func (c *batchConsumer) Run(ctx context.Context) error {
 	defer c.logger.Info("leaving consumer ", c.name)
+	defer c.ticker.Stop()
+	defer c.consumeBatch(context.Background())
```
applike-ss: This is more like a `processBatch`, right?

pkg/stream/batch_consumer.go
```diff
@@ -12,53 +11,101 @@ import (
...
+	c.tracer = tracer
+	c.input = input
+	c.metric = metricWriter
+	c.settings = settings
+	c.ticker = time.NewTicker(settings.IdleTimeout)
+	c.batch = make([]*Message, 0, c.settings.BatchSize)
+}
+
+func (c *batchConsumer) Run(ctx context.Context) error {
 	defer c.logger.Info("leaving consumer ", c.name)
```
applike-ss: i think, this log message should i fact be the last deferred function that is called. That way someone who's looking at the logs doesn't get confused.

pkg/stream/batch_consumer.go
applike-ss: What's the reason to use the background context? I consider you did this because aws services cancel the request when the context gets cancelled? In that case, i think it would be ok to use the supplied context of the `Run` function

pkg/stream/batch_consumer.go
```diff
@@ -67,47 +114,86 @@ func (c *baseBatchConsumer) Run(ctx context.Context) error {
...
+
 		return
 	}
 
-	c.m.Lock()
-	defer c.m.Unlock()
+	c.metric.WriteOne(&mon.MetricDatum{
+		MetricName: MetricNameConsumerReceivedCount,
+		Value:      float64(batchSize),
+	})
```
applike-ss: should we put this into its own `writeMetric` function?

pkg/stream/consumer.go
```diff
@@ -14,68 +14,105 @@ import (
 
 const metricNameConsumerProcessedCount = "ConsumerProcessedCount"
 
+//go:generate mockery -name=Consumer
```
applike-ss: maybe you want to commit this mock too?

pkg/stream/consumer.go
```diff
@@ -14,68 +14,105 @@ import (
...
+type Consumer interface {
+	Boot(config cfg.Config, logger mon.Logger) error
+	Run(ctx context.Context) error
+	GetType() string
+}
+
 //go:generate mockery -name=ConsumerCallback
```
applike-ss: maybe you want to commit this mock too?

pkg/stream/pipeline.go
```diff
@@ -83,7 +83,7 @@ func (p *Pipeline) BootWithInterfaces(logger mon.Logger, metric mon.MetricWriter
 
 func (p *Pipeline) Run(ctx context.Context) error {
 	defer p.logger.Info("leaving pipeline")
```
applike-ss: again i believe, this log should be the last thing that is done within this function

pkg/stream/consumer.go
```diff
@@ -14,68 +14,105 @@ import (
...
-	idleTimeout := config.GetDuration("consumer_idle_timeout")
-	c.ticker = time.NewTicker(idleTimeout * time.Second)
+	mw := mon.NewMetricDaemonWriter(defaultMetrics...)
 
 	inputName := config.GetString("consumer_input")
 	input := NewConfigurableInput(config, logger, inputName)
 
-	c.input = input
-	c.ConsumerAcknowledge = NewConsumerAcknowledgeWithInterfaces(logger, input)
+	idleTimeout := config.GetDuration("consumer_idle_timeout") * time.Second
```
applike-ss: here we use consumer_idle_timeout multiplied with time.Second, in the other occurence we don't. Are we having a bug here?

pkg/stream/batch_consumer.go
Teaxtor: In this case, I want to make sure that the last batch is processed properly. When the context is already dead / canceled, any subsequent calls using it would fail. Hence, a new context.

pkg/stream/batch_consumer.go
```diff
@@ -67,47 +114,86 @@ func (c *baseBatchConsumer) Run(ctx context.Context) error {
...
-		c.consumeBatch()
+		if len(c.batch) >= c.settings.BatchSize || force {
+			c.consumeBatch(ctx)
+		}
 	}
 }
 
-func (c *baseBatchConsumer) consumeBatch() {
-	if len(c.batch) == 0 {
+func (c *batchConsumer) consumeBatch(ctx context.Context) {
```
applike-ss: wouldn't this rather be a `processBatch` or `process` function?

rchernanko:
Closing as discussed with @selm0

PR #65, opened 2019-11-07
sqs, timeutils: added metrics to queues and detect connection resets;

ajscholl:
- Detect connection resets in sqs to avoid triggering alarms for single events
- Write metrics to cloud watch to measure number of connection errors over time, so we can trigger an alarm should a problem continue for a larger timeframe
- Added TimeProvider to timeutils to have simple solution for tests having to deal with global time


j4k4:
already merged in another pr

PR #66, opened 2019-11-08
add redis client features, generate mock and add a unit test for logger

netsi:
Introduce redis functions:
    - Incr IncrBy
    - Decr DecrBy
    - Exists
    - Expire
Generated Mock for ddb.SimpleRepository
Added an unit test for mon.Logger, that tests context with fields


j4k4:
looks good

PR #72, opened 2019-11-13
sns & sqs error handling

j4k4:


pkg/cloud/request.go
```diff
@@ -0,0 +1,196 @@
...
+	stop   chan struct{}
+}
+
+func (c *delayedCancelContext) Stop() {
+	close(c.stop)
+}
+
+func WithDelayedCancelContext(parentCtx context.Context, delay time.Duration) *delayedCancelContext {
+	stop := make(chan struct{})
+	delayedCtx, cancel := context.WithCancel(context.Background())
```
ajscholl: Do we want to carry over any tracing or other things from the parent context?

pkg/cloud/request.go
```diff
@@ -0,0 +1,196 @@
...
+	}
+
+	if e.settings.Blocking {
+		backoffConfig.MaxElapsedTime = 0
+	}
+
+	backoffCtx := backoff.WithContext(backoffConfig, cancelCtx)
+
+	notify := func(err error, duration time.Duration) {
+		logger.Warnf("retrying aws service %s %s after error: %s", e.res.Type, e.res.Name, err.Error())
```
ajscholl: Will this trigger an alarm?

pkg/cloud/request.go
```diff
@@ -0,0 +1,196 @@
...
+		e.writeMetric(metricNameRetryCount)
+	}
+
+	_ = backoff.RetryNotify(func() error {
+		out, err = f(delayedCtx)
+
+		if !IsConnectionError(err) {
```
ajscholl: If we get a different error than a connection error, this will swallow an error

pkg/cloud/request.go
j4k4: if we want to do that, i would want to do that in a follow up pr..
currently it will not hurt that much, as we're not doing any logging in the AwsRequestFunction

pkg/cloud/request.go
j4k4: plan is to adjust the warning alarms that they are not triggering every single time

pkg/cloud/request.go
j4k4: no, if it is not a connection error, the backoff will be stopped and the error will still be returned at the end of the execute function

ajscholl: Right, overlooked that we ignore the error. So this should work then.

Teaxtor: Here, go 1.13 would be an advantage.. But I wasn't able to wrap my head around a overall good approach with the new error handling yet.

PR #74, opened 2019-11-14
[parquet] read generic parquet format, adapt to new datalake format;

selm0:


selm0:
Closed due to deprecation

PR #90, opened 2019-11-25
Postgre db

selm0:
Allow for arbitrary, sql based, databases to be used.
Any database needs to support the sql.Driver interface.


j4k4:
+1

PR #91, opened 2019-11-26
Sub kvstore: use app_group instead of app_name;

rchernanko:


j4k4:
one should use the target settings of the subscription instead

PR #92, opened 2019-11-26
handle time.Time for configuration

j4k4:
* added `GetTime` to config getters  
* sanitizing `time.Time` to string on merging settings
* decode string to `time.Time` on `UnmarshalKey`


pkg/cfg/sanitizer.go
```diff
@@ -0,0 +1,47 @@
...
+func Sanitize(in map[string]interface{}, sanitizers []Sanitizer) (interface{}, error) {
+	var err error
+
+	if len(sanitizers) == 0 {
+		return in, nil
+	}
+
+	for key, val := range in {
+		for _, san := range sanitizers {
+			if in[key], err = san(val); err != nil {
```
selm0: I suggest to not automically overwrite the map content, as in case of err the value returned from the sanitizer should not (must not) be trusted. This could lead to erasing contents.

pkg/cfg/sanitizer.go
```diff
@@ -0,0 +1,47 @@
...
+				return nil, fmt.Errorf("can not apply sanitizer on key %s: %w", key, err)
+			}
+		}
+
+		if valMap, ok := val.(map[string]interface{}); ok {
```
selm0: Will this code be actually reached? When I understand it correctly, any error (of course depending on the sanitizer), will trigger a return from this function.. Thus giving a map as a parameter to all the sanitizers might return prematurely.

pkg/cfg/sanitizer.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+func TimeSanitizer(in interface{}) (interface{}, error) {
+	if reflect.TypeOf(in) != reflect.TypeOf(time.Time{}) {
+		return in, nil
+	}
+
+	tm, ok := in.(time.Time)
+
+	if !ok {
```
selm0: Will this branch be reachable? You already check the value type of the input to be of `time.Time`, why should the type assertion fail?

pkg/cfg/sanitizer_test.go
```diff
@@ -0,0 +1,46 @@
...
+	san, err := cfg.Sanitize(in, []cfg.Sanitizer{
+		cfg.TimeSanitizer,
+	})
+
+	assert.NoError(t, err)
+	assert.IsType(t, map[string]interface{}{}, san)
+
+	s := objx.New(san)
+
+	assert.Equal(t, "bar", s.Get("foo").Data())
```
selm0: You should also try to cover the error cases in your code - are they actually reachable / viable?
You don't want to introduce errors while handling errors ;-)

pkg/cfg/sanitizer.go
```diff
@@ -0,0 +1,47 @@
...
+		return in, fmt.Errorf("value is not of type time.Time but instead of %T", in)
+	}
+
+	return tm.Format(time.RFC3339), nil
```
selm0: +1 for a good standard :D

pkg/cfg/config.go
```diff
@@ -380,9 +401,15 @@ func (c *config) keyCheck(key string) {
 }
 
 func (c *config) mergeSettings(settings map[string]interface{}) error {
+	sanitized, err := Sanitize(settings, c.sanitizers)
+
+	if err != nil {
+		return fmt.Errorf("could not sanitzie settings on merge: %w", err)
```
selm0: typo

pkg/cfg/config.go
```diff
@@ -231,6 +234,23 @@ func (c *config) GetStringSlice(key string) []string {
...
+	c.lck.Lock()
+	defer c.lck.Unlock()
+
+	c.keyCheck(key)
+
+	data := c.get(key)
+	tm, err := cast.ToTimeE(data)
+
+	if err != nil {
+		c.err(err, "can not cast value %v[%T] of key %s to []string", data, data, key)
```
selm0: I find this error message misleading. You're not actually casting anything to a string slice. And the type looks like a slice/map access operation.
Instead, I suggest going with something like `can not cast value %v [type %T] of key %s to time.Time`

pkg/cfg/config.go
j4k4: fixed

pkg/cfg/config.go
j4k4: fixed

pkg/cfg/sanitizer.go
j4k4: indeed

pkg/cfg/sanitizer.go
j4k4: well, i think one could skip that..

PR #95, opened 2019-11-29
db-repo: do not update createdAt to 0000-00-00 00:00:00 if it is not set;

ajscholl:


pkg/db-repo/model.go
```diff
@@ -24,6 +24,11 @@ type TimeStampable interface {
 	SetCreatedAt(createdAt *time.Time)
 }
 
+type HasTimestamps interface {
```
j4k4: usually we use the `*Aware` pattern like `TimestampAware`..

ajscholl: Fixed.

PR #100, opened 2019-12-04
handle used connection error

j4k4:


rchernanko:
Merging as discussed with Jan

PR #105, opened 2019-12-11
[parquet] generic read and write;

selm0:
w mapstruct


pkg/parquet/reader.go
```diff
@@ -209,6 +252,76 @@ func (r *s3Reader) listFiles(prefix string) ([]string, error) {
...
+	return func(f reflect.Type, t reflect.Type, data interface{}) (interface{}, error) {
+		if f.Kind() != reflect.String || t != reflect.TypeOf(time.Time{}) {
+			return data, nil
+		}
+
+		parquetDate := []byte(data.(string))
+
+		nano := binary.LittleEndian.Uint64(parquetDate[:8])
+		dt := binary.LittleEndian.Uint32(parquetDate[8:])
+
```
floelm: maybe we should have a short explanation what happens here

PR #110, opened 2019-12-13
parquet: write time.Time like firehose;

selm0:
Fix writing timestamps, use the same format as AWS firehose

### DIFF
--- a/pkg/parquet/reader.go
+++ b/pkg/parquet/reader.go
@@ -2,7 +2,6 @@ package parquet
 
 import (
 	"context"
-	"encoding/binary"
 	"fmt"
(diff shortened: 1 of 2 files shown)


pkg/parquet/writer.go
```diff
@@ -73,6 +73,7 @@ func (w *s3Writer) Write(ctx context.Context, items interface{}) error {
 
 		if current.IsZero() || item.GetCreatedAt().Sub(current) > w.settings.Interval {
 			current = item.GetCreatedAt()
+			fmt.Printf("%v", current)
```
floelm: can we remove this?

selm0: Forgot to remove it

PR #112, opened 2019-12-15
stream: cleaned up the multiple output implementation;

### LINKED ISSUE #111: stream output multiple cleanup
instead of residing in `output_configurable_multiple.go` the implementation should be in `output_multiple.go` to be in line with the other outputs.  

The signature of `func NewConfigurableMultiOutput(config cfg.Config, logger mon.Logger, base string) Output` should be `func NewConfigurableMultiOutput(config cfg.Config, logger mon.Logger, outputs []Output) Output` and `func NewConfigurableMultiOutputWithInterfaces(logger mon.Logger, outputs []Output)` to be able to create this output type without configuration file.

To configure this output in `output_configurable.go` a structure of
```go
type multipleOutputConfiguration struct {
	Outputs []string `cfg:"outputs"`
}
```
should be used in conjunction with a 'builder' function ` newMultipleOutputFromConfig` which reads the config and creates the outputs to provide them to the `NewConfigurableMultiOutput` function.

in the end the multiple output should be configured like this:  
```yml
stream:
  output:
    pipeline:
      type: multiple
      outputs: [outputA, outputB]
    outputA:
      type: kinesis
      stream_name: kinesis-stream-a
    outputB:
      type: kinesis
      stream_name: kinesis-stream-b
```

applike-ss:
fixes #111


pkg/stream/output_configurable.go
```diff
@@ -62,9 +62,19 @@ func newKinesisOutputFromConfig(config cfg.Config, logger mon.Logger, name strin
 		StreamName: settings.StreamName,
 	})
 }
+func newMultipleOutputFromConfig(config cfg.Config, logger mon.Logger, name string) Output {
```
j4k4: we have an empty line missing here ;)

pkg/stream/output_multiple.go
```diff
@@ -2,20 +2,36 @@ package stream
 
 import (
 	"context"
-	"fmt"
 	"github.com/applike/gosoline/pkg/cfg"
 	"github.com/applike/gosoline/pkg/mon"
 	"github.com/hashicorp/go-multierror"
 )
 
+type MultipleOutputConfiguration struct {
```
j4k4: this struct belongs into `output_configurable.go` together with the other configuration structs

j4k4 (commented):
had a brief look and added 2 comments

PR #117, opened 2020-01-06
cloud: fetching services by tag instead of hardcoded assumptions;

applike-ss:


pkg/cloud/services.go
```diff
@@ -275,21 +258,81 @@ func (c *AwsServiceClient) GetServiceList(filter *FilterServicesInput) []Service
...
+	for key, value := range tags {
+		v, ok := listing.Tags[key]
+
+		if !ok {
+			return false
+		}
+
+		foundValue := false
+
+		for _, subvalue := range value {
```
rchernanko: camelCase subValue please

pkg/cloud/services.go
```diff
@@ -275,21 +258,81 @@ func (c *AwsServiceClient) GetServiceList(filter *FilterServicesInput) []Service
...
+				foundValue = true
+			}
+		}
+
+		if !foundValue {
+			return false
+		}
+	}
+
+	return true
```
rchernanko: so if there are no 'tags' this function would return true. is that what we want?

pkg/cloud/services.go
```diff
@@ -275,21 +258,81 @@ func (c *AwsServiceClient) GetServiceList(filter *FilterServicesInput) []Service
...
+
+		foundValue := false
+
+		for _, subvalue := range value {
+			if subvalue == v {
+				foundValue = true
+			}
+		}
+
+		if !foundValue {
```
rchernanko: do we need this? can we not just return foundValue?

pkg/cloud/services.go
```diff
@@ -275,21 +258,81 @@ func (c *AwsServiceClient) GetServiceList(filter *FilterServicesInput) []Service
...
+		},
+	})
 
 	if err != nil {
-		panic(err)
+		return nil, err
+	}
+
+	if len(svc.Services) > 1 {
+		return nil, fmt.Errorf("there is more than one service with the specified arn")
```
rchernanko: Should we not use errors.New instead (given that there is no string formatting)? Same applies for below

pkg/cloud/services.go
```diff
@@ -275,21 +258,81 @@ func (c *AwsServiceClient) GetServiceList(filter *FilterServicesInput) []Service
...
+
+		if !ok {
+			return false
+		}
+
+		foundValue := false
+
+		for _, subvalue := range value {
+			if subvalue == v {
+				foundValue = true
```
rchernanko: this value can be overridden though. is that right?

pkg/cloud/services.go
applike-ss: true

pkg/cloud/services.go
applike-ss: no, because we are in a loop and want to check that for each tag name there is one matching value

pkg/cloud/services.go
applike-ss: if we don't search for any tags, then is has the tags that were searched for :)

pkg/cloud/services.go
applike-ss: fine with that :+1:

pkg/cloud/services.go
applike-ss: i don't understand this, however i added a break to early abort searching any further

pkg/cloud/services.go
rchernanko: makes sense

pkg/cloud/services.go
rchernanko: ok

pkg/cloud/services.go
rchernanko: remember to change the one on line 282 as well

pkg/cloud/services.go
rchernanko: as discussed, all good

pkg/cloud/services.go
applike-ss: nice, did that

PR #130, opened 2020-01-16
logger: copy fields and structs passed into the logger deeply;

ajscholl:
In the past, we would only store a reference to maps passed into
our logger with WithFields, WithContext, a context resolver or any
other means. As we needed to transform the errors to strings for
them to appear in the json encoded message, we replaced the value
in the map directly. However, this caused concurrent reads and writes
to the map to happen if more than one thread was using the same map.

We later started to take a copy of the map, so we were a lot less likely
to run into a concurrent read and write. Still, we did not account for
nested maps and objects, so a concurrent read and write panic could
still happen.

The new solution is to perform a deep clone of any maps, structs or
arrays (basically anything meaningful represented by a pointer) and
avoid writing anything into the maps we get passed as input. Performance
wise this could be a lot slower, but in most cases one should not pass
nested maps or pass them as map[string]interface{}. In the latter case
we have a special case to do the copy directly without resorting to
reflection. All other cases are handled by reflection.

### DIFF
--- a/pkg/mon/logger.go
+++ b/pkg/mon/logger.go
@@ -8,6 +8,7 @@ import (
 	"github.com/jonboulle/clockwork"
 	"io"
 	"os"
+	"reflect"
 	"runtime"
 	"strconv"
 	"strings"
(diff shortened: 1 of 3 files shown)


j4k4:
+1

PR #140, opened 2020-01-23
kernel: introduce staged kernel;

### LINKED ISSUE #135: metric: not written after last foreground module exited
I have one foreground module which did some http request, wrote a metric and exited. It seems like the metric writer did not in that time start and finished writing the metric (maybe it was still waiting to batch the metric request). The kernel then started to shut down which sometimes caused the metric module not to write the metric at all - even though it was logged. It could be that the context was already canceled and thus the metric was not written.

TODO: Investigate whether this is indeed the case. If yes, we might need to use something like a delayed cancel context for the metric writer to avoid losing metrics written only during kernel shutdown.

ajscholl:
Closes #135.

This commit addresses the following problems:
 * If you invoke a second kernel, the timeout killer of the first kernel
       would still run and call os.Exit(1) eventually. We now stop the
       timeout after a proper kernel exit.
 * The metric writer could never be sure when every metric was written
       to its channel and it could safely assume the channel was empty.
       Therefore we introduced kernel stages - now every other module exists
       before the metric writer does, therefore no-one should be able to
       still write some metrics.


pkg/kernel/signal_once.go
```diff
@@ -0,0 +1,48 @@
+package kernel
+
+import (
+	"sync"
+)
+
+//go:generate mockery -name=SignalOnce
+type SignalOnce interface {
```
josketres: 👍 for this abstraction. I think it was already being used like that in `kernel.go`. There was a channel that was just being closed. But having this type makes the intention of that channel much more clear.

PR #145, opened 2020-01-27
redis: added HSetNX operation;

applike-ss:


pkg/redis/redis_client.go
```diff
@@ -39,6 +39,7 @@ type Client interface {
 	HKeys(string) ([]string, error)
 	HGet(string, string) (string, error)
 	HSet(string, string, interface{}) error
+	HSetNX(key string, field string, value interface{}) (bool, error)
```
bogdanfinn: the parameters are named here. Maybe also use unnamed parameters here for consistency?

pkg/redis/mocks/Client.go
```diff
@@ -253,6 +253,27 @@ func (_m *Client) HSet(_a0 string, _a1 string, _a2 interface{}) error {
 	return r0
 }
 
+// HSetNX provides a mock function with given fields: key, field, value
+func (_m *Client) HSetNX(key string, field string, value interface{}) (bool, error) {
```
bogdanfinn: Maybe write a test in pkg/redis/redis_client_test.go ?

pkg/redis/redis_client.go
applike-ss: named parameters are wanted as of now, because a user of the interface should not be required to read the implementation

pkg/redis/mocks/Client.go
applike-ss: done that :)

PR #150, opened 2020-01-31
http: response should include the header

netsi:
In some cases its important to read the response Header.

Not sure if we should wrap the `net/http Header`


ajscholl:
Any chance we could get this merged eventually? Or is there a reason why we did not yet merge this?

j4k4:
yes, we can merge it, just wasn't aware of it.. :)

PR #162, opened 2020-02-12
Lock for file output

floelm:


pkg/stream/output_file.go
```diff
@@ -30,6 +32,9 @@ func (o *fileOutput) WriteOne(ctx context.Context, msg *Message) error {
 }
 
 func (o *fileOutput) Write(ctx context.Context, batch []*Message) error {
+	o.lck.Lock()
+	defer o.lck.Unlock()
+
 	flags := os.O_CREATE | os.O_WRONLY
```
applike-ss: wouldn't you need to use os.O_EXCL for this to have any effect?
Usually you would lock something if a given resource is being used, however in this case you are:
* locking a mutex lock
* opening a file
* writing to the file
* closing the file handle
* unlocking the mutex lock

applike-ss: In other words: should we open the file handle when creating the output instead or should we ensure that we are exclusively opening the file on operating system side?

PR #175, opened 2020-02-18
first base structure of a fixture loader module which handles multipl…

bogdanfinn:
…e input and output options;

Config would look like this in the current iteration:
```
fixture_loader:
  - enable: true
    purge: true
    input:
      path: 'file://../../some/path/to/fixtures/directory'
      encoding: json
    output: mysql
  - enable: true
    purge: true
    input:
      path: 'file://../../some/path/to/fixtures/directory'
      encoding: json
    output: mysql
```
enable is default to false.
purge is default to true.

### DIFF
--- a/pkg/fixtures/json_reader.go
+++ b/pkg/fixtures/json_reader.go
@@ -1,0 +1,16 @@
+package fixtures
+
+import (
+	"github.com/applike/gosoline/pkg/cfg"
+	"github.com/applike/gosoline/pkg/mon"
+)
+
+type jsonFixtureReader struct {}
+
+func NewJsonFixtureReader(cfg cfg.Config, logger mon.Logger) FixtureReader {
+	return &jsonFixtureReader{}
+}
+
+func (j *jsonFixtureReader) ReadFixtures() (interface{}, error){
+	return "",nil
+}
--- a/pkg/fixtures/loader.go
+++ b/pkg/fixtures/loader.go
@@ -1,0 +1,84 @@
+package fixtures
+
+import (
+	"context"
+	"fmt"
+	"github.com/applike/gosoline/pkg/cfg"
+	"github.com/applike/gosoline/pkg/kernel"
+	"github.com/applike/gosoline/pkg/mon"
+)
+
+type FixtureLoader struct {
+	kernel.BackgroundModule
+	logger       mon.Logger
+	Writers []FixtureWriter
+	Settings []FixtureLoaderSettings `cfg:"fixture_loader"`
+}
+
+func NewFixtureLoader() *FixtureLoader {
+	return &FixtureLoader{}
+}
+
+func (f *FixtureLoader) Boot(config cfg.Config, logger mon.Logger) error {
+	f.logger = logger
+
+	if !config.IsSet("fixture_loader"){
+		f.logger.Info("fixture loader module not configured")
+		return nil
+	}
+
+	var s []FixtureLoaderSettings
+	config.UnmarshalKey("fixture_loader", &s)
+	f.Settings = s
+
+	err := f.BootFixtureWriters(config, logger)
+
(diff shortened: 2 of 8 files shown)


pkg/fixtures/json_reader.go
```diff
@@ -0,0 +1,16 @@
+package fixtures
+
+import (
+	"github.com/applike/gosoline/pkg/cfg"
+	"github.com/applike/gosoline/pkg/mon"
+)
+
+type JsonFixtureReader struct {}
```
josketres: 💅 - this struct can be private `jsonFixtureReader `

pkg/fixtures/loader.go
```diff
@@ -0,0 +1,90 @@
...
+
+		fixtureWriter := fixtureWriterFactory(config, logger)
+		fixtureWriter.AddFixtureReader(readerFactory(config, logger))
+		f.Writers[s.Output] = fixtureWriter
+		continue
+	}
+}
+
+func (f *FixtureLoader) Run(ctx context.Context) error {
+	f.logger.Info("Run Fixture Loader")
```
josketres: Maybe we can replace this with a comment 
```suggestion
   // do nothing: fixtures are loaded during Boot
```

pkg/fixtures/loader.go
```diff
@@ -0,0 +1,90 @@
...
+		if !s.Enabled {
+			f.logger.Info(fmt.Sprintf("Fixture Writer for Input %s and Output %s not enabled", s.Input.Encoding, s.Output))
+			continue
+		}
+
+		readerFactory, ok := fixtureReaders[s.Input.Encoding]
+
+		if !ok {
+			f.logger.Warn("No reader implemented for input encoding ", s.Input.Encoding)
+			continue
```
josketres: I would prefer to throw an error here. 

Fixture loading is something that we only activate for tests or for development, so if we start an app with an unsupported configuration it is better to just fail immediately with an error indicating what went wrong.

I think it is a better development user experience: DEVUX

pkg/fixtures/loader.go
```diff
@@ -0,0 +1,90 @@
...
+		if fixtureWriter, ok := f.Writers[s.Output]; ok {
+			fixtureWriter.AddFixtureReader(readerFactory(config, logger))
+			f.Writers[s.Output] = fixtureWriter
+			continue
+		}
+
+		fixtureWriterFactory, ok := fixtureWriters[s.Output]
+
+		if !ok {
+			f.logger.Warn("No writer existing for output type ", s.Output)
```
josketres: Same as above. We should fail fast here.

pkg/fixtures/loader.go
```diff
@@ -0,0 +1,90 @@
...
+
+		readerFactory, ok := fixtureReaders[s.Input.Encoding]
+
+		if !ok {
+			f.logger.Warn("No reader implemented for input encoding ", s.Input.Encoding)
+			continue
+		}
+
+
+		if fixtureWriter, ok := f.Writers[s.Output]; ok {
```
josketres: This part with the fixtureWriter can be simplified. It is difficult to understand.

pkg/fixtures/loader.go
```diff
@@ -0,0 +1,84 @@
...
+	if err != nil {
+		return err
+	}
+
+	for _, writer := range f.Writers {
+		err := writer.WriteFixtures()
+
+		if err != nil {
+			f.logger.Error(err, "error occurred during fixture loading")
+			return err
```
josketres: Can be replaced with error:
```go
return fmt.Errorf("error occurred during fixture loading: %w", err)
```

josketres: See https://blog.golang.org/go1.13-errors

PR #179, opened 2020-02-19
test: fixes health checks for components, sometimes there is an issue…

applike-ss:
… because components are not ready yet;


pkg/test/component_kinesis.go
```diff
@@ -65,9 +66,16 @@ func doRunKinesis(name string, configMap configInput) {
...
-			_, err := kinesisClient.ListStreams(&kinesis.ListStreamsInput{})
 
-			return err
+			_, err = kinesisClient.ListStreams(&kinesis.ListStreamsInput{})
+
+			if err != nil {
+				return err
+			}
+
+			time.Sleep(1 * time.Second)
```
josketres: 😱

pkg/test/component_sns.go
```diff
@@ -85,9 +86,16 @@ func doRunSns(name string, configMap configInput) {
 		},
 		HealthCheck: func() error {
 			snsClient := ProvideSnsClient(name)
-			_, err := snsClient.ListTopics(&sns.ListTopicsInput{})
 
-			return err
+			_, err = snsClient.ListTopics(&sns.ListTopicsInput{})
```
josketres: Did you take a look at the `ListTopicsResponse`?
Maybe it contains something interesting like a "PleaseRetryAgainIn1Second" response.

pkg/test/component_kinesis.go
josketres: if we really need to add this sleep here maybe we should add a small comment explaining why it was done.

pkg/test/component_sns.go
applike-ss: nothing interesting there that we could make use of

josketres:
PR Duplicated.

PR #183, opened 2020-02-20
fixtures: always use update for fixtures and create if not existing;

bogdanfinn:


bogdanfinn:
@josketres Update will create the entry if it is not present...

PR #185, opened 2020-02-21
stream: added kinesis stream auto create;

applike-ss:
closes #185

### DIFF
--- a/pkg/stream/kinesis.go
+++ b/pkg/stream/kinesis.go
@@ -1,0 +1,58 @@
+package stream
+
+import (
+	"github.com/applike/gosoline/pkg/cfg"
+	"github.com/applike/gosoline/pkg/mon"
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/kinesis"
+	"github.com/aws/aws-sdk-go/service/kinesis/kinesisiface"
+)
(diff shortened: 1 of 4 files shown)


pkg/stream/kinesis.go
```diff
@@ -0,0 +1,58 @@
...
+	"github.com/aws/aws-sdk-go/service/kinesis"
+	"github.com/aws/aws-sdk-go/service/kinesis/kinesisiface"
+)
+
+type ResourceNameGetter interface {
+	GetResourceName() string
+}
+
+func createKinesisStream(config cfg.Config, logger mon.Logger, client kinesisiface.KinesisAPI, settings ResourceNameGetter) {
+	autoCreate := config.GetBool("aws_kinesis_autoCreate")
```
josketres: It would be a nice thing to add a check here so that we can run a application without being forced to add this key to the configuration.
```suggestion
	if  !config.IsSet("aws_kinesis_autoCreate") {
		return
	}
	
	autoCreate := config.GetBool("aws_kinesis_autoCreate")
```

applike-ss: i don't like it because the behaviour is kind of magic to someone who doesn't know gosoline and who would want to use such a feature but is not aware of its existence.

PR #188, opened 2020-02-25
fixtures: removed WriterMetdata property in favour of a factory;

### LINKED ISSUE #186: fixtures: Move WriterMetadata to FixtureWriter factories
in the FixtureSet struct, we don't need the WriterMetadata field, because we want to set the Writer via a factory method that already has the required metadata information.

applike-ss:
closes #186 #188

### DIFF
--- a/pkg/fixtures/ddb_kv_writer.go
+++ b/pkg/fixtures/ddb_kv_writer.go
@@ -16,34 +16,40 @@ type KvstoreFixture struct {
 }
 
 type dynamoDbKeyValueFixtureWriter struct {
-	config cfg.Config
-	logger mon.Logger
+	config  cfg.Config
+	logger  mon.Logger
+	modelId *mdl.ModelId
 }
 
-func NewDynamoDbKvStoreFixtureWriter(cfg cfg.Config, logger mon.Logger) FixtureWriter {
-	return cachedWriters.New("dynamoDbKvStore", func() FixtureWriter {
-		return &dynamoDbKeyValueFixtureWriter{
-			config: cfg,
-			logger: logger,
-		}
-	})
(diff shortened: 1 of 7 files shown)


pkg/fixtures/ddb_kv_writer.go
```diff
@@ -16,8 +16,18 @@ type KvstoreFixture struct {
...
-	config cfg.Config
-	logger mon.Logger
+	config  cfg.Config
+	logger  mon.Logger
+	modelId *mdl.ModelId
+}
+
+func DynamoDbKvStoreFixtureWriterFactory(modelId *mdl.ModelId) FixtureWriterFactory {
+	return func(cfg cfg.Config, logger mon.Logger) FixtureWriter {
+		writer := NewDynamoDbKvStoreFixtureWriter(cfg, logger).(*dynamoDbKeyValueFixtureWriter)
```
josketres: Instead of the type assertion we should change the return type of `NewDynamoDbKvStoreFixtureWriter`, and we can also make it private: `newDynamoDbKvStoreFixtureWriter `

josketres: Same comment applies for all other writers

applike-ss: if we did that, then we would not be able to use the cachedWriter. Still want that change?

applike-ss: i just had a second look at it and it might be that we don't need the cache anymore, because each fixture writer would now be a different instance of the same struct due to the additional modelId/metadata that i added in this ticket.

josketres: 👍

josketres: Makes sense. Kill the cache!

PR #193, opened 2020-02-26
fixtures: cleanup + safeguard from running in prod;

josketres:


pkg/fixtures/loader.go
```diff
@@ -6,48 +8,46 @@ import (
...
-	logger = logger.WithChannel("fixture_loader")
+func (f *fixtureLoader) Load(fixtureSets []*FixtureSet) error {
+	settings := fixtureLoaderSettings{}
+	f.config.UnmarshalKey("fixtures", &settings)
 
-	if !config.IsSet("fixture_loader_enabled") {
-		logger.Info("fixture loader is not configured")
-		return nil
-	}
+	f.logger = f.logger.WithChannel("fixture_loader")
```
applike-ss: This belongs in the NewFixtureLoader func, right?

test/test_configs/config.fixtures_dynamodb.test.yml
```diff
@@ -1,6 +1,7 @@
-fixture_loader_enabled: true
-
 aws_cloudwatch_endpoint: http://172.17.0.1:4582
 aws_dynamoDb_endpoint: http://172.17.0.1:4569
 aws_dynamoDb_autoCreate: true
 aws_sdk_retries: 30
+
+fixtures:
```
applike-ss: should we name this fixture_loader instead?

pkg/fixtures/loader.go
josketres: ✅

test/test_configs/config.fixtures_dynamodb.test.yml
josketres: Will keep it as it is to be consistent and use the package name.

PR #197, opened 2020-03-02
fixtures: cleanup config usage;

josketres:
**Summary:**

- Do not store an instance of `cfg.Config` in any of our objects.
- Use the existing pattern of using a factory function to store the config in its closure.


pkg/fixtures/loader.go
```diff
@@ -8,29 +8,36 @@ import (
...
 
+// fixtureWriterFactoryWrapper saves a config instance in its closure and uses it to invoke FixtureWriterFactory
+type fixtureWriterFactoryWrapper func(factory FixtureWriterFactory) FixtureWriter
+
 type fixtureLoader struct {
-	config cfg.Config
-	logger mon.Logger
+	logger        mon.Logger
+	settings      *fixtureLoaderSettings
+	writerFactory fixtureWriterFactoryWrapper
```
bogdanfinn: inline `func(factory FixtureWriterFactory) FixtureWriter`

PR #199, opened 2020-03-03
fixtures: implemented new fixture loader for plain mysql;

bogdanfinn:


pkg/db/insert_builder.go
```diff
@@ -0,0 +1,40 @@
...
+)
+
+//go:generate mockery -name InsertBuilder
+type InsertBuilder interface {
+	Table(table string) InsertBuilder
+	Columns(columns []string) InsertBuilder
+	Values(values ...interface{}) InsertBuilder
+}
+
+type RawInsertBuilder struct {
```
josketres: Do we need to export this type?
If not I'd be nice to make it private `rawInsertBuilder`

pkg/db/insert_builder.go
```diff
@@ -0,0 +1,40 @@
...
+	Builder squirrel.InsertBuilder
```
josketres: Do we need to export this field?
I'd suggest we make this private (see principle of encapsulation).

test/fixtures_mysql_test.go
```diff
@@ -45,33 +76,48 @@ func mysqlTestFixtures() []*fixtures.FixtureSet {
...
+			Writer:  fixtures.PlainMySqlFixtureWriterFactory(),
+			Fixtures: []interface{}{
+				&fixtures.PlainMySqlFixture{
+					TableName: "mysql_test_models",
+					Rows: []fixtures.PlainMySqlFixtureRow{
+						{
+							"id":   2,
+							"name": "testName2",
+						},
+					},
```
josketres: I really like this approach of writing a map!

bogdanfinn:
We updated the squirrel to 1.2.0 due to the fact we need the "Replace Into" while building sql statements.

https://github.com/Masterminds/squirrel/compare/v1.1.0...v1.2.0

pkg/fixtures/writer_plain_mysql.go
```diff
@@ -70,13 +71,12 @@ func (m *plainMySqlFixtureWriter) Write(fs *FixtureSet) error {
 }
 
 func buildSql(fixture *PlainMySqlFixture) (string, []interface{}, error) {
-	insertBuilder := squirrel.InsertBuilder{}.PlaceholderFormat(squirrel.Question)
+	insertBuilder := squirrel.Replace(fixture.TableName).PlaceholderFormat(squirrel.Question)
```
josketres: 💅 
```go
insertBuilder := squirrel.Replace(fixture.TableName).
   PlaceholderFormat(squirrel.Question).
   Columns(fixture.Columns...)
```

PR #202, opened 2020-03-05
fixtures: added redis and redis kvstore fixture writer;

### LINKED ISSUE #167: fixture-loader: redis
- Implement a fixture-loader for `redis`.
- Add an example/test.

applike-ss:
closes #167


go.mod
```diff
@@ -33,7 +33,7 @@ require (
 	github.com/golang-migrate/migrate/v4 v4.2.5
 	github.com/gomodule/redigo v2.0.0+incompatible // indirect
 	github.com/google/go-querystring v1.0.0
-	github.com/google/uuid v1.0.0
+	github.com/google/uuid v1.1.1
```
josketres: Are this changes intended?

pkg/fixtures/writer_ddb.go
```diff
@@ -45,7 +45,7 @@ func (d *dynamoDbFixtureWriter) Write(fs *FixtureSet) error {
 		}
 	}
 
-	d.logger.Infof("loaded %d dynamo db fixtures", len(fs.Fixtures))
+	d.logger.Infof("loaded %d dynamodb fixtures", len(fs.Fixtures))
```
josketres: 👍

pkg/fixtures/writer_redis.go
```diff
@@ -0,0 +1,76 @@
...
+	"fmt"
+	"github.com/applike/gosoline/pkg/cfg"
+	"github.com/applike/gosoline/pkg/mon"
+	"github.com/applike/gosoline/pkg/redis"
+	"reflect"
+	"time"
+)
+
+const (
+	RedisOpRpush = "RPUSH"
```
josketres: should this constants be exported ?

pkg/fixtures/writer_redis.go
```diff
@@ -0,0 +1,76 @@
...
+	RedisOpSet   = "SET"
+)
+
+type redisOpHandler func(client redis.Client, data ...interface{}) error
+
+var redisHandlers = map[string]redisOpHandler{
```
josketres: 👍

test/fixtures_redis_test.go
```diff
@@ -0,0 +1,135 @@
...
+	assert.JSONEq(s.T(), `{"name":"foo","age":123}`, res)
+}
+
+func redisFixtures() []*fixtures.FixtureSet {
+	return []*fixtures.FixtureSet{
+		{
+			Enabled: true,
+			Writer:  fixtures.RedisFixtureWriterFactory(aws.String("default"), aws.String(fixtures.RedisOpSet)),
+			Fixtures: []interface{}{
+				[]interface{}{
```
josketres: I think we can improve the types here.
Instead of having `[]interface{}` we could have a proper struct. Something like this:
```go
type RedisFixture struct {
    Key: string,
    Value: interface{},
    Duration: time.Duration
}
```

go.mod
applike-ss: no, it happened automatically. Afaik we decided to just let go.mod and go.sum be updated when it happens. Maybe ask @j4k4 if you think this is the wrong behaviour.

test/fixtures_redis_test.go
applike-ss: 👍 i put that (slightly altered) into the redis fixture writer file

pkg/cfg/config.go
```diff
@@ -74,6 +74,21 @@ func NewWithInterfaces(lookupEnv LookupEnv) *config {
 	return cfg
 }
 
+// load and merge multiple configuration files
+func FromFiles(filePaths []string, fileType string) GosoConf {
```
josketres: 👍 Cool I can use this also in my branch. 

Just a small suggestion. If you change the order of the arguments you can make this a variadic function which would be nicer to use.

pkg/cfg/config.go
```diff
@@ -74,6 +74,21 @@ func NewWithInterfaces(lookupEnv LookupEnv) *config {
...
+	cfg := New()
+
+	for _, filePath := range filePaths {
+		err := WithConfigFile(filePath, fileType)(cfg.(*config))
```
josketres: I did clean that up already in my branch. I would suggest to use the `Option` method instead. 
```suggestion
		err := cfg.Option(WithConfigFile(filePath, fileType))
```

pkg/cfg/config.go
j4k4: Please don't extend the interface of the config..
We're using the option pattern here. Please try to stick with that..

applike-ss: @j4k4 can we define some rules when convenience functions are allowed and when not?

applike-ss:
@josketres i did the changes requested by @j4k4, can you re-evaluate if i can squash, rebase, merge this?

pkg/cfg/config.go
j4k4: Actually, i have misread it.. thought the function would have a config receiver and would be in the interface..

So this one is not as bad as I thought.. :grimacing:

Nevertheless, I would move this one to the test package as this is more of an edge case or special use case?

go.mod
j4k4: Do we have defined a general rule here?
But minor updates should be safe, right?

pkg/cfg/config.go
j4k4: It's hard to define rules for this kind of stuff..
Actually, at first i read this one wrong and thought this function would have the Config as a receiver.. As this is a convenience function, it's not as bad as i thought..
Nevertheless, i wouldn't put it into the cfg package as this function is based on your use case in the test package? So it would be better to have this in there?

PR #203, opened 2020-03-06
blob, ddb: avoid error log on existing bucket and strip out additional json arguments;

ajscholl:
If you are using a real (no localstack) endpoint for S3 and use
autoCreate, S3 might return an error telling you the bucket does already
exist. We handled this in the past (failing to create a bucket because
it already exsits is not really an error, especially in the context of
autoCreate), but S3 can return a second error telling you you already
own the bucket instead. As this is basically the same situation, also
ignore that error.

For ddb, if we automatically generate the table, we have to strip any
json tags, otherwise a line like 'Foo string `json:"foo,omitempty"`'
would try to create a field called "foo,omitempty" instead of foo. This
then breakes if the the field is used for any keys because it can later
not be found (because you only reference foo instead of foo,omitempty).

### DIFF
--- a/pkg/blob/store.go
+++ b/pkg/blob/store.go
@@ -274,7 +274,8 @@ func isBucketAlreadyExistsError(err error) bool {
 	}
 
 	if aerr, ok := err.(awserr.Error); ok {
-		return aerr.Code() == s3.ErrCodeBucketAlreadyExists
+		return aerr.Code() == s3.ErrCodeBucketAlreadyExists ||
+			aerr.Code() == s3.ErrCodeBucketAlreadyOwnedByYou
 	}
 
 	return false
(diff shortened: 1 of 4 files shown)


josketres:
Looks good to me. 

You may also want to test if the metadata factory change is compatible with all possible values for tag. According to the [json package documentation](https://golang.org/pkg/encoding/json/):

```
// Field appears in JSON as key "myName".
Field int `json:"myName"`

// Field appears in JSON as key "myName" and
// the field is omitted from the object if its value is empty,
// as defined above.
Field int `json:"myName,omitempty"`

// Field appears in JSON as key "Field" (the default), but
// the field is skipped if empty.
// Note the leading comma.
Field int `json:",omitempty"`

// Field is ignored by this package.
Field int `json:"-"`

// Field appears in JSON as key "-".
Field int `json:"-,"`
```

josketres:
@ajscholl 
Can you also fix your local git config for this repo?

```bash
git config user.name ajscholl
git config user.email foo@example.com
```

You may also want to amend the author of your commit.

PR #205, opened 2020-03-09
fixtures: add documentation;

bogdanfinn:


examples/gosoline-fixture-loading/main.go
```diff
@@ -0,0 +1,105 @@
+//+build fixtures
+
+package main
+
+import (
+	"github.com/applike/gosoline/pkg/application"
+	db_repo "github.com/applike/gosoline/pkg/db-repo"
```
applike-ss: please try not to add these as an alias, they are not required. Also look around in all files that have been touched for unnecessary aliases

pkg/fixtures/writer_plain_mysql.go
```diff
@@ -0,0 +1,81 @@
...
+	"github.com/applike/gosoline/pkg/cfg"
+	"github.com/applike/gosoline/pkg/db"
+	"github.com/applike/gosoline/pkg/mon"
+	"reflect"
+)
+
+type PlainMySqlFixtureValue []interface{}
+
+type PlainMySqlFixture struct {
+	TableName string
```
applike-ss: usually this kind of metadata we set for the whole fixture writer instance and not per fixture (metadata that applies to all fixtures for this writer)

pkg/fixtures/writer_plain_mysql.go
```diff
@@ -0,0 +1,81 @@
...
+	Columns   []string
```
applike-ss: should we add these as metadata to the factory instead too? i assume the columns are always the same for each fixture of a table?

test/fixtures_mysql_test.go
```diff
@@ -31,11 +62,11 @@ var TestModelMetadata = db_repo.Metadata{
 	},
 }
 
-func mysqlTestFixtures() []*fixtures.FixtureSet {
+func ormMysqlTestFixtures() []*fixtures.FixtureSet {
```
applike-ss: actually i think, i would split the test.
This is due to testing different implementations.
Think of it like: we don't test all implementations in one file, so we should have each test in a different file.

docs/fixtures/readme.md
```diff
@@ -0,0 +1,75 @@
...
+	Fixtures []interface{}
+}
+```
+You can easily define multiple fixtures to different destinations in one file and enable or disable them for each FixtureWriter.
+
+* Currently there are 4 different FixtureWriterFactories implemented to load fixtures. 
+    * `DynamoDbFixtureWriterFactory`
+    * `DynamoDbKvStoreFixtureWriterFactory` 
+    * `MysqlOrmFixtureWriterFactory` 
+    * `PlainMySqlFixtureWriterFactory`
```
applike-ss: should we adjust this name to MysqlPlainFixtureWriterFactory?

docs/fixtures/readme.md
```diff
@@ -0,0 +1,75 @@
...
+You can easily define multiple fixtures to different destinations in one file and enable or disable them for each FixtureWriter.
+
+* Currently there are 4 different FixtureWriterFactories implemented to load fixtures. 
+    * `DynamoDbFixtureWriterFactory`
+    * `DynamoDbKvStoreFixtureWriterFactory` 
+    * `MysqlOrmFixtureWriterFactory` 
+    * `PlainMySqlFixtureWriterFactory` 
+
+## Quick Usage
+* During the creation of you Application make sure to pass the `WithFixtures` option and provide fixtures as an argument of type `[]*fixtures.FixtureSet`
```
applike-ss: please correct to `During the creation of your Application...`

pkg/fixtures/writer_plain_mysql.go
bogdanfinn: Done.

pkg/fixtures/writer_plain_mysql.go
bogdanfinn: Done.

docs/fixtures/readme.md
bogdanfinn: Yeah. true

examples/gosoline-fixture-loading/main.go
```diff
@@ -0,0 +1,102 @@
...
+		},
+		{
+			Enabled: true,
+			Writer: fixtures.MySqlPlainFixtureWriterFactory(fixtures.MySqlPlainMetaData{
+				TableName: "plain_fixture_example",
+				Columns:   []string{"id", "name"},
+			}),
+			Fixtures: []interface{}{
+				&fixtures.MySqlPlainFixtureValues{1, "testName1"},
+				&fixtures.MySqlPlainFixtureValues{2, "testName2"},
```
applike-ss: To be consisten, we should write MySql always as Mysql

pkg/fixtures/writer_plain_mysql.go
```diff
@@ -0,0 +1,78 @@
...
+		m.logger.Info(fmt.Sprintf("affected rows while fixture loading: %d", ar))
+	}
+
+	m.logger.Infof("loaded %d plain mysql fixtures", len(fs.Fixtures))
+
+	return nil
+}
+
+func (m *mySqlPlainFixtureWriter) buildSql(values *MySqlPlainFixtureValues) (string, []interface{}, error) {
+	insertBuilder := squirrel.Replace(m.metaData.TableName).PlaceholderFormat(squirrel.Question).Columns(m.metaData.Columns...).Values(*values...)
```
applike-ss: from the last commits it looks like logic changed a bit, but i guess you checked it still works properly?
I would like to see this line to become a bit shorter, how about this:
```suggestion
	insertBuilder := squirrel.Replace(m.metaData.TableName).
	                          PlaceholderFormat(squirrel.Question).
	                          Columns(m.metaData.Columns...).
	                          Values(*values...)
```

test/fixtures_mysql_test.go
bogdanfinn: wont do it because of the "boot mysql multiple times during testing"-argument

PR #208, opened 2020-03-09
stream: added multi producer based on std lib multiwriter

willemschots:
Adds a multi producer that send input to multiple streams. Based on [io.MultiWriter ](https://golang.org/pkg/io/#MultiWriter)

### DIFF
--- a/pkg/stream/multi_producer.go
+++ b/pkg/stream/multi_producer.go
@@ -1,0 +1,33 @@
+package stream
+
+import (
+	"context"
+)
+
+type multiProducer []Producer
+
+func NewMultiProducer(producers ...Producer) Producer {
+	return multiProducer(producers)
+}
+
+func (mp multiProducer) WriteOne(ctx context.Context, model interface{}, attributeSets ...map[string]interface{}) error {
+	for _, p := range mp {
+		err := p.WriteOne(ctx, model, attributeSets...)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (mp multiProducer) Write(ctx context.Context, models interface{}, attributeSets ...map[string]interface{}) error {
+	for _, p := range mp {
+		err := p.Write(ctx, models, attributeSets...)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
--- a/pkg/stream/multi_producer_test.go
+++ b/pkg/stream/multi_producer_test.go
@@ -1,0 +1,140 @@
+package stream_test
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"github.com/applike/gosoline/pkg/stream"
+	"github.com/applike/gosoline/pkg/stream/mocks"
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+type mockConf struct {
+	err        error
+	expectCall bool
+}
+
+func TestMultiProducerTestCallingOfChildren(t *testing.T) {
+	testErr := errors.New("testErr")
+
+	tt := []struct {
+		name      string
+		mocks     []mockConf
+		expectErr error
+	}{
+		{
+			name:      "none",
+			mocks:     []mockConf{},
+			expectErr: nil,
+		},
+		{
+			name: "one success",
+			mocks: []mockConf{
+				{nil, true},
+			},
+			expectErr: nil,
+		},
+		{
+			name: "all success",
+			mocks: []mockConf{
+				{nil, true},
+				{nil, true},
+				{nil, true},
+			},
+			expectErr: nil,
+		},
+		{
+			name: "last fails",
+			mocks: []mockConf{
+				{nil, true},
+				{nil, true},
+				{testErr, true},
+			},
+			expectErr: testErr,
+		},
+		{
+			name: "first fails",
+			mocks: []mockConf{
+				{testErr, true},
+				{nil, false},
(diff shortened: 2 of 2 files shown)


pkg/stream/multi_producer.go
```diff
@@ -0,0 +1,31 @@
...
+}
+
+func (mp multiProducer) WriteOne(ctx context.Context, model interface{}, attributeSets ...map[string]interface{}) error {
+	for _, p := range mp {
+		err := p.WriteOne(ctx, model, attributeSets...)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
```
applike-ss: please always leave an empty line before a return except it is the only statement in a block

pkg/stream/multi_producer_test.go
```diff
@@ -0,0 +1,198 @@
...
+func newParams(context context.Context, model interface{}, sets []map[string]interface{}) []interface{} {
+	params := []interface{}{context, model}
+	for _, set := range sets {
+		params = append(params, set)
+	}
+	return params
+}
+
+func TestMultiProducerWriteOne(t *testing.T) {
+	type mockConf struct {
```
applike-ss: please don't create types in functions

pkg/stream/multi_producer_test.go
```diff
@@ -0,0 +1,198 @@
...
+		err        error
+		expectCall bool
+	}
+
+	testErr := errors.New("testErr")
+
+	tt := []struct {
```
applike-ss: please don't create inline structs

pkg/stream/multi_producer_test.go
```diff
@@ -0,0 +1,198 @@
...
+				if mock.expectCall {
+					// hack around variadic parameter limit of mocking package
+					params := newParams(context.Background(), msg, attributeSets)
+					p.On("WriteOne", params...).Return(mock.err)
+				}
+				children[i] = p
+			}
+
+			// When
+			producer := stream.NewMultiProducer(children...)
```
applike-ss: this can be unified for both test cases

pkg/stream/multi_producer.go
willemschots: Will do

pkg/stream/multi_producer_test.go
willemschots: I'll restructure the tests a bit to get rid of the duplication.

This is a fairly standard pattern for table driven tests in Go though, since the structs are generally local to the specific test. So I don't really see the harm in using it.

applike-ss: but you using it in both tests, so why not remove code duplication here?

willemschots: Those are seperate issues right, not using inline structs in general vs removing duplicate code. 
I'll push a new solution in a bit.

pkg/stream/multi_producer.go
```diff
@@ -0,0 +1,31 @@
+package stream
+
+import (
+	"context"
+)
+
+type multiProducer []Producer
```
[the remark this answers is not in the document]
willemschots: Worth considering. Inside the stream package, would you expect to be able to call `WriteOne` and `Write` on producers? Naming it multiProducer makes it a little more explicit imo.

pkg/stream/multi_producer_test.go
```diff
@@ -0,0 +1,139 @@
...
+
+func newParams(context context.Context, model interface{}, sets []map[string]interface{}) []interface{} {
+	params := []interface{}{context, model}
+	for _, set := range sets {
+		params = append(params, set)
+	}
+	return params
+}
+
+type mockConf struct {
```
applike-ss: type definitions usually live above functions in our code

pkg/stream/multi_producer_test.go
```diff
@@ -0,0 +1,139 @@
...
+	"context"
+	"errors"
+	"fmt"
+	"github.com/applike/gosoline/pkg/stream"
+	"github.com/applike/gosoline/pkg/stream/mocks"
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+func newMessage() interface{} {
```
applike-ss: please order funtions in execution order

pkg/stream/multi_producer_test.go
applike-ss: your last push didn't contain a fix here?

pkg/stream/multi_producer_test.go
```diff
@@ -0,0 +1,139 @@
...
+				{testErr, true},
+				{nil, false},
+				{nil, false},
+			},
+			expectErr: testErr,
+		},
+	}
+
+	for _, tc := range tt {
+		t.Run(fmt.Sprintf("writeOne_%s", tc.name), func(t *testing.T) {
```
applike-ss: now we are not testing the batch write anymore :-(

pkg/stream/multi_producer_test.go
willemschots: The testing table is fully local to `TestMultiProducerTestCallingOfChildren` at the moment. So there's no need to define a type for it.

pkg/stream/multi_producer_test.go
willemschots: See line 125

willemschots: :D

Issue #196, opened 2020-03-02
fixture-loader: purge mysql


josketres:
Whenever fixture loader runs, we want to purge specific (configured) database tables.

This is a convenience feature for development only. 

Note: We should be careful not to purge databases in production systems.

bogdanfinn:
As discussed we only purge destionations we also would write fixtures to.
Means for example for mysql fixtures we only truncate the tables we are also loading data into.

PR #214, opened 2020-03-11
purge-before-fixture-loading

bogdanfinn:


pkg/fixtures/writer_mysql_plain.go
```diff
@@ -19,21 +19,38 @@ type mysqlPlainFixtureWriter struct {
...
 		dbClient := db.NewClient(config, logger)
-		return NewMysqlPlainFixtureWriterWithInterfaces(logger, dbClient, metaData)
+		purger := newMysqlPurger(config, logger, metaData.TableName)
+
+		return NewMysqlPlainFixtureWriterWithInterfaces(logger, dbClient, metaData, purger)
 	}
 }
 
-func NewMysqlPlainFixtureWriterWithInterfaces(logger mon.Logger, dbClient db.Client, metaData *MysqlPlainMetaData) FixtureWriter {
+func NewMysqlPlainFixtureWriterWithInterfaces(logger mon.Logger, dbClient db.Client, metaData *MysqlPlainMetaData, purger *mysqlPurger) FixtureWriter {
```
josketres: make this private to be consistent with the other writer.

applike-ss:
closes #196

pkg/fixtures/writer_mysql_plain.go
applike-ss: but then we can not unit test that, right?

PR #219, opened 2020-03-12
fixtures: added purging for dynamodb and dynamodbKvstore implementation

### LINKED ISSUE #216: fixture-loader: purge dynamo db
We already have https://github.com/applike/gosoline/issues/196 for mysql.

Now we want also purge dynamo db.

applike-ss:
closes #216

### DIFF
--- a/pkg/fixtures/writer_ddb.go
+++ b/pkg/fixtures/writer_ddb.go
@@ -7,14 +7,17 @@ import (
 	"github.com/applike/gosoline/pkg/mon"
 )
 
+type ddbRepoFactory func() ddb.Repository
+
 type dynamoDbFixtureWriter struct {
-	logger mon.Logger
-	repo   ddb.Repository
+	logger  mon.Logger
+	factory ddbRepoFactory
+	purger  *dynamodbPurger
 }
 
 func DynamoDbFixtureWriterFactory(settings *ddb.Settings) FixtureWriterFactory {
 	return func(config cfg.Config, logger mon.Logger) FixtureWriter {
-		repo := ddb.NewRepository(config, logger, &ddb.Settings{
+		settings := &ddb.Settings{
 			ModelId:    settings.ModelId,
 			AutoCreate: true,
 			Main: ddb.MainSettings{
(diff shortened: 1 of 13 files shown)


pkg/fixtures/writer_ddb.go
```diff
@@ -23,22 +26,35 @@ func DynamoDbFixtureWriterFactory(settings *ddb.Settings) FixtureWriterFactory {
 				WriteCapacityUnits: 1,
 			},
 			Global: settings.Global,
-		})
+		}
+
+		factory := func() ddb.Repository {
```
bogdanfinn: i do not get the reason why we need a factory here

pkg/fixtures/writer_ddb_kvstore.go
```diff
@@ -3,49 +3,66 @@ package fixtures
...
 	return func(config cfg.Config, logger mon.Logger) FixtureWriter {
-		store := kvstore.NewDdbKvStore(config, logger, &kvstore.Settings{
-			AppId: cfg.AppId{
-				Project:     modelId.Project,
-				Environment: modelId.Environment,
-				Family:      modelId.Family,
-				Application: modelId.Application,
-			},
-			Name: modelId.Name,
+		factory := func() kvstore.KvStore {
```
bogdanfinn: i do not get the reason why we need a factory here

test/fixtures_dynamodb_test.go
```diff
@@ -59,12 +106,42 @@ func (s FixturesDynamoDbSuite) TestDynamoDb() {
 		TableName: aws.String("gosoline-test-integration-test-test-application-testModel"),
 	})
 
+	err = loader.Load(dynamoDbEnabledPurgeFixtures())
```
bogdanfinn: this should happen after the assert of `gio`? even if it works the readability is strange because you are already loading new fixtures even if you have not asserted that the previous ones are loaded correctly?

pkg/fixtures/writer_ddb.go
applike-ss: solved in private conversation

pkg/fixtures/writer_ddb_kvstore.go
applike-ss: solved in private conversation

test/fixtures_dynamodb_test.go
applike-ss: you are perfectly right! Will fix it

test/fixtures_dynamodb_test.go
```diff
@@ -231,6 +232,19 @@ func (s FixturesDynamoDbSuite) TestDynamoDbKvStoreWithPurge() {
...
+	gio, err = s.db.GetItem(&dynamodb.GetItemInput{
+		Key: map[string]*dynamodb.AttributeValue{
+			"key": {
+				S: aws.String("Ash"),
+			},
+		},
+		TableName: aws.String("gosoline-test-integration-test-test-application-kvstore-testModel"),
+	})
+
+	// should have created the item
```
bogdanfinn: the comment might be confusing.

PR #223, opened 2020-03-13
stream: fix concurrent write in file output;

josketres:
Closes #162

### DIFF
--- a/pkg/stream/output_file_test.go
+++ b/pkg/stream/output_file_test.go
@@ -1,0 +1,46 @@
+package stream_test
+
+import (
+	"context"
+	"fmt"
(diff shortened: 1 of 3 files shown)


pkg/stream/output_file_test.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/stretchr/testify/assert"
+	"io/ioutil"
+	"os"
+	"regexp"
+	"sync"
+	"testing"
+)
+
+func TestOutputFile_ConcurrentWrite(t *testing.T) {
+
```
applike-ss: please remove empty line in beginning of block

PR #224, opened 2020-03-13
redis: added SetNX method;

ajscholl:


pkg/redis/redis_client.go
```diff
@@ -27,6 +27,7 @@ type Client interface {
 	Expire(key string, ttl time.Duration) (bool, error)
 	FlushDB() (string, error)
 	Set(string, interface{}, time.Duration) error
+	SetNX(string, interface{}, time.Duration) (bool, error)
```
applike-ss: we decided that we want to have parameter names in interfaces, even though we did not yet add them everywhere. Please adhere to this rule when you add code. Maybe you could also quickly add the missing parameter names in the other functions?

pkg/tracing/message_encode_handler_test.go
```diff
@@ -67,4 +67,6 @@ func TestMessageWithTraceEncoder_Decode_Warning(t *testing.T) {
 	assert.NoError(t, err)
 	assert.Contains(t, decodedAttributes, "traceId")
 	assert.Nil(t, trace)
+
+	logger.AssertExpectations(t)
```
applike-ss: :+1:

pkg/redis/redis_client.go
ajscholl: Should be much better now.

PR #226, opened 2020-03-16
sqs, stream: added message deduplication id to forwarded attributes;

### LINKED ISSUE #220: Message deduplication is broken
As a developer I am seeing one message object which is send to SQS. If I then enable content based deduplication I expect to only ever send one such object to my queue (in a certain time frame). What is happening is however a different story. Gosoline collects some attributes and stores an object in the queue with attribute and body fields. Thus, the traceId from the context can end up in my message. However, that ID is random, so I will always get a different message from SQSs perspective and content based deduplication no longer works.

To fix this, we should send a message deduplication ID to SQS together with the message instead of letting SQS handle this itself. We could also move the attributes to SQS attributes, but that does sound a lot more complicated.

ajscholl:
Closes #220

### DIFF
--- a/pkg/db/client.go
+++ b/pkg/db/client.go
@@ -1,6 +1,7 @@
 package db
 
 import (
+	"context"
 	"database/sql"
 	"errors"
 	"fmt"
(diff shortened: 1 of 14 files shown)


j4k4:
Not sure if I'm a fan of this.. 
Could still be better just to use the sqs package?

PR #228, opened 2020-03-18
test: make docker expiration period configurable;

josketres:
**Summary:**

* In some cases, the tests are not cleaning up the docker containers they started. 
* This causes problems when trying to run the tests again (locally or on a CI environment).
* Set the default expiration period to 1 minute (was previously 1 hour)
* Make it configurable so that it can be adapted if needed e.g. for long-running tests for example.
* Add more logging to investigate failing tests due to docker problems in the CI.

image

### DIFF
--- a/pkg/test/docker.go
+++ b/pkg/test/docker.go
@@ -23,6 +23,7 @@ type containerConfig struct {
 	PortBindings portBinding
 	HealthCheck  func() error
 	PrintLogs    bool
+	ExpireAfter  time.Duration
 }
 
 type dockerRunner struct {
@@ -60,6 +61,10 @@ func (d *dockerRunner) Run(name string, config containerConfig) {
 
 	containerName := d.getContainerName(name)
 
+	logger := d.logger.WithFields(map[string]interface{}{
+		"container": containerName,
+	})
+
 	bindings := make(map[docker.Port][]docker.PortBinding)
 	for containerPort, hostPort := range config.PortBindings {
 		bindings[docker.Port(containerPort)] = []docker.PortBinding{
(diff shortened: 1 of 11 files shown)


pkg/test/docker.go
```diff
@@ -83,12 +88,14 @@ func (d *dockerRunner) Run(name string, config containerConfig) {
...
 	}
 
-	err = resource.Expire(60 * 60)
+	err = resource.Expire(uint(config.ExpireAfter.Seconds()))
 
 	if err != nil {
 		panic(fmt.Errorf("could not expire %s container: %w", containerName, err))
 	}
 
+	logger.Infof("container will expire after %s", config.ExpireAfter)
```
applike-ss:
```suggestion
	logger.WithFields(map[string]interface{}{
		"expire_after": config.ExpireAfter.Seconds(),
	}.Info("set container expiry")
```

test/test_configs/config.mysql.test.yml
```diff
@@ -6,3 +6,4 @@ mocks:
     debug: false
     dbName: myDbName
     version: "8"
+    expire_after: 1m30s
```
applike-ss: is this really required or is it just a test?

josketres: I'm just testing that it works.

pkg/test/docker.go
josketres: ✅

test/test_configs/config.mysql.test.yml
josketres: 1m should be more than enough for this test

PR #229, opened 2020-03-18
Smaller kernel fixes separated from !140

### LINKED ISSUE #201: Race condition in kernel startup
When the kernel is starting up, all modules are ranged over and the the Run function of those modules is called.
However, due to the current nature of the modules there is no reporting back when all modules are in a usable state.
This can be easily seen when doing integration tests against the api server (repeat them multiple times and see how your client suddenly fails to get a reponse from the given port).
Also this can be seen in the logs where you see that the kernel is up and running, while no module has yet reported to be running.

ajscholl:
Closes #201 

This MR includes the smaller fixes from !140 without yet introducing the staged kernel.

### DIFF
--- a/pkg/kernel/kernel.go
+++ b/pkg/kernel/kernel.go
@@ -80,7 +80,16 @@ func (k *kernel) Add(name string, module Module) {
 		IsRunning: false,
 	}
 
-	k.modules.Store(name, state)
+	if _, didExist := k.modules.LoadOrStore(name, state); didExist {
+		// if we overwrite an existing module, the module count will be off and the application will hang while waiting
+		// until stage.moduleCount modules have booted.
+		k.logger.Panicf(
+			errors.New("module must not be redeclared"),
+			"failed to add new module %s: module exists",
+			name,
+		)
+	}
+
 	k.moduleCount++
 }
 
--- a/pkg/kernel/kernel.go
+++ b/pkg/kernel/kernel.go
@@ -293,6 +293,7 @@ func (k *kernel) checkRunningEssentialModules() {
 }
 
 func (k *kernel) doCheckRunningEssentialModules() {
+	_ = <-k.running
 	if !k.isRunning {
 		return
 	}
@@ -323,6 +324,7 @@ func (k *kernel) doCheckRunningEssentialModules() {
 }
 
 func (k *kernel) checkRunningForegroundModules() {
+	_ = <-k.running
 	if !k.isRunning {
 		return
 	}
--- a/pkg/mon/metric_daemon.go
+++ b/pkg/mon/metric_daemon.go
@@ -40,13 +40,53 @@ type cwDaemon struct {
 	logger   Logger
 	settings *MetricSettings
 
-	channel chan *MetricDatum
+	channel *metricChannel
 	ticker  *time.Ticker
 	writers []MetricWriter
 
 	batch          map[string]*BatchedMetricDatum
 	defaults       map[string]*MetricDatum
 	dataPointCount int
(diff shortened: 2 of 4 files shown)


pkg/kernel/kernel.go
```diff
@@ -80,7 +80,16 @@ func (k *kernel) Add(name string, module Module) {
 		IsRunning: false,
 	}
 
-	k.modules.Store(name, state)
+	if _, didExist := k.modules.LoadOrStore(name, state); didExist {
+		// if we overwrite an existing module, the module count will be off and the application will hang while waiting
+		// until stage.moduleCount modules have booted.
+		k.logger.Panicf(
+			errors.New("module must not be redeclared"),
+			"Failed to add new module %s: module exists. Each module name needs to be unique",
```
josketres: This may be too much nit-picking, but I noticed that most of the logs and error messages in Gosoline are written in lowercase.
So to keep consistency there I suggest doing the same.

josketres: Also the text "Each module name needs to be unique" is redundant. We could just remove it.

pkg/mon/metric_daemon.go
```diff
@@ -131,6 +134,25 @@ func (d *cwDaemon) AddDefaults(data ...*MetricDatum) {
 	}
 }
 
+func (d *cwDaemon) emptyChannel() {
```
josketres: Can you try to close the channel instead of doing this sentinel workaround?

```go
// here
close(d.channel)

// in the Run method
   case closed, dat := <-d.channel:
      if closed {
         return
      }
      d.append(dat)					
```

ajscholl: I think if we close the channel and there is still someone writing to it, we get a panic in that thread. As we don't want metrics to change the behavior of the program, I am not sure this is something we want to risk just because someone has a process still running (especially before we have a kernel with multiple stages)

pkg/kernel/kernel.go
```diff
@@ -284,6 +293,7 @@ func (k *kernel) checkRunningEssentialModules() {
 }
 
 func (k *kernel) doCheckRunningEssentialModules() {
+	_ = <-k.running
```
josketres: What is the purpose of this?

ajscholl: It actually blocks until the kernel was running at some point. Otherwise you have the following scenario:

1. Module A starts
2. Module A exists, checks for running forground modules
3. Kernel still launching module B and C
4. Module A sees the kernel is not (yet) running, assumes the kernel is no longer running
5. Kernel sets running to true, but module A did already exist. You now have a kernel without any running foreground modules which did not exit

Instead, if module A waits until the kernel closes the running channel, the check correctly sees that there are no running foreground modules anymore and stops the kernel

pkg/kernel/kernel.go
ajscholl: Will change it

pkg/mon/metric_daemon.go
josketres: I agree, but.
The `channel` is a private field of our daemon. 
To prevent this we should add a check before we write in the channel.
(check if it is closed)

I think closing the channel is just a more elegant solution.
 It is the use case that feature was made for (channel closing).

pkg/kernel/kernel.go
josketres: I understand. But this method is called inside of an infinite for loop.
So, even if this happens, it will be called some milliseconds after and the kernel will be stoped properly. Am I right?

PR #230, opened 2020-03-19
S3 - Keep track of read files + option to delete files in writer

rchernanko:


pkg/parquet/reader.go
```diff
@@ -30,6 +30,8 @@ type ReadResult map[string]interface{}
 type ReadResults []ReadResult
 type ResultCallback func(progress Progress, results interface{}) (bool, error)
 
+var S3FilesToDelete []string
```
rchernanko: I don't like this, but essentially need a way of the writer being able to access the files to delete. Alternatively, I could add the file name onto each spend event item, and then keep track of the files to delete that way... Then I could just pass in a slice to the DeleteFilesInS3...

pkg/parquet/reader.go
```diff
@@ -156,6 +158,7 @@ func (r *s3Reader) ReadDateAsync(ctx context.Context, datetime time.Time, target
 				stop = true
 				return
 			}
+			S3FilesToDelete = append(S3FilesToDelete, file)
```
floelm: this should be rather something like `readS3Files` and then depending on the options delete the fiiles on read date right?

rchernanko: That is a very good point. Will update the name

pkg/parquet/reader.go
rchernanko: will discuss with Flo in chat

pkg/parquet/writer.go
```diff
@@ -185,3 +193,37 @@ func makeTags(tags map[string]string) []*s3.Tag {
 
 	return s3Tags
 }
+
+func (w *s3Writer) DeleteReadFiles(ctx context.Context) error {
```
floelm: DeleteReadFiles and DeleteWrittenFiles look quite similar, can we reuse something?

Issue #225, opened 2020-03-13
test: simplify hostname in test setup


josketres:
- Currently, the hostname is "static".  Configured in `test.config.yaml`.
- Use `localhost` as a default hostname.
- Allow changing the hostname for CI.

PR #233, opened 2020-03-26
Test integration tests random ports

### LINKED ISSUE #218: test: support random ports for test components
The ports for the integration tests are static.
This prevents us from running multiple tests at the same time on one CI instance.
We should instead randomize the docker assigned port, so we can run as many tests as we have resources for.

applike-ss:
closes #218

### DIFF
--- a/pkg/test/test.go
+++ b/pkg/test/test.go
@@ -4,13 +4,17 @@ import (
 	"fmt"
 	"github.com/applike/gosoline/pkg/cfg"
 	"github.com/applike/gosoline/pkg/mon"
+	"github.com/ory/dockertest"
+	"github.com/ory/dockertest/docker"
+	"strconv"
 	"sync"
 	"time"
 )
 
 type mockComponent interface {
-	Boot(config cfg.Config, runner *dockerRunner, settings *mockSettings, name string)
-	Start()
+	Boot(config cfg.Config, logger mon.Logger, runner *dockerRunner, settings *mockSettings, name string)
+	Start() error
+	Ports() map[string]int
 }
 
 type mockSettings struct {
@@ -20,11 +24,20 @@ type mockSettings struct {
 	ExpireAfter time.Duration `cfg:"expire_after" default:"60s"`
 }
 
+type baseComponent struct {
+	logger mon.Logger
+	runner *dockerRunner
+	ports  map[string]int
+	name   string
+}
+
 type Mocks struct {
+	lck          sync.Mutex
 	waitGroup    sync.WaitGroup
 	dockerRunner *dockerRunner
 	components   map[string]mockComponent
 	logger       mon.Logger
+	bootFailed   bool
 }
 
 func newMocks() *Mocks {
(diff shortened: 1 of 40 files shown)


test/test_configs/config.fixtures_mysql.test.yml
```diff
@@ -5,7 +5,7 @@ app_name: fixture-loader
 
 db_drivername: "mysql"
 db_hostname: "172.17.0.1"
```
bogdanfinn: `GOSOLINE_MYSQL_MYSQL_HOST` ?

applike-ss: i would say, we do this when we need it, so with #225

j4k4:
first thoughts:

- the config string augment still feels a little bit misused
- Don't you get a race condition with your current get free port method? timeframe between finding and using?
- Does this also mean, that every test needs a different component name? Would 2 tests with the same dynamodb component name overwrite each others env variable?
- try to avoid using panics.. return errors and handle them appropriately

applike-ss:
@j4k4 
1st point: please feel free to suggest how to fix this
2nd point: yes, but it is extremely unlikely that sone grabs the port meanwhile. If you mind about it, how about fixing it in a future issue?
3rd point: yes that is true and has already been an issue beforehand
4th point: ok, will do

j4k4:
I would like to try to evaluate a different approach:
1. make the port configurable again with '0' as default..
2. set '0' as host port in the binding on starting the container
3. the resource returned by the `d.pool.RunWithOptions` contains the port which got assigned by the os in the network settings
4. return component info like ports in the mocks object after boot `mocks := pkgTest.Boot`
5. use this info to set config values in the app with the option `application.WithConfigMap` or use them directly to create a client.

This should have no race conditions, no _hacky_ way to find a free port and every test could have the same component name? Feel free to correct me here.

@josketres What is your take on this?

josketres:
I agree with the approach suggested by @j4k4 . It is much more elegant (less hacky) and solves the race condition problem. 

I wasn't aware that `docker run -P 0:80` would publish the containers port 80 to a free port on the host machine.

applike-ss:
looks like a clean approach, will try to integrate that

pkg/test/test.go
```diff
@@ -90,15 +102,23 @@ func (m *Mocks) runComponent(component mockComponent) {
 	m.waitGroup.Add(1)
 	go func() {
 		defer m.waitGroup.Done()
-		component.Start()
+
+		err := component.Start()
+		if err != nil {
+			m.logger.Fatalf(err, "failed to start component")
```
josketres: This is going to call `os.Exit(1)` which will completely skip the deferred shutdown of the containers.

applike-ss: you're right, but since the components by default do expire after 1 minute and due to the fact that we do use dynamic port allocation i didn't really see an issue here. But i am fine to use Panicf instead

PR #234, opened 2020-03-26
test: docker containers not being purged due to race condition;

josketres:
When starting 2 test components concurrently, there's an issue with the cleanup.

* Starting 1 succeeds.
* Starting 2 fails. We panic.
* Container 1 was not yet added to the cleanup list.
* Shutdown/cleanup is called because it was deferred in the test.
* Container 1 was not stopped and maybe not even marked for expiration, so it will hang there forever.

### DIFF
--- a/pkg/test/docker.go
+++ b/pkg/test/docker.go
@@ -27,11 +27,11 @@ type containerConfig struct {
 }
 
 type dockerRunner struct {
-	pool           *dockertest.Pool
-	resources      []*dockertest.Resource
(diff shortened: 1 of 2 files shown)


j4k4:
Isn't one of the problems here, that we're using panics?  
Commented in #233 as well, that instead of panics, errors should be returned and handled accordingly.

josketres:
@j4k4 Yes, I think that is actually one of the root causes. I'll try and remove the panic calls in this PR.

PR #235, opened 2020-03-30
S3 parquet deleter

ajscholl:
Added S3 parquet deleter (and renamer). Using this, you can rewrite a bunch of parquet files in S3 and later move them to a temporary location before deleting them (or deleting them right away if you feel lucky).

Commits will later be squashed


j4k4:
+1

PR #236, opened 2020-03-30
test: provide ports + less panics;

josketres:
**Summary:**

- Rework approach to provide ports for test components using same approach we used to provide the clients.
- Move some docker-specific logic back into `docker.go`.
- Remove some panics in the boot process.
- Rework docker containers cleanup to avoid zombie containers due to race-conditions.


pkg/test/docker.go
```diff
@@ -43,23 +44,25 @@ func newDockerRunner() *dockerRunner {
 
 	pool.MaxWait = 2 * time.Minute
 
-	resources := make([]*dockertest.Resource, 0)
+	containers := make([]string, 0)
```
applike-ss: is there any reason not to save a pointer to a dockertest.Resource? i find this quite nice when debugging to be able to see all properties of a container. Especially when adding func functionality in the integration test of course

pkg/test/docker.go
```diff
@@ -84,41 +87,53 @@ func (d *dockerRunner) Run(name string, config containerConfig) (*dockertest.Res
 	})
 
 	if err != nil {
-		return nil, fmt.Errorf("could not start %s container: %w", containerName, err)
+		return fmt.Errorf("could not start %s container: %w", containerName, err)
```
applike-ss:
```suggestion
		return fmt.Errorf("could not start container %s: %w", containerName, err)
```

pkg/test/docker.go
```diff
@@ -84,41 +87,53 @@ func (d *dockerRunner) Run(name string, config containerConfig) (*dockertest.Res
...
 	}
 
 	err = resource.Expire(uint(config.ExpireAfter.Seconds()))
 
 	if err != nil {
-		return nil, fmt.Errorf("could not expire %s container: %w", containerName, err)
+		return fmt.Errorf("could not expire %s container: %w", containerName, err)
```
applike-ss:
```suggestion
		return fmt.Errorf("could not expire container %s: %w", containerName, err)
```

pkg/test/docker.go
```diff
@@ -43,23 +44,25 @@ func newDockerRunner() *dockerRunner {
...
+		logger:     logger,
+		containers: containers,
 	}
 }
 
-func (d *dockerRunner) Run(name string, config containerConfig) (*dockertest.Resource, error) {
+func (d *dockerRunner) Run(name string, config containerConfig) error {
 	containerName := d.getContainerName(name)
 
+	d.markForCleanup(containerName)
```
applike-ss: why would we do this before the container did even start successfully?

pkg/test/docker.go
```diff
@@ -84,41 +87,53 @@ func (d *dockerRunner) Run(name string, config containerConfig) (*dockertest.Res
...
 	}
 
 	logger.Info("container up and running")
 
-	return resource, nil
+	return nil
 }
 
-func (d *dockerRunner) printContainerLogs(resource *dockertest.Resource) {
+func (d *dockerRunner) markForCleanup(containerName string) {
```
applike-ss: please ensure functions are sorted in code execution order

pkg/test/docker.go
```diff
@@ -128,40 +143,40 @@ func (d *dockerRunner) printContainerLogs(resource *dockertest.Resource) {
...
+}
 
-	err := d.pool.Client.Logs(docker.LogsOptions{
-		Container:    containerName,
-		OutputStream: logs,
-		Stdout:       true,
-		Stderr:       true,
-	})
+func (d *dockerRunner) setPortMapping(resource *dockertest.Resource, containerPort string, hostPort *int) error {
+	dockerPort := docker.Port(containerPort)
```
applike-ss: maybe use resource.GetPort(containerPort)?

pkg/test/docker.go
josketres: By using the container name we can add it to the list even before starting the container. In other words: we are 100% sure that if we started container X that X will be in the list of containers (used later for cleanup).

And if we use the name instead of the resource, then we do not need to save them anymore. If you need it for debugging then you can save them temporarily in a variable.

pkg/test/docker.go
josketres: to make sure it is cleaned up correctly. 
If the container failed to start, then it doesn't need to be removed, so we will ignore the error when trying to remove it.

pkg/test/docker.go
applike-ss: If the container did not start, then there is no need to try and clean it up. Do i not see something here?

pkg/test/docker.go
applike-ss: That sounds like a case to me that should never happen and also it does auto heal, because we do expire the container

pkg/test/docker.go
applike-ss: never mind, just had a look at dockertest. Due to the way they implemented `RunWithOptions` there is no better way.

pkg/test/docker.go
applike-ss: never mind, just had a look at dockertest. Due to the way they implemented `RunWithOptions` there is no better way.

PR #240, opened 2020-04-01
stream: add in_memory input;

josketres:
This input is very useful when writing integration tests.


josketres:
Update:

This PR is on hold. We'll revisit it next week and see if a better solution was found. If not we can just close it.

The main pain point is the `stream. SendToInMemoryInput` function.

PR #242, opened 2020-04-02
deployments: added fifo capability to tf queue creation

applike-ss:
deployments: added fifo capability to tf queue creation


deployments/terraform/modules/sqs/main.tf
```diff
@@ -1,11 +1,18 @@
+locals {
+  queueName           = var.queueName
+  fifo                = length(regexall(".*\\.fifo$", var.queueName)) > 0
```
bogdanfinn, line 3: hm not sure if i would go this way to detect the queue type out of the name.
i would have done something like build the queue name out of the "base name" and the queue type. Like:

queueName = "myAwesomeQueue"
fifo = true

==>. myAwesomeQueue.fifo

Because otherwise you are depending on a strict naming of .fifo to use fifo queues. not sure if this is how it should be?

PR #243, opened 2020-04-02
test: provide hostname mapping;

### LINKED ISSUE #225: test: simplify hostname in test setup

josketres:
Closes #225 

In an ideal world, we would always use localhost for Docker-based tests. 

But this is not the case:

- In CI we use docker-in-docker (dind), so we cannot use localhost. We can use either the container ip or the gateway ip for the Bridge network.
- For Mac users this approach does not work. They should use localhost or create a network alias to support the gateway ip adress.

To simplify the usage of the tests package we take care of it, check which address is reacheable and provide the host via `mock.ProvideXxxPort()` methods.

### DIFF
--- a/pkg/test/docker.go
+++ b/pkg/test/docker.go
@@ -7,6 +7,7 @@ import (
 	"github.com/ory/dockertest"
 	"github.com/ory/dockertest/docker"
 	"log"
+	"net"
 	"os"
 	"strconv"
 	"sync"
@@ -17,6 +18,11 @@ type portBinding map[string]string
 
 type portMapping map[string]*int
 
+type hostMapping struct {
+	dialPort *int
+	setHost  *string
+}
+
 type containerConfig struct {
 	Repository   string
 	Tag          string
@@ -24,6 +30,7 @@ type containerConfig struct {
 	Cmd          []string
 	PortBindings portBinding
 	PortMappings portMapping
+	HostMapping  hostMapping
 	HealthCheck  func() error
 	PrintLogs    bool
 	ExpireAfter  time.Duration
@@ -65,7 +72,7 @@ func (d *dockerRunner) Run(name string, config containerConfig) error {
 
 	d.markForCleanup(containerName)
 
-	logger := d.logger.WithFields(map[string]interface{}{
+	logger := d.logger.WithFields(mon.Fields{
 		"container": containerName,
 	})
 
@@ -98,7 +105,7 @@ func (d *dockerRunner) Run(name string, config containerConfig) error {
 		return fmt.Errorf("could not expire container %s: %w", containerName, err)
 	}
 
-	logger.WithFields(map[string]interface{}{
+	logger.WithFields(mon.Fields{
 		"expire_after": config.ExpireAfter,
 	}).Info("set container expiry")
 
@@ -110,11 +117,19 @@ func (d *dockerRunner) Run(name string, config containerConfig) error {
 	}
 
 	err = d.pool.Retry(func() error {
+		return d.setHostMapping(resource, config.HostMapping)
+	})
+
+	if err != nil {
+		return fmt.Errorf("could not set host mapping for container %s: %w", containerName, err)
+	}
+
+	err = d.pool.Retry(func() error {
 		return config.HealthCheck()
 	})
 
 	if err != nil {
-		return fmt.Errorf("could not bring up %s container: %w", containerName, err)
+		return fmt.Errorf("could not check health of container %s: %w", containerName, err)
 	}
 
 	if config.PrintLogs {
@@ -145,9 +160,10 @@ func (d *dockerRunner) setPortMapping(resource *dockertest.Resource, containerPo
 		return err
 	}
 
-	d.logger.WithFields(map[string]interface{}{
-		"container": resource.Container.Name[1:],
-	}).Infof("set port mapping %s:%d", containerPort, port)
+	d.logger.WithFields(mon.Fields{
+		"container":    resource.Container.Name[1:],
+		"port_mapping": fmt.Sprintf("%s:%d", containerPort, port),
+	}).Info("set port mapping")
 
 	*hostPort = port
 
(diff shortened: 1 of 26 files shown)


pkg/test/docker.go
```diff
@@ -180,3 +195,46 @@ func (d *dockerRunner) RemoveAllContainers() {
 		}
 	}
 }
+
+func (d *dockerRunner) setHostMapping(resource *dockertest.Resource, mapping hostMapping) error {
+	timeout := time.Duration(1) * time.Millisecond
+	containerIp := resource.Container.NetworkSettings.Networks["bridge"].IPAddress
+	gatewayIp := resource.Container.NetworkSettings.Networks["bridge"].Gateway
+
+	addresses := []string{containerIp, gatewayIp, "127.0.0.1"}
```
applike-ss, line 204: i think due to the nature of container port exposing, the containerIp will never be reachable under the dialPort unless it is the one-in-a-million case where traffic port and exposed port are actually the same.
If you want to reach the container ip with a given port, it needs to be the one that is exposed afaik. (say 8080 for localstack healthcheck)

pkg/test/docker.go
```diff
@@ -17,13 +18,19 @@ type portBinding map[string]string
 
 type portMapping map[string]*int
 
+type hostMapping struct {
+	dialPort *int
+	setHost  *string
```
applike-ss, line 23: Why not simply `host` and `port`?

pkg/test/docker.go
```diff
@@ -180,3 +195,46 @@ func (d *dockerRunner) RemoveAllContainers() {
 		}
 	}
 }
+
+func (d *dockerRunner) setHostMapping(resource *dockertest.Resource, mapping hostMapping) error {
+	timeout := time.Duration(1) * time.Millisecond
```
applike-ss, line 200: i think this is way too fast. In your mind you may think that we are only handling binary applications so it must be this fast, but what about programs that are not as fast (java, php, ...)?
```suggestion
	timeout := time.Duration(100) * time.Millisecond
```

pkg/test/docker.go
applike-ss: i would have expected our "CI" settable ip to be in here via os.GetEnv

pkg/test/docker.go
```diff
@@ -180,3 +195,46 @@ func (d *dockerRunner) RemoveAllContainers() {
...
+	containerIp := resource.Container.NetworkSettings.Networks["bridge"].IPAddress
+	gatewayIp := resource.Container.NetworkSettings.Networks["bridge"].Gateway
+
+	addresses := []string{containerIp, gatewayIp, "127.0.0.1"}
+	for _, ip := range addresses {
+		address := fmt.Sprintf("%s:%d", ip, *mapping.dialPort)
+		if d.isReachable(address, timeout) {
+			d.logger.WithFields(map[string]interface{}{
```
applike-ss, line 208:
```suggestion
			d.logger.WithFields(mon.Fields{
```

pkg/test/docker.go
```diff
@@ -180,3 +195,46 @@ func (d *dockerRunner) RemoveAllContainers() {
...
+				"container": resource.Container.Name[1:],
```
applike-ss, line 209: why not like this?
```suggestion
				"container": resource.Container.Name,
```

pkg/test/docker.go
```diff
@@ -180,3 +195,46 @@ func (d *dockerRunner) RemoveAllContainers() {
...
+			}).Infof("set host mapping %s", ip)
```
applike-ss, line 210:
```suggestion
			}).
			   WithFields(mon.Fields{
			       "ip": ip
			   }).
			   Info("set host mapping")
```

pkg/test/test.go
```diff
@@ -15,9 +15,9 @@ type mockComponent interface {
 }
 
 type mockSettings struct {
+	Host        string
 	Debug       bool          `cfg:"debug"`
 	Component   string        `cfg:"component"`
-	Host        string        `cfg:"host"`
```
applike-ss, line 20: why not keep it configurable as we did for the ports?

pkg/test/docker.go
applike-ss: should we maybe add the mapping.setHost value into the slice too?

test/fixtures_dynamodb_test.go
```diff
@@ -1,4 +1,4 @@
-//+build integration
+//+build integration,fixtures
```
applike-ss, line 1: even though we need the fixtures for the test to succeed, i don't think this is a good idea. Because if we forget to add fixtures to the build flags then we don't even see this test failing

applike-ss: same for following tests with the fixtures flag

test/fixtures_redis_test.go
```diff
@@ -54,12 +54,10 @@ func TestFixturesRedisSuite(t *testing.T) {
 
 func (s FixturesRedisSuite) TestRedis() {
 	config := cfg.New()
-	config.Option(
+	_ = config.Option(
```
applike-ss, line 57: how dare you to ignore the error :P

pkg/test/docker.go
josketres: Because for some reason `resource.Container.Name` appends a slash to the container name, so I'm just removing it.

pkg/test/test.go
josketres: For the port you can set 0 to indicate that you want to use dynamic ports.
We cannot do the same for the host. So we just remove it from the configuration.

applike-ss: could it make sense to use this as the configurable ci host/ip?

pkg/test/docker.go
josketres: Yes, it works, because the `dialPort` is a pointer, and before this code is executed the ports are updated. The container-port actually works like a charm on my mac.

pkg/test/docker.go
josketres: I just wanted to be very explicit about the purpose of those pointers:

- dialPort - is a port we are going to use to do a tcp dial to check if a host is reacheable
- setHost - is a pointer where we are going to set a host, once we can reach it

test/fixtures_redis_test.go
josketres: In the `testing` universe, everything is allowed ;)

test/fixtures_dynamodb_test.go
josketres: I think this is a good idea because by adding all the needed flags here, the IDE automatically adds them to the test run configuration, so you will never have to mark that stupid checkbox again: 

image

pkg/test/docker.go
applike-ss: but you are executing the setPortMapping before your host mapping. Given you have a container that internally exposes port 8080 and your config has set port 0 (so auto-assign), then you already have set the settings.Port to the auto-assigned port and your mac is likely to use 127.0.0.1, no?

pkg/test/docker.go
applike-ss: hmm ok, not sure how much i like it, but if we at some point decide to change it its not going to change the api - so fine with me for the moment

test/fixtures_dynamodb_test.go
applike-ss: i am not talking about local execution actually, but about CI. Someone who sets up a new ci process for a new repo might "forget" or just not know that you have to set these

test/fixtures_redis_test.go
applike-ss: i guess it's going to fail anyway if something would have errored

PR #244, opened 2020-04-02
fixed named output nil case

applike-ss:


pkg/mdl/named.go
```diff
@@ -13,6 +15,10 @@ type Nameable interface {
 }
 
 func NamedOutput(in interface{}) interface{} {
+	if in == nil || reflect.ValueOf(in).IsNil() {
```
ajscholl, line 18: IsNil will panic if `in` is not a chan, func, interface, map, pointer, or slice value

applike-ss: afaik, namedOutput should always get a pointer as of now

pkg/mdl/factory.go
```diff
@@ -85,3 +88,29 @@ func EmptyUintIfNil(i *uint) uint {
...
+	switch v.Kind() {
+	case reflect.Chan,
+		reflect.Func,
+		reflect.Map,
+		reflect.Ptr,
+		reflect.UnsafePointer,
+		reflect.Interface,
+		reflect.Slice:
+	default:
+		return true
```
ajscholl, line 108: Shouldn't this be `return false`? If you add some tests for this, I would expect that an empty struct is not nil, but I would also expect the current implementation would say it is nil.

pkg/mdl/factory.go
```diff
@@ -85,3 +88,29 @@ func EmptyUintIfNil(i *uint) uint {
...
+	}
+
+	if v.IsNil() {
+		return true
```
ajscholl, line 112: You can just `return v.IsNil()`

pkg/mdl/factory.go
```diff
@@ -85,3 +88,29 @@ func EmptyUintIfNil(i *uint) uint {
...
+	v := reflect.ValueOf(m)
+
+	switch v.Kind() {
+	case reflect.Chan,
+		reflect.Func,
+		reflect.Map,
+		reflect.Ptr,
+		reflect.UnsafePointer,
+		reflect.Interface,
+		reflect.Slice:
```
ajscholl, line 106: Actually, can we move `return v.IsNil()` here? Would be easier to read IMO

pkg/mdl/factory.go
applike-ss: i would rather want to check it properly, but what kind of types should i add checks for? don't want to make this a 100 liner

PR #245, opened 2020-04-07
db_repo: change history

josketres:
**Summary:**

- Add a `db_repo.MigrateChangeHistory` function.
- The migration creates and updates a history table and a set of triggers.
- This has only been tested on mysql.
- Changed connection cache on `pkg/db/connection.go` to allow connections with different settings (integration tests were failing when running after each other).

**Open points:**

- It is not possible to get the change_author for delete operations with this approach.
- Need a way to prevent 2 applications to run the migration concurrently. (won't implement it in this iteration)
- Needs mysql option `log_bin_trust_function_creators=1`

### DIFF
--- a/pkg/db-repo/change_history.go
+++ b/pkg/db-repo/change_history.go
@@ -1,0 +1,41 @@
+package db_repo
+
+import (
+	"github.com/applike/gosoline/pkg/cfg"
+	"github.com/applike/gosoline/pkg/mon"
+	"time"
+)
+
+type ChangeHistoryModelBased interface {
+	GetHistoryAction() string
+	GetHistoryActionAt() time.Time
+	GetHistoryRevision() int
+}
+
+type ChangeHistoryModel struct {
+	ChangeHistoryAction   string    `sql:"type: VARCHAR(8) NOT NULL DEFAULT 'insert'"`
+	ChangeHistoryActionAt time.Time `gorm:"default:CURRENT_TIMESTAMP"`
+	ChangeHistoryRevision int       `gorm:"primary_key"`
+}
+
+func (c *ChangeHistoryModel) GetHistoryAction() string {
+	return c.ChangeHistoryAction
+}
+
+func (c *ChangeHistoryModel) GetHistoryActionAt() time.Time {
+	return c.ChangeHistoryActionAt
+}
+
+func (c *ChangeHistoryModel) GetHistoryRevision() int {
+	return c.ChangeHistoryRevision
+}
+
+func MigrateChangeHistory(config cfg.Config, logger mon.Logger, models ...ModelBased) {
+	for _, model := range models {
+		manager := newChangeHistoryManager(config, logger, model)
+		err := manager.runMigration()
+		if err != nil {
+			panic(err)
+		}
+	}
+}
--- a/pkg/db-repo/change_history_manager.go
+++ b/pkg/db-repo/change_history_manager.go
@@ -1,0 +1,232 @@
+package db_repo
+
+import (
+	"fmt"
+	"github.com/applike/gosoline/pkg/cfg"
+	"github.com/applike/gosoline/pkg/mon"
+	"github.com/jinzhu/gorm"
+	"reflect"
+	"strings"
+)
+
+type changeHistoryManagerSettings struct {
+	ChangeAuthorField string `cfg:"change_author_column"`
+	TableSuffix       string `cfg:"table_suffix" default:"history"`
+}
+
+type changeHistoryManager struct {
+	orm           *gorm.DB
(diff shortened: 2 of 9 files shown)


pkg/db-repo/change_history_manager.go
```diff
@@ -0,0 +1,236 @@
...
+	"github.com/applike/gosoline/pkg/cfg"
+	"github.com/applike/gosoline/pkg/mon"
+	"github.com/jinzhu/gorm"
+	"reflect"
+	"strings"
+)
+
+type changeHistoryManagerSettings struct {
+	ChangeAuthorField string `cfg:"change_author_column"`
+	TableSuffix       string `cfg:"table_suffix" default:"history"`
```
bogdanfinn, line 14: has to be default pluralized right? because this pluralization stuff is also part of gosoline. otherwise the default settings wont work?

bogdanfinn: also currently thinking about, what happens when someone by accident has a table with real data which name matches the history table pattern. then it will be updated right? even if this would happen by accident? 

might be a problem for the future

pkg/db-repo/change_history_manager.go
```diff
@@ -0,0 +1,232 @@
...
+func (c *changeHistoryManager) runMigration() error {
+	c.buildOriginalTableMetadata()
+	c.buildHistoryTableMetadata()
+
+	if !c.historyTable.exists {
+		c.createHistoryTable()
+		c.dropHistoryTriggers()
+		c.createHistoryTriggers()
+		c.logger.Info("creating change history setup")
+		return c.execute()
```
bogdanfinn, line 61: i would extract this whole Statement generation in a seperate file. just to keep ist more readable.

pkg/db-repo/change_history_manager.go
```diff
@@ -0,0 +1,232 @@
...
+	columnNames := c.originalTable.columnNamesQuotedExcludingValue(c.settings.ChangeAuthorField)
+	var conditions []string
+	for _, columnName := range columnNames {
+		condition := fmt.Sprintf("OLD.%s != NEW.%s", columnName, columnName)
+		conditions = append(conditions, condition)
+	}
+	return strings.Join(conditions, " OR ")
+}
+
+func (c *changeHistoryManager) updateHistoryTable() bool {
```
bogdanfinn, line 201: not sure how this will be in production later. we dont do auto migrations there...

pkg/db-repo/change_history.go
```diff
@@ -0,0 +1,41 @@
...
+
+func (c *ChangeHistoryModel) GetHistoryActionAt() time.Time {
+	return c.ChangeHistoryActionAt
+}
+
+func (c *ChangeHistoryModel) GetHistoryRevision() int {
+	return c.ChangeHistoryRevision
+}
+
+func MigrateChangeHistory(config cfg.Config, logger mon.Logger, models ...ModelBased) {
```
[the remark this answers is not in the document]
josketres: No. This one is the original model. For example `domain.Campaign`

pkg/db-repo/change_history_manager.go
josketres: 👍  I'll change the default to `history_entries`

josketres: The name conflict is a problem for later. I doubt it will happen.

pkg/db-repo/change_history_manager.go
josketres: this auto migration is safe because it only adds columns.
But you're right, maybe we need a way to disable this from the configuration.
Let's wait for @j4k4 opinion about this.

pkg/db-repo/change_history_manager.go
josketres: Interesting idea. I'll give it a try.

pkg/db-repo/change_history.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+type ChangeHistoryModelBased interface {
+	GetHistoryAction() string
+	GetHistoryActionAt() time.Time
+	GetHistoryRevision() int
+}
+
+type ChangeHistoryModel struct {
+	ChangeHistoryAction   string    `sql:"type: VARCHAR(8) NOT NULL DEFAULT 'insert'"`
```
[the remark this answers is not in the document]
josketres: The `db_repo.Model` struct would normally have already been embeded in the struct you are using.

Example:
```go
type Campaign struct {
   db_repo.Model
   Name *string
   ...
}

type CampaignHistoryRevision struct {
   Campaign
   db_repo.ChangeHistoryModel
}

```

PR #248, opened 2020-04-08
parquet: restructure reader to speed up processing on multi-core systems;

ajscholl:
We often have a lot of small files as well as some big files for each
date of events in parquet files. Thus, after processing the small files,
there is only a single big file left to process and we can not take
advantage of multi core systems at all. Instead, we now process a date
set (what we would have done in the end in our applications) and already
start processing the next day while the current day is still not done.
We also split the operation into downloading and processing a file to
make sure we saturate an internet connection much better (if the file is
not already cached locally). When processing big files we however use
quite some memory at once, thus we introduced a new metric about how
many items we want to keep in memory at once (1 million items => 2gb).
This might limit concurrency again, but avoids running out of memory if
you have not too much of that to spare.


applike-ss:
Is this still up-to-date? If so, please rebase on master and let us know, so we can merge it eventually.

PR #250, opened 2020-04-08
metric, closes #146: added retry logic to writing metrics to cw;

### LINKED ISSUE #146: metric: add retry logic
If we write metrics to cloudwatch, we do not retry beyond the normal retries by the AWS sdk. As we already know from other places that these are sometimes not enough, we could add our own retry and backoff logic there, too. 

See for example gosoline/pkg/mon/metric_writer_cw.go:144

ajscholl:
If writing to cloud watch failes, we log an error and thus raise an
alarm. However, neither is this useful nor necessary as we can retry
writing the metric just fine. So this commit introducing a retry logic,
but needs to refactor our package layout a little bit as mon and cloud
can not depend on each other.


applike-ss (changes requested):
Is this still up-to-date? It looks like you could make use of the exec package.

PR #251, opened 2020-04-14
stream: consumer go routine refactor and runnable callback;

j4k4:
Added a RunnableConsumerCallback to the stream to enable the consumer callback
to start additional processing routines for consumed messages.


pkg/stream/consumer.go
```diff
@@ -143,15 +154,41 @@ func (c *Consumer) boolCallback(config cfg.Config, logger mon.Logger) error {
 	return nil
 }
 
-func (c *Consumer) consume() error {
+func (c *Consumer) runCallback(ctx context.Context) error {
```
ajscholl, line 157: You could flip the check to make this a little bit simpler:

```
func (c *Consumer) runCallback(ctx context.Context) error {
	defer c.logger.Debug("runCallback is ending")

	if runnable, ok := c.callback.(RunnableConsumerCallback); ok {
		return runnable.Run(ctx)
	}

	return nil
}
```

j4k4: true

PR #256, opened 2020-04-20
kernel: ensure stages that stopped early are not reported as blocking;

applike-ss:


pkg/kernel/stage.go
```diff
@@ -49,6 +49,14 @@ func (s *stage) boot(k *kernel, bootCoffin coffin.Coffin) {
 
 func (s *stage) stopWait(stageIndex int, logger mon.Logger) {
 	s.cfn.Kill(ErrKernelStopping)
+
+	// in case a complete stage is already not running anymore, we can't rely on s.cfn.Wait
+	if _, ok := <-s.running.Channel(); !ok {
+		logger.Infof("early stopped stage %d", stageIndex)
+		s.terminated.Signal()
```
ajscholl, line 56: You now might have a race between a stage not being started and then already being set to terminated - this seems more like a band aid than a real fix to the underlying problem

ajscholl:
Can you maybe write a test which triggers the problem with the current behavior and then of course no longer triggers with your fix?

PR #266, opened 2020-04-27
cfg: added defaults to basic getter and merge function;

j4k4:


pkg/cfg/config.go
```diff
@@ -20,26 +20,26 @@ import (
...
+	GetStringMap(key string, optionalDefault ...map[string]interface{}) map[string]interface{}
+	GetStringMapString(key string, optionalDefault ...map[string]string) map[string]string
+	GetStringSlice(key string, optionalDefault ...[]string) []string
+	GetTime(key string, optionalDefault ...time.Time) time.Time
 	IsSet(string) bool
 	UnmarshalKey(key string, val interface{})
 }
 
-//go:generate mockery -name GosoConf
+//  go:generate mockery -name GosoConf
```
ajscholl, line 42: There shouldn't be any spaces here, right

pkg/cfg/mapstruct_test.go
```diff
@@ -166,6 +167,75 @@ func TestMapStructIO_ReadZeroAndDefaultValues_Unexported(t *testing.T) {
...
+		D    time.Duration `cfg:"d" default:"1s"`
+		I    int           `cfg:"i" default:"1"`
+		I8   int8          `cfg:"i8" default:"2"`
+		I16  int16         `cfg:"i16" default:"3"`
+		I32  int32         `cfg:"i32" default:"4"`
+		I64  int64         `cfg:"i64" default:"5"`
+		F32  float32       `cfg:"f32" default:"1.1"`
+		F64  float64       `cfg:"f64" default:"1.2"`
+		S    string        `cfg:"s" default:"string"`
+		T    time.Time     `cfg:"t" default:"2020-04-21"`
```
ajscholl, line 184: Why is this default different than the value given in source, but the rest is the same?

pkg/ddb/repository.go
```diff
@@ -78,12 +78,21 @@ func NewRepository(config cfg.Config, logger mon.Logger, settings *Settings) Rep
...
+	tableName := namingStrategy(settings.ModelId)
+
 	tracer := tracing.ProviderTracer(config, logger)
 	client := ProvideClient(config, logger, settings)
 
+	backoffSettings := &cloud.BackoffSettings{}
+	config.UnmarshalKey("ddb.backoff", backoffSettings)
+
+	if err := cfg.Merge(backoffSettings, settings.Backoff); err != nil {
+		logger.Fatalf(err, "could not merge backoff settings for ddb table %s", tableName)
```
ajscholl, line 90: Why do we need to kill the whole server in this case? Shouldn't `logger.Panicf` be enough?

pkg/cfg/config_test.go
```diff
@@ -1,163 +1,259 @@
...
-	assert.Len(t, keys, 2)
+	options := []cfg.Option{
+		cfg.WithErrorHandlers(s.errorHandler),
+		cfg.WithEnvKeyReplacer(strings.NewReplacer(".", "_")),
+		cfg.WithSanitizers(cfg.TimeSanitizer),
+	}
+
+	s.config = cfg.NewWithInterfaces(s.environmentMock)
+
+	if err := s.config.Option(options...); err != nil {
```
ajscholl, line 32: This seems duplicated from line 37, maybe we can just call that

pkg/cfg/config_test.go
```diff
@@ -1,163 +1,259 @@
...
-	assert.True(t, config.IsSet("n.c"))
-	assert.False(t, config.IsSet("b"))
+func (s *ConfigTestSuite) errorHandler(err error, msg string, args ...interface{}) {
+	s.FailNow(err.Error(), fmt.Sprintf(msg, args...))
 }
 
-func TestConfig_Get(t *testing.T) {
-	config := getNewTestableConfig(baseSettings, map[string]string{})
+func (s *ConfigTestSuite) environmentMock(key string) (string, bool) {
+	if value, ok := s.environment[key]; ok {
```
ajscholl, line 48: We could simplify this to

```golang
func (s *ConfigTestSuite) environmentMock(key string) (string, bool) {
	value, ok := s.environment[key]
	
	return value, ok
}
```

pkg/cfg/config_test.go
```diff
@@ -1,163 +1,259 @@
...
+			"b": true,
+		},
+	})
+
+	expectedMap := map[string]interface{}{
+		"b": true,
+	}
+
+	s.Equal(1, s.config.Get("i"))
+	s.Equal(expectedMap, s.config.Get("ms"))
```
ajscholl, line 101: It would also be interesting to add a test case that we get the value from the config even if we supply a different default:

```golang
	s.Equal(1, s.config.Get("i"), "2")
	s.Equal(expectedMap, s.config.Get("ms"), map[string]interface{}{
		"c": false,
	})
```

pkg/cfg/config.go
j4k4: no, i need spaces here, as mockery is broken in this case
had to adjust the mocks manually.. so i don't want them overwritten by generate

ajscholl: Hmm, maybe we should put a comment there in that case, otherwise someone might think this is an error and try to "fix" it

pkg/cfg/mapstruct_test.go
j4k4: dropped the default tags as they are not needed here

pkg/ddb/repository.go
j4k4: true

pkg/cfg/config_test.go
j4k4: +1

pkg/cfg/config_test.go
j4k4: copied :D

pkg/cfg/config_test.go
j4k4: added the test

PR #273, opened 2020-06-11
dbrepo: added capability to use mysql set data type;

applike-ss:


pkg/db-repo/repository.go
```diff
@@ -357,6 +371,120 @@ func ignoreCreatedAtIfNeeded(scope *gorm.Scope) {
...
+		return
+	}
+	vt := reflect.TypeOf(vo.Interface())
+
+	for i := 0; i < vt.NumField(); i++ {
+		vtf := vt.Field(i)
+		vof := vo.Field(i)
+		fn := vtf.Name
+
+		testType := reflect.TypeOf(make([]string, 0))
```
ajscholl, line 403: You could pull this out of the loop

pkg/db-repo/repository.go
```diff
@@ -357,6 +371,120 @@ func ignoreCreatedAtIfNeeded(scope *gorm.Scope) {
 	}
 }
 
+// parses the []uint8 (from mysql) for field x, if there is a field xSet with type []string
+func GormParseMysqlSetType(scope *gorm.Scope) {
+	vo := reflect.ValueOf(scope.Value)
+	if vo.Kind() == reflect.Invalid {
```
ajscholl, line 377: I think you can get rid of this check as you later check whether you have a struct and if not just return

pkg/db-repo/repository.go
```diff
@@ -357,6 +371,120 @@ func ignoreCreatedAtIfNeeded(scope *gorm.Scope) {
...
+		stringVal := string(uintSlice)
+		s := reflect.ValueOf(strings.Split(stringVal, ","))
+		vof.Set(s)
+	}
+}
+
+// sets the []uint8 (for mysql) for field x, if there is a field xSet with type []string
+func GormSetMysqlSetType(scope *gorm.Scope) {
+	vo := reflect.ValueOf(scope.Value)
+	if vo.Kind() == reflect.Invalid {
```
ajscholl, line 433: Same as above

pkg/db-repo/repository.go
```diff
@@ -357,6 +371,120 @@ func ignoreCreatedAtIfNeeded(scope *gorm.Scope) {
...
+		return
+	}
+	for vo.Kind() == reflect.Ptr {
+		vo = vo.Elem()
+	}
+	if vo.Kind() == reflect.Slice {
+		for i := 0; i < vo.Len(); i++ {
+			parseMysqlSetType(vo.Index(i))
```
ajscholl, line 385: Can we have the case of a `[][]Foo`? If so, this is not sufficient, but can't think of anything like this from my head...

pkg/db-repo/repository.go
```diff
@@ -357,6 +371,120 @@ func ignoreCreatedAtIfNeeded(scope *gorm.Scope) {
...
+	vt := reflect.TypeOf(vo.Interface())
+
+	for i := 0; i < vt.NumField(); i++ {
+		vtf := vt.Field(i)
+		vof := vo.Field(i)
+		fn := vtf.Name
+
+		testType := reflect.TypeOf(make([]string, 0))
+		if !vof.Type().AssignableTo(testType) {
+			continue
```
ajscholl, line 405: If we have some struct as a member of another struct (foreign key stuff and so on), will we visit it?

```go
type Foo struct {
    ....
    BarId *uint `gorm:...`
    Bar   *Bar
}

type Bar struct {
    Field []uint8
    FieldSet []string
}
```

And now we get query for a `Foo`...

pkg/db-repo/repository_test.go
```diff
@@ -675,6 +676,111 @@ func TestRepository_DeleteHasManyNoRelation(t *testing.T) {
...
+
+func TestGormParseMysqlSetType(t *testing.T) {
+	assert.NotPanics(t, func() {
+		scope := gorm.Scope{}
+		db_repo.GormParseMysqlSetType(&scope)
+	}, "should not panic on nil")
+	assert.NotPanics(t, func() {
+		scope := gorm.Scope{
+			Value: &gormSetTypeValid{
+				T:    []uint8{97, 44, 98, 44, 99},
```
ajscholl, line 706: What happens if T is nil? If T contains something else than utf8?

pkg/db-repo/repository.go
applike-ss: i do have that in case i have to iterate over a slice which can be nil, so it is still needed.

pkg/db-repo/repository.go
applike-ss: Same as above

pkg/db-repo/repository.go
applike-ss: i think we only either have Foo or []Foo, but never [][]Foo. Why? i would expect that the recursion would do just fine?

pkg/db-repo/repository.go
applike-ss: i did not check that, however i do expect that tbd as this relation is being pulled by the same mechanisms within gorm, isn't it?

pkg/db-repo/repository_test.go
applike-ss: still works :)

pkg/db-repo/repository.go
ajscholl: If the kind is Invalid, we won't enter any of the other if-blocks or loops, will eventually call `parseMysqlSetType` and there see the kind is not a struct and return. So I don't think this block really changes anything.

pkg/db-repo/repository.go
ajscholl: If this whole block would be in `parseMysqlSetType`, it would not matter if you have `[]Foo` or `[][][][][][][]*[][]Foo`

pkg/db-repo/repository.go
```diff
@@ -357,6 +371,132 @@ func ignoreCreatedAtIfNeeded(scope *gorm.Scope) {
 	}
 }
 
+// parses the []uint8 (from mysql) for field x, if there is a field xSet with type []string
+func GormParseMysqlSetType(scope *gorm.Scope) {
+	vo := reflect.ValueOf(scope.Value)
+	if vo.Kind() == reflect.Invalid {
+		return
+	}
+	for vo.Kind() == reflect.Ptr {
```
ajscholl, line 380: This loop is now redundant if you move the slice handling to `parseMysqlSetType`

pkg/db-repo/repository.go
ajscholl: If so, why do we need the handling for slices? But sounds reasonable that gorm should handle this... Can we have some integration test for this?

pkg/db-repo/repository_test.go
```diff
@@ -675,6 +676,151 @@ func TestRepository_DeleteHasManyNoRelation(t *testing.T) {
...
+	assert.NotPanics(t, func() {
+		scope := gorm.Scope{
+			Value: &gormSetTypeValid{
+				T:    nil,
+				TSet: make([]string, 0),
+			},
+		}
+		db_repo.GormParseMysqlSetType(&scope)
+		assert.Equal(t, []string{}, scope.Value.(*gormSetTypeValid).TSet)
+	}, "should not panic on a,b,c in test struct")
```
ajscholl, line 722: There is no a,b,c in the test struct, it is nil.

pkg/db-repo/repository_test.go
```diff
@@ -675,6 +676,151 @@ func TestRepository_DeleteHasManyNoRelation(t *testing.T) {
...
+	assert.NotPanics(t, func() {
+		scope := gorm.Scope{
+			Value: &gormSetTypeValid{
+				T:    nil,
+				TSet: nil,
+			},
+		}
+		db_repo.GormParseMysqlSetType(&scope)
+		assert.IsType(t, []string{}, scope.Value.(*gormSetTypeValid).TSet)
+	}, "should not panic on a,b,c in test struct")
```
ajscholl, line 732: Same again. Also, isn't `assert.IsType(t, []string{}, scope.Value.(*gormSetTypeValid).TSet)` always true because `TSet` is declared to have that type?

applike-ss:
will close this, as we are not using it and implementation is and will be incomplete with the current version of gorm

PR #284, opened 2020-07-15
stream: configure unmarshall

selm0:
Not all stream messages have the message format. Added a configuration mechanism to the unmarshalling of stream inputs.

### DIFF
--- a/pkg/stream/input_sqs.go
+++ b/pkg/stream/input_sqs.go
@@ -21,13 +21,14 @@ type SqsInputSettings struct {
 	RedrivePolicy     sqs.RedrivePolicy     `cfg:"redrive_policy"`
 	Client            cloud.ClientSettings  `cfg:"client"`
 	Backoff           cloud.BackoffSettings `cfg:"backoff"`
+	Unmarshaller      string                `cfg:"unmarshaller" default:"msg"`
 }
 
 type sqsInput struct {
 	logger      mon.Logger
 	queue       sqs.Queue
 	settings    SqsInputSettings
-	unmarshaler MessageUnmarshaler
+	unmarshaler UnmarshallerFunc
 
 	cfn     coffin.Coffin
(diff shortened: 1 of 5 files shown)


pkg/stream/input_sqs.go
```diff
@@ -21,13 +21,14 @@ type SqsInputSettings struct {
...
 	Backoff           cloud.BackoffSettings `cfg:"backoff"`
+	Unmarshaller      string                `cfg:"unmarshaller" default:"msg"`
 }
 
 type sqsInput struct {
 	logger      mon.Logger
 	queue       sqs.Queue
 	settings    SqsInputSettings
-	unmarshaler MessageUnmarshaler
+	unmarshaler UnmarshallerFunc
```
j4k4, line 31: typo in `unmarshaler`?

pkg/stream/input_sqs.go
```diff
@@ -47,10 +48,23 @@ func NewSqsInput(config cfg.Config, logger mon.Logger, s SqsInputSettings) *sqsI
...
+	case UnmarshallerMsg:
+		unmarshaller = MessageUnmarshaller
+	case UnmarshallerRaw:
+		unmarshaller = RawUnmarshaller
+	case UnmarshallerSns:
+		unmarshaller = SnsUnmarshaller
+	default:
+		logger.Fatal(fmt.Errorf("unknown unmarshaller %s", s.Unmarshaller), "")
+	}
+
```
j4k4, line 63: would be better if we store the different unmarshallers in a map inside the stream package..  
makes it open for extension, too..

pkg/stream/input_sqs.go
selm0: According to Java notation, that's the correct way to write it: https://docs.oracle.com/javase/7/docs/api/javax/xml/bind/Unmarshaller.html

PR #290, opened 2020-07-28
cloud: refactored error handling such that we more easily detect a broader range of errors;

ajscholl:
In the past we tried to detect certain (network) on our own. As go now
features errors.Is() and errors.As(), we can simplify this code a lot
and improve the error detection rate a little bit at the same time.


j4k4 (approved):
+1

PR #298, opened 2020-08-10
http: add retry conditions for http client;

bogdanfinn:


pkg/http/client.go
```diff
@@ -35,6 +35,7 @@ type Client interface {
 	SetUserAgent(ua string)
 	SetProxyUrl(p string)
 	SetRedirectValidator(allowRequest func(request *http.Request) bool)
+	AddRetryCondition(f resty.RetryConditionFunc)
```
applike-ss, line 38: we shouldn't expose a direct dependency here and instead use our own type

PR #306, opened 2020-08-25
kvstore: refactored error handling and made chain store return useful results if you are caching missing values;

ajscholl:
In the past, we would return that a value is not missing should the
missingCache contain it and instead return write an empty value to the
out parameter. However, this proved to be a mistake as you would now
have to check whether something was found and whether that something was
actually a valid value. Instead, we now transparently return "nothing
was found" should the cache tell us that we don't need to check for the
value, so there is no difference on the API anymore whether missing
value caching is enabled or not.


j4k4 (approved):
+1

selm0 (approved):

PR #307, opened 2020-08-26
test: added tmpfs for mysql for faster integration testing;

applike-ss:


pkg/test/env/components.go
```diff
@@ -51,7 +51,9 @@ func (c *ComponentBaseSettings) SetType(typ string) {
 }
 
 type ComponentContainerSettings struct {
-	ExpireAfter time.Duration `cfg:"expire_after" default:"60s"`
+	ExpireAfter time.Duration          `cfg:"expire_after" default:"60s"`
+	TmpfsIface  map[string]interface{} `cfg:"tmpfs"`
```
j4k4, line 55: not sure about this :thinking:

j4k4 (changes requested):

PR #309, opened 2020-08-27
exec wip

j4k4:


pkg/exec/error_test.go
```diff
@@ -0,0 +1,120 @@
...
+
+		if listener, err = net.Listen("tcp", "localhost:0"); err != nil {
+			assert.FailNow(t, err.Error(), "can not create listener")
+			return
+		}
+
+		close(waitListen)
+
+		if conn, err = listener.Accept(); err != nil {
+			assert.FailNow(t, err.Error(), "can not accept connection")
```
ajscholl, line 93: `/usr/lib/go/src/testing/testing.go:633` writes:

// A test ends when its Test function returns or calls any of the methods
// FailNow, Fatal, Fatalf, SkipNow, Skip, or Skipf. Those methods, as well as
// the Parallel method, must be called only from the goroutine running the
// Test function.

This is most likely fine (some implementations just do `exit(1)`), but in theory this code is not obeying these rules.

pkg/cloud/aws/error.go
```diff
@@ -0,0 +1,64 @@
...
+	var awsErr awserr.Error
+
+	if errors.As(err, &awsErr) {
+		err = awsErr.OrigErr()
+	}
+
+	return exec.IsConnectionError(awsErr)
+}
+
+func CheckErrorRetryable(_ interface{}, err error) exec.ErrorType {
```
ajscholl, line 50: Why is this function and the next not in exec, but in this package?

pkg/cloud/aws/executor.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+func (b BackoffExecutor) Execute(ctx context.Context, f RequestFunction) (interface{}, error) {
+	return b.executor.Execute(ctx, func(ctx context.Context) (interface{}, error) {
+		req, out := f()
+
+		req.SetContext(ctx)
+		err := req.Send()
+
+		if req.HTTPResponse.StatusCode >= 500 && req.HTTPResponse.StatusCode != 501 {
```
ajscholl, line 88: You could use the constants like `http.StatusInternalServerError` here

pkg/exec/error.go
```diff
@@ -1,32 +1,62 @@
...
 )
 
-func IsAwsError(err error, awsCode string) bool {
-	var aerr awserr.Error
-	if errors.As(err, &aerr) {
-		return aerr.Code() == awsCode
+type ErrorType int
+
+const (
+	ErrorOk ErrorType = iota
```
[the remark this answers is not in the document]
j4k4: ya, i chose a alphabetical order by default but makes sense to put  Unknown first

pkg/exec/error.go
```diff
@@ -38,33 +68,40 @@ func IsRequestCanceled(err error) bool {
 			}
 		}
 
-		return len(multiErr.Errors) > 0
+		if len(multiErr.Errors) > 0 {
```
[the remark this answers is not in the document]
j4k4: looked like this due some refactorings.. redone..

pkg/cloud/aws/error.go
j4k4: because these are special aws error handlers.. i don't want to have them in the generic exec package..

pkg/cloud/aws/executor.go
j4k4: true

ajscholl (approved):

pkg/exec/executor_backoff_test.go
```diff
@@ -0,0 +1,119 @@
...
+		return nil, okError
+	})
+
+	s.EqualError(err, okError.Error())
+	s.Equal(1, tries)
+}
+
+func (s *ExecutorBackoffTestSuite) TestRetryable() {
+	tries := 0
+	retryableError := fmt.Errorf("ok error")
```
ajscholl, line 78: typo?

PR #311, opened 2020-08-28
sqs: set max number of messages in sqs input via cfg;

bogdanfinn:
default value is 10 when no settings key provided so it should not introduce effects on older applications


j4k4 (approved):
+1

PR #317, opened 2020-09-02
sns, stream: enable sns filter policies;

j4k4:


pkg/sns/topic.go
```diff
@@ -182,3 +212,66 @@ func (t *snsTopic) listSubscriptions() ([]*sns.Subscription, error) {
...
+
+	if err = json.Unmarshal([]byte(*actualFilterPolicy), &actualAttributes); err != nil {
+		return false, fmt.Errorf("can not unmarshal actual filter policy: %w", err)
+	}
+
+	if expectedFilterPolicy, err = json.Marshal(attributes); err != nil {
+		return false, fmt.Errorf("can not marshal expected filter policy: %w", err)
+	}
+
+	if err = json.Unmarshal(expectedFilterPolicy, &expectedAttributes); err != nil {
```
ajscholl, line 240: Let me guess, this converts numbers in the map to float64 and similar?

pkg/sns/topic_test.go
```diff
@@ -4,36 +4,29 @@ import (
...
-
-	exec := gosoAws.NewTestableExecutor([]gosoAws.TestExecution{{
-		Output: nil,
-		Err:    nil,
-	}})
+type TopicTestSuite struct {
+	suite.Suite
+	client   *snsMocks.Client
+	executor *gosoAws.TestableExecutor
+	topic    *sns.SnsTopic
```
ajscholl, line 22: Can this be an interface, so SnsTopic does not need to be public?

pkg/sns/topic_test.go
```diff
@@ -43,133 +36,179 @@ func TestTopic_Publish(t *testing.T) {
...
-	exec := gosoAws.NewTestableExecutor([]gosoAws.TestExecution{{
-		Output: nil,
-		Err:    errors.New("error"),
-	}})
+	err := s.topic.Publish(context.Background(), aws.String("test"), map[string]interface{}{})
+	s.NoError(err)
 
-	client := new(snsMocks.Client)
-	client.On("PublishRequest", input).Return(nil, nil)
+	s.True(s.executor.HasExecutedAll())
```
ajscholl, line 63: If this is the only use of `HasExecutedAll`, maybe it should be replaced by some `AssertExpectations` function

pkg/sns/topic.go
```diff
@@ -26,14 +29,14 @@ type Settings struct {
 	Backoff exec.BackoffSettings
 }
 
-type snsTopic struct {
+type SnsTopic struct {
```
ajscholl, line 32: We could leave this private if we add another interface which also includes `SubscribeSqs` (would be enough to have this in the test only)

pkg/sns/topic.go
j4k4: yes

pkg/sns/topic_test.go
j4k4: yeah, that was my first idea, too, but then you have a dependency to the testing package in `cloud/aws`..  
that doesn't felt nice..

pkg/sns/topic.go
j4k4: okay

pkg/sns/topic_test.go
j4k4: see below

pkg/sns/topic.go
selm0: Would be nice to have an easier way to do that :thinking: Does mapstruct allow serialization from struct to struct? I know it allows weak typing

pkg/sns/topic_test.go
selm0: Does it make sense to move the TestableExecutor to a test package? As it's only supposed to be used in tests? There you would be able to depend on the testing package, right?

Or replace the executor entirely by a mock?
I don't see the benefit of having a stub here, especially as you're reimplementing the expectations

ajscholl (approved):

PR #320, opened 2020-09-03
parquet: lock the memory partitioner before flushing it;

ajscholl:
The `flush` function of the `memoryPartitioner` type was operating on a
map without taking the appropriate lock first. This would cause a crash
should the partitioner get a lot of requests (reading from a queue would
be most likely fine, but inventing the elements out of thin air causes
the crash repeatedly).


j4k4 (commented):
+1

PR #326, opened 2020-09-07
coffin, kernel: added stack trace to recovered errors and changed kernel

ajscholl:
to not ignore panics of the last foreground module;

### DIFF
--- a/pkg/coffin/recover.go
+++ b/pkg/coffin/recover.go
@@ -2,6 +2,8 @@ package coffin
 
 import (
 	"fmt"
+	"runtime"
+	"strings"
 )
 
 func ResolveRecovery(unknownErr interface{}) error {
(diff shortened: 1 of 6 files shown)


pkg/coffin/recover.go
```diff
@@ -10,12 +13,32 @@ func ResolveRecovery(unknownErr interface{}) error {
...
+	done := false
+	for !done {
+		frame, more := st.Next()
+
+		stack = append(stack, fmt.Sprintf("%s:%d %s", frame.File, frame.Line, frame.Function))
+
+		done = !more
+	}
+
+	return errors.WithMessagef(err, "%s\nstack:\n\t%s", err.Error(), strings.Join(stack, "\n\t"))
```
j4k4, line 43: Can we do this with the built in errors package? 🤔

ajscholl: Sure, I initially tried to avoid them because I had the feeling they were getting swallowed in some cases, but it was just the kernel shutting down as the last foreground module exited before the error could be propagated. Changed to `fmt`.

j4k4 (approved):
+1

PR #327, opened 2020-09-09
ddb: allow passing of whole entities to BatchDeleteItems;

selm0:
When passing an entity struct with all values filled to repository.BatchDeleteItems, then you get an error as the request is only expecting the keys.
I added the transformation of the structs to their respective keys.

### DIFF
--- a/pkg/ddb/repository.go
+++ b/pkg/ddb/repository.go
@@ -66,8 +66,9 @@ type repository struct {
 	client   dynamodbiface.DynamoDBAPI
 	executor aws.Executor
 
-	metadata *Metadata
-	settings *Settings
+	keyBuilder keyBuilder
+	metadata   *Metadata
+	settings   *Settings
 }
 
 func NewRepository(config cfg.Config, logger mon.Logger, settings *Settings) Repository {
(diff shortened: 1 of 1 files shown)


pkg/ddb/repository.go
```diff
@@ -212,8 +252,8 @@ func (r *repository) BatchDeleteItems(ctx context.Context, value interface{}) (*
 	})
 }
 
-func (r *repository) batchWriteItem(ctx context.Context, value interface{}, reqBuilder func(map[string]*dynamodb.AttributeValue) *dynamodb.WriteRequest) (*OperationResult, error) {
-	items, err := refl.InterfaceToInterfaceSlice(value)
+func (r *repository) batchWriteItem(ctx context.Context, values []map[string]*dynamodb.AttributeValue, reqBuilder func(map[string]*dynamodb.AttributeValue) *dynamodb.WriteRequest) (*OperationResult, error) {
+	items, err := refl.InterfaceToInterfaceSlice(interface{}(values))
```
ajscholl, line 256: It doesn't look right to use reflection for this - you already have a slice, it is just of a too specific type. Instead, you could easily change chunk to accept your slice, then you also don't need to cast later

pkg/ddb/repository.go
```diff
@@ -190,7 +196,24 @@ func (r *repository) BatchPutItems(ctx context.Context, value interface{}) (*Ope
...
 
-	return r.batchWriteItem(ctx, value, func(item map[string]*dynamodb.AttributeValue) *dynamodb.WriteRequest {
+	items, err := refl.InterfaceToInterfaceSlice(value)
+
+	if err != nil {
+		return nil, fmt.Errorf("no slice of items provided for batchWriteItem operation on table %s: %w", r.metadata.TableName, err)
+	}
+
+	marshalledItems := make([]map[string]*dynamodb.AttributeValue, len(items))
+	for i := 0; i < len(items); i++ {
```
ajscholl, line 206: Why not use a normal range loop?

ajscholl (changes requested):

pkg/ddb/repository.go
selm0: That actually never was reflection, but chunking on generic (=interface) level

j4k4:
So the interface stays the same and it is just more convenient to use? 🤔

selm0:
> So the interface stays the same and it is just more convenient to use?

Yes. Before, you weren't allowed to pass structs with additional properties besides hash/range key. Now you can do that.

ajscholl (approved):

j4k4:
👍

PR #328, opened 2020-09-12
application,cli: made cli logging output adjustable;

ss89:
Making a cli application with gosoline didn't seem as straight forward as it was with for consumer/subscriber/api based applications.
This is a first step which makes the logger output "file" configurable (which defaults to stdout if not overwritten).
Overwriting is highly recommended for cli applications, as it might otherwise bloat the output.


pkg/application/options.go
```diff
@@ -243,8 +248,14 @@ func WithLoggerSettingsFromConfig(app *App) {
 		settings := &loggerSettings{}
 		config.UnmarshalKey("mon.logger", settings)
 
+		outputFile, err := os.OpenFile(settings.Output.File, os.O_CREATE|os.O_WRONLY, 0755)
```
ajscholl, line 251: Should be mode 0644; Also please try whether O_APPEND works or seek to the end of the file if it is a seekable file - we would want to avoid overwriting existing logs.

pkg/cli/cli.go
```diff
@@ -19,41 +20,29 @@ type cliModule struct {
...
-
-	if err := module.Boot(config, logger); err != nil {
-		defaultErrorHandler(err, "can not boot the module")
-	}
-
-	settings := &kernelSettings{}
-	config.UnmarshalKey("kernel", settings)
-
-	k := kernel.New(config, logger, kernel.KillTimeout(settings.KillTimeout))
+	k := application.New([]application.Option{
```
ajscholl, line 30: Should be

```go
	app := application.New(
		application.WithUTCClock(true),
		application.WithConfigErrorHandlers(defaultErrorHandler),
		application.WithConfigFile("./config.dist.yml", "yml"),
		application.WithConfigFileFlag,
		application.WithConfigEnvKeyReplacer(strings.NewReplacer(".", "_", "-", "_")),
		application.WithConfigSanitizers(cfg.TimeSanitizer),
		application.WithLoggerFormat(mon.FormatConsole),
		application.WithLoggerApplicationTag,
		application.WithLoggerTagsFromConfig,
		application.WithLoggerSettingsFromConfig,
		application.WithLoggerContextFieldsMessageEncoder(),
		application.WithLoggerContextFieldsResolver(mon.ContextLoggerFieldsResolver),
		application.WithLoggerSentryHook(mon.SentryExtraConfigProvider, mon.SentryExtraEcsMetadataProvider),
		application.WithKernelSettingsFromConfig,
	)
```

ajscholl (changes requested):

pkg/application/options.go
selm0, line 251: Why can't you pass the logger output file as an additional option when creating your application?

E.g. `app.New(application.WithLoggerOutput(os.Stdout))`

I don't see a benefit of requiring an output file for any kind of application.

selm0 (changes requested):

pkg/application/options.go
ss89: the permissions i do agree with.

For the O_APPEND, i do get this as output when trying to write to a log file on mac:
```
Failed to write to log, write output.log: bad file descriptor
```

ss89: @selm0 anything is a file in unix/linux/mac world. Even stdout/stderr are.

I do not want to have to add it as an option, because it makes sense to have it as a logging option instead.
Usually you can configure applications, where they should log to.
Consider any kind of linux service.

ss89:
@selm0 @ajscholl can you please re-review and tell me if there are further changes requested?

I would like to merge this asap and also already have the next change set on top of this to add the ability to send logs to remote receivers.
However i would add this new change set in a separate merge request

pkg/cli/cli.go
```diff
@@ -19,41 +20,29 @@ type cliModule struct {
...
-
-	if err := module.Boot(config, logger); err != nil {
-		defaultErrorHandler(err, "can not boot the module")
-	}
-
-	settings := &kernelSettings{}
-	config.UnmarshalKey("kernel", settings)
-
-	k := kernel.New(config, logger, kernel.KillTimeout(settings.KillTimeout))
+	k := application.New(
```
ajscholl, line 30: k is now an app and not a kernel, you could rename the variable. But this is only minor nitpicking.

Another thing would be to have more shared initialization logic - the app options you are listing are a subset of the defaults, so maybe we should move them all to the application package and then add the options for a service app to that list - otherwise one might forget to add an option for a CLI app when adding it to a server which should've been added to a CLI app.

A last thing, do we still need the cliModule type? The only use I see is having the module be an essential module, but you can also pass that as an option to Add().

pkg/application/options.go
```diff
@@ -243,8 +248,14 @@ func WithLoggerSettingsFromConfig(app *App) {
 		settings := &loggerSettings{}
 		config.UnmarshalKey("mon.logger", settings)
 
+		outputFile, err := os.OpenFile(settings.Output.File, os.O_CREATE|os.O_WRONLY|os.O_APPEND, 0644)
+		if err != nil {
+			return err
```
ajscholl, line 253: Can we do something like `return fmt.Errorf("Failed to open logfile '%s': %w", settings.Output.File, err)`? I would avoid to have a failure of `error: permission denied` without knowing where this is coming from (not quite sure what os.OpenFile returns in case of an error, but I alread have seen it quite a lot to be told something does not exist for example without getting told what does not exist).

pkg/application/options.go
selm0: Having a file output configurable doesn't reveal on how it's implemented.

I still think it should be up to the application to define, where they want to log to.
With your addition of the new option stack to the cli, it's even easier to use `WithLoggerOutput`.

Assume I want to pass my logs via a channel to a consumer, then I'd still be stuck with a file being opened unnecessarily.

ss89: > Having a file output configurable doesn't reveal on how it's implemented.

Why would it need to reveal to the "user" how it is implemented? Nothing changes in how the log is written, so the requirement shouldn't change within this PR imho.

> I still think it should be up to the application to define, where they want to log to.

This is not how one would usually write applications.
The user/administrator of the pc executing it is in charge to configure it, if they don't like the defaults.
Or do you mean that there should just be the default os.StdOut via an application option, but still my config flag?

> Assume I want to pass my logs via a channel to a consumer, then I'd still be stuck with a file being opened unnecessarily.

Totally new scenario here that we never talked about.
With a channel do you mean a generic communication channel as in a connection or are we talking about go channels here?
If you mean a generic communication channel, then this is what my next PR is actually about: Logging to an external destination via tcp or udp.

pkg/cli/cli.go
ss89: Makes sense, applied all of that, but not the last part "remove cliModule".

In fact we don't need it, it is just a handy helper like we have for the `RunConsumer`

pkg/application/options.go
ss89: definitely a good idea, thanks :)

ss89:
@selm0 @ajscholl mind a re-review?

j4k4:
i agree, that it should be possible to write logs to a file.. especially for cli applications.  
but i really don't like file output on the logger settings. we should try to go into a appender (log4j) / handler (monolog) direction.  
meaning registering different appenders/handlers into the logger and then selecting the appropriate one via config.

ajscholl:
I would agree with @j4k4 here that the logger settings get really focused on logging to a file. Do we need them to be that way? If you just want to have all log output go to a file, just redirect stdout to the file when invoking your app, right? If you later want to have warnings and errors in stdout or err and all logs in a file (think running some batch application) your single file is no longer enough and you need to refactor it completely again to have multiple outputs/appenders/handlers, however you want to call them.

ss89:
> I would agree with @j4k4 here that the logger settings get really focused on logging to a file. Do we need them to be that way? If you just want to have all log output go to a file, just redirect stdout to the file when invoking your app, right? If you later want to have warnings and errors in stdout or err and all logs in a file (think running some batch application) your single file is no longer enough and you need to refactor it completely again to have multiple outputs/appenders/handlers, however you want to call them.

That doesn't work for me. I am writing an app that is called by a 3rdparty and has to integrate into an interface that is not in my control. That being said: I need stdout and stderr to stay clean and have anything else written to it then what i suppose to be written there.
I know that this sounds very special, but in the end the solution is as simple as logging to a file and ensuring that no other component of gosoline accidentally writes to stdout and stderr.
I don't really see the issue of calling it a "file" here, because that's what it is in unix terminology, however i see the advantage of having the logger better configurable.

j4k4:
let me know, when you're ready for another review

ss89:
i have produced a first working draft, definitely in for comments now. Maybe @selm0 and me can have a look first before presenting it to you.

pkg/application/defaults.go
```diff
@@ -0,0 +1,47 @@
+package application
+
+import (
+	"github.com/applike/gosoline/pkg/cfg"
+	"github.com/applike/gosoline/pkg/mon"
+	"strings"
+)
+
+var DefaultServiceApp = []Option{
```
selm0, line 9: You could extract duplicate options between the two app types and make them available for custom app types

pkg/mon/logger_output_iowriter.go
```diff
@@ -0,0 +1,38 @@
+package mon
+
+import (
+	"fmt"
+	"github.com/applike/gosoline/pkg/clock"
+	"io"
+	"os"
+)
+
+func newIowriterLoggerWithInterfaces(clock clock.Clock, formatter formatter, output *os.File, timestampFormt string) (Handler, error) {
```
selm0, line 10: Like the abstraction here, maybe you could just pass a writer interface, you can pass different things than os.File?

Also: `newIoWriter` and `timestampFormat` :D

selm0: And maybe make this one public, so others can use it too

pkg/mon/logger_output_iowriter.go
```diff
@@ -0,0 +1,38 @@
...
+
+func (s *iowriterLogger) Write(level string, msg string, logErr error, metadata Metadata) {
+	timestamp := s.clock.Now().Format(s.timestampFormat)
+
+	buffer, err := s.formatter(timestamp, level, msg, logErr, &metadata)
+	if err != nil {
+		_, _ = fmt.Fprintf(os.Stderr, "Failed to format log, %v\n", err)
+	}
+
+	_, err = s.output.Write(buffer)
```
ajscholl, line 34: What happens if only part of the log message got written? I know we didn't in the past, but maybe you should check the number of written bytes, especially should the io.Writer be something like a network stream

applike-ss: and then log the error to... what? ^^

ajscholl: The same place you are logging the error you can already get (stderr that is). But as long as the writer abides to the rules of `io.Writer`, we should always get an error for a partial write (that is just completely different from the libc write which might return success for a partial write...)

ss89:
i just pushed a cleanup commit to make the unit tests and go vet happy.

This did also imply the removal of two options WithFormat and WithTimestampFormat, however maybe we can bring them back somehow for the handlers one day?

ss89:
@j4k4 can you give me a hint how to make the default value issue with the logger cleaner? i would like to have the "stdout" logger handler by default in mon.logger.handler[stdout/default] for existing applications. Or is this a case we have to live with and should set it in each application? I could also not use the post processors because they run later and if a moved them to a spot before the logger options are loaded then they would crash when parsing the tags. Using the example api application for testing btw.

ss89:
ready for review @j4k4, but i'm not very happy what's going on in application/options.go. Hoping you have a good suggestion here.

pkg/application/options.go
```diff
@@ -245,14 +233,69 @@ func WithLoggerSettingsFromConfig(app *App) {
...
+		loggerOptions, err := getLoggerHandlersFromConfigWithFallback(config, loggerOptions)
+		if err != nil {
+			return err
 		}
 
 		return logger.Option(loggerOptions...)
 	})
 }
 
+func getLoggerHandlersFromConfigWithFallback(config cfg.GosoConf, loggerOptions []mon.LoggerOption) ([]mon.LoggerOption, error) {
```
j4k4, line 247: i agree, that this is not optimal..
i would need some time to think about it.. does it have time until next year?

ss89: Definitely, i do see that such big changes should be thought through, so we don't have to re-implement everything in a worst case scenario.

applike-ss:
As this is implemented (slightly different) already, i closed this PR

PR #331, opened 2020-09-24
sqs: fix issue with messages being too large;

applike-ss:


selm0 (approved):

j4k4:
Can we have a test for this?

PR #332, opened 2020-09-28
sqs: added unit test for SendBatch;

applike-ss:


pkg/sqs/queue_test.go
```diff
@@ -0,0 +1,88 @@
...
+	qs.executor.
+		On("Execute", context.Background(), mock.AnythingOfType("aws.RequestFunction")).
+		Once().
+		Return(nil, awsErr)
+	err = qs.queue.SendBatch(context.Background(), msgs)
+	if !qs.Equal(awsErr, err) {
+		qs.FailNow("SendBatch should fail with only one message which triggers a request too long error")
+	}
+
+	// Test case 3: 3 messages in batch, but triggers a request too long error
```
selm0, line 62: Can you please move the different est cases to different test methods? The queue is stateless, thus there shouldn't be any issue.
It's also slightly better structured :wink:

selm0 (changes requested):

pkg/sqs/queue_test.go
applike-ss: please have a look at my freshly pushed code

selm0 (approved):

PR #335, opened 2020-10-12
tracing: add support for sampling;

netsi:


j4k4 (commented):
without configuring sampling it keeps the default behaviour?

netsi:
The best would have been if I could define a `*SamplingConfiguration` at the tracing configuration, so that it would simply ignore that configuration. But currently the cfg pkg does not support it. 

But due to the default values setted in the `SampleRule` struct if the configuration is missing it should behave as before.
We could also add an enabled flag at the `SamplingConfiguration` but that would a misleading because by default the aws go SDK uses the sampling as defined in `github.com/aws/aws-xray-sdk-go/resources/DefaultSamplingRules.json`

j4k4 (approved):
+1

PR #339, opened 2020-10-21
mon: replace WithFields(MSI) with WithFields(Fields);

ajscholl:
WithFields used to accept any map[string]interface{}, but was almost
always called with mon.Fields instead. While the two types are
compatible and client code should not be affected, tests who check what
gets logged are. In those you were in the awkward situation that you
passed mon.Fields in your code under test, but the mock needed to listen
for map[string]interface{}, or it would not match. This is no longer the
case.


j4k4 (approved):
+1

PR #343, opened 2020-10-23
kvstore: added metrics to the kvstore;

ajscholl:
This change measures on each request to the kvstore whether it was
issued, successful, and, if available, the size of the store at that
time. Thus, you can compute the cache hit rate as reads / readRequests
where reads denotes the number of successful reads and readRequests the
number of started reads.

This also fixes a bug in the batch delete method of the inMemory store.


j4k4 (approved):
+1

PR #344, opened 2020-10-23
stream: reworked batch consumer;

applike-ss:


selm0 (changes requested):
Please unify the consumer and batch consumer by adding a base consumer.
Adhere to DRY.

pkg/stream/base_consumer.go
```diff
@@ -0,0 +1,363 @@
...
+func (c *baseConsumer) recover() {
+	err := coffin.ResolveRecovery(recover())
+	if err == nil {
+		return
+	}
+
+	c.logger.Error(err, err.Error())
+}
+
+func (c *baseConsumer) callBatchCallback(ctx context.Context, batch []*Message, models []interface{}, attributes []map[string]interface{}) int {
```
selm0, line 320: The base consumer must not know about the consumption process, i.e. whether it's batch or not.
Those are implementation details of the actual consumers.
Please move anything related to that to the specific implementation.

Consider adding a new consumer type which does not fit the batch or single processing pattern, then you would need to touch the base consumer again.

selm0: I think this requires moving of `doConsuming` aswell

selm0 (changes requested):

pkg/stream/base_consumer.go
applike-ss: This is what we had before you told me to migrate the two. Maybe you want to finish the implementation yourself?

j4k4 (approved):
+1

selm0 (approved):

PR #348, opened 2020-10-29
clock: added ticker implementation;

ajscholl:
This adds an improved ticker compared to the default ticker of go.
Instead of just ticking, it can also be reset, meaning the next tick
occurs only after the whole duration again. We are already using this
pattern in quite some places, but mostly with replacing the ticker
itself. Instead, you now can just call Ticker.Reset (which is even
thread-safe).

### DIFF
--- a/pkg/clock/ticker.go
+++ b/pkg/clock/ticker.go
@@ -1,0 +1,173 @@
+package clock
+
+import (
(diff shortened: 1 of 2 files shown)


j4k4 (approved):
+1

PR #349, opened 2020-10-30
stream: change interface of Output to accept any WritableMessage;

ajscholl:
If you want to publish a message without any extra marshalling done by
gosoline, you were left alone in the past. Now you can pass anything
which can be marshalled to a string or []byte to Write and WriteOne.
Also, a new type RawMessage was added which allows you to specify the
encoding function yourself if needed.

This is a breaking change as you now need to pass []WritableMessage
instead of []*stream.Message to Output.Write. If you can't easily change
your code, there is stream.MessagesToWritableMessages to transform your
messages.


j4k4 (approved):
+1

PR #351, opened 2020-11-03
kvstore: exposed inMemory configuration options to config file;

### LINKED ISSUE #143: kvstore: add an in-memory cache with configurable max item count
We should add another cache layer into the kvstore which would be in-memory to reduce network usage whenever possible.

This is useful the most for applications that hold a small data set in their kvstore that is frequently used.

ajscholl:
This fixes #143 by allowing a user to specify different cache sizes etc.
than the defaults. If no or invalid values are specified the cache falls
back to the default values.

Closes #143.

### DIFF
--- a/pkg/kvstore/configurable.go
+++ b/pkg/kvstore/configurable.go
@@ -15,15 +15,25 @@ const (
 )
 
 type ChainConfiguration struct {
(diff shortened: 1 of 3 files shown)


j4k4 (approved):
+1

PR #356, opened 2020-11-09
stream: allow a few seconds for the producer daemon to write its messages upon shutdown;

ajscholl:
When the kernel initiates a shutdown while there are queued messages, we
need to give at least a few seconds to write them to their output,
otherwise we immediately receive a context.Canceled error.


j4k4 (approved):
+1

j4k4 (approved):
+1

PR #365, opened 2020-11-25
change design of module create / boot

j4k4:
idea is to combine create and boot of a module via a factory func  

```
type ModuleFactory func(ctx context.Context, config cfg.Config, logger mon.Logger) (Module, error)
```


ajscholl:
I like it

justIrina:
maybe changes also are required for `/pkg/currency/module.go:46`

PR #371, opened 2020-11-30
make all paths absolut;

bogdanfinn:
We need this change for a build tool which has problems with relative paths. The changes done here should produce the same behavior like before but with absolute paths.


ajscholl:
This sounds like your build tool has problems with setting the correct current working directory. With your change, a user can no longer install an executable e.g. in /usr/local/bin and execute it with a local config.dist.yml, but instead the config has to be located in /usr/local/bin/config.dist.yml.

applike-ss:
can you share some information about your build tool and your setup?

PR #372, opened 2020-11-30
kernel: remove unix sys calls;

bogdanfinn:
One might use / build with gosoline on windows ... :-D


ajscholl:
You are replacing the unix package with a deprecated package:

https://golang.org/pkg/syscall/

"Deprecated: this package is locked down. Callers should use the corresponding package in the golang.org/x/sys repository instead. That is also where updates required by new systems or versions should be applied. See https://golang.org/s/go1.4-syscall for more information."

Does it actually build on windows with the syscall package instead of the unix package?

PR #376, opened 2020-12-07
apiserver/auth: Added the token as a second argument to the TokenBearerProvider

ajscholl:
It might be the case that you need the token to decide which TokenBearer
to use. Thus you can now access it as an additional argument in the
TokenBearerProvider type.


j4k4 (approved):
+1

PR #377, opened 2020-12-07
integration: add static code analysis tool;

ss89:
@selm0 what do you think about this?

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -10,7 +10,7 @@ on:
 
 jobs:
   terraform-fmt:
-    name: Terraform fmt
+    name: terraform fmt
     runs-on: ubuntu-20.04
     steps:
       - name: Check out code into the Go module directory
@@ -20,7 +20,7 @@ jobs:
         run: terraform fmt -check -recursive .
 
   go-fmt:
-    name: Go fmt
+    name: go fmt
     runs-on: ubuntu-20.04
     steps:
       - name: Check out code into the Go module directory
@@ -33,11 +33,11 @@ jobs:
         run: test -z $(gofmt -l .)
 
   build:
-    name: Go build
+    name: go build
     runs-on: ubuntu-20.04
     steps:
       - name: Set up Go 1.15
-        uses: actions/setup-go@v1
+        uses: actions/setup-go@v2
         with:
           go-version: 1.15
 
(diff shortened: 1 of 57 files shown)


selm0:
> @selm0 what do you think about this?

I was thinking about going for this one: https://golangci-lint.run/ it has your tool integrated.
First step, leaving the config untouched as it sounds quite reasonable. What's your opinion?

ss89:
this is going to take some time, it doesn't really state where the errors are, just writes some names...

applike-ss:
i have executed this locally and it spits out way too many false positives for example for "unused" properties in a struct.

That may be the case, when you just look at the struct itsself, but since we embed that given struct in other structs, it is in fact used.

ss89:
@selm0 mind approving this?

j4k4:
i like the idea of integrating static code analysis..  
plz let me have a look at the tools, too :)

selm0:
> i have executed this locally and it spits out way too many false positives for example for "unused" properties in a struct.
> 
> That may be the case, when you just look at the struct itsself, but since we embed that given struct in other structs, it is in fact used.
> 
> EDIT: This is the bug: [golangci/golangci-lint#826](https://github.com/golangci/golangci-lint/issues/826)
> 
> EDIT 2: For the moment i would stay with the staticcheck tool

You could disable `structcheck`.
How does the integration with Goland work with `staticcheck`?

ss89:
in fact i didn't integrate it into my goland, but used the go linter plugin which basically is a wrapper around golangci-lint. That however was putting my macbooks down on its knees because it was too weak to handle it. Or in other words: i wouldn't be able to be productive on the machine ^^

ss89:
@j4k4 i guess you will let me know about your findings and i will wait for you then?

j4k4:
you will be using https://golangci-lint.run/ now?

ss89:
Might try the suggestion of @selm0.

this might also be worth a try: https://blog.jetbrains.com/phpstorm/2020/12/early-access-program-for-qodana-a-new-static-analysis-and-quality-management-tool-by-jetbrains-is-open/

ss89:
@selm0 mind having a look again? would really like to see this one merged soonish. Also please have a very close look (maybe together with @j4k4?) about the funcs i removed. Not sure if these should've been used at a later point or if they are left overs from the past.

j4k4 (approved):
i like adding the static checks

selm0 (approved):
Looks good to me.

ss89:
@j4k4 i removed some stuff you added yesterday. Please take a look at my rebase fixes commit if that is ok with you.

j4k4:
all good

PR #378, opened 2020-12-08
cloud,ddb,fixtures: fixed dynamodb client providing based on endpoint…

applike-ss:
… url;


applike-ss:
@selm0 do you approve this?

PR #382, opened 2020-12-13
http: move http client creation from NewHttpClientWithInterfaces to N…

ss89:
…ewHttpClient func;

### DIFF
--- a/pkg/http/client.go
+++ b/pkg/http/client.go
@@ -56,10 +56,10 @@ type Response struct {
 type headers map[string]string
 
 type client struct {
-	mo             mon.MetricWriter
 	logger         mon.Logger
-	http           *resty.Client
 	defaultHeaders headers
+	http           restyClient
+	mo             mon.MetricWriter
 }
 
 type Settings struct {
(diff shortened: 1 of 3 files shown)


ss89:
@selm0 could you quickly review this?

pkg/http/client.go
```diff
@@ -76,22 +76,22 @@ func NewHttpClient(config cfg.Config, logger mon.Logger) Client {
...
 	httpClient.SetRedirectPolicy(resty.FlexibleRedirectPolicy(10))
 	httpClient.SetRetryCount(settings.RetryCount)
 	httpClient.SetTimeout(settings.Timeout)
 	httpClient.SetRetryWaitTime(settings.RetryWaitTime)
 	httpClient.SetRetryMaxWaitTime(settings.RetryMaxWaitTime)
 
+	return NewHttpClientWithInterfaces(logger, mo, httpClient)
+}
+
+func NewHttpClientWithInterfaces(logger mon.Logger, mo mon.MetricWriter, httpClient *resty.Client) Client {
```
selm0, line 89: What's your use case here? `httpClient *resty.Client`: I don't see the benefit of directly passing the client to the test method..

selm0 (changes requested):

pkg/http/client.go
ss89: sorry, as said should have been an interface. Was just too sleepy to realize i was passing a pointer

ss89:
@selm0 could you have a quick look and approve?

ss89:
@selm0 ?

selm0 (approved):

PR #383, opened 2020-12-17
redis: "hide" the error returned by redis if an element is not found from the executor;

ajscholl:
Redis returns `redis.Nil` as an error should no element be found. This
however confuses the executor as it thinks the command failed and needs
to be retried. Instead, we have to return nil to the executor and later
restore the error for the caller to ensure the use of the executor stays
invisible to the caller, but we don't retry requests we don't need to.

Also remove a rare crash in the aws executor should an error be returned.

### DIFF
--- a/pkg/cloud/aws/executor.go
+++ b/pkg/cloud/aws/executor.go
@@ -61,7 +61,7 @@ func (b BackoffExecutor) Execute(ctx context.Context, f RequestFunction) (interf
 		req.SetContext(ctx)
 		err := req.Send()
 
-		if req.HTTPResponse.StatusCode >= http.StatusInternalServerError && req.HTTPResponse.StatusCode != http.StatusNotImplemented {
+		if err == nil && req.HTTPResponse.StatusCode >= http.StatusInternalServerError && req.HTTPResponse.StatusCode != http.StatusNotImplemented {
 			return nil, &InvalidStatusError{
 				Status: req.HTTPResponse.StatusCode,
 			}
--- a/pkg/exec/error.go
+++ b/pkg/exec/error.go
@@ -12,9 +12,16 @@ import (
 type ErrorType int
 
 const (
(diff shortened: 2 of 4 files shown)


pkg/cloud/aws/executor.go
```diff
@@ -61,7 +61,7 @@ func (b BackoffExecutor) Execute(ctx context.Context, f RequestFunction) (interf
 		req.SetContext(ctx)
 		err := req.Send()
```
applike-ss, line 62: shouldn't we actually check this first before checking the reponses status code?

pkg/cloud/aws/executor.go
```diff
@@ -61,7 +61,7 @@ func (b BackoffExecutor) Execute(ctx context.Context, f RequestFunction) (interf
...
 
-		if req.HTTPResponse.StatusCode >= http.StatusInternalServerError && req.HTTPResponse.StatusCode != http.StatusNotImplemented {
+		if err == nil && req.HTTPResponse.StatusCode >= http.StatusInternalServerError && req.HTTPResponse.StatusCode != http.StatusNotImplemented {
 			return nil, &InvalidStatusError{
 				Status: req.HTTPResponse.StatusCode,
 			}
```
applike-ss, line 67: also `return out, nil` instead of `return out, err` please

pkg/cloud/aws/executor.go
ajscholl: I... do exactly this with my change set? If the error is not nil, the status code is not checked and I just propagate the error?

pkg/cloud/aws/executor.go
ajscholl: Not sure why this comment is on this line, but the `return out, err` line needs to stay as it is, otherwise it would swallow an error

j4k4 (approved):
+1

PR #384, opened 2020-12-18
db-repo: add handling for duplicate entry error

justIrina:


pkg/db-repo/repository.go
```diff
@@ -24,6 +26,7 @@ const (
 
 var operations = []string{Create, Read, Update, Delete, Query}
 var RecordNotFound = gorm.ErrRecordNotFound
+var DuplicateEntry = errors.New("duplicate entry")
```
ajscholl, line 29: You don't need that anymore

justIrina: actually we also don't need record not found here. Or do we want to keep it for the future?

ajscholl: We need RecordNotFound because gorm can return it to us and we then return that to our caller, but our caller should not need to import gorm to check for that error, but instead use db_repo.RecordNotFound.

pkg/db-repo/errors.go
```diff
@@ -0,0 +1,31 @@
...
+
+type DuplicateEntryError struct {
+	Err error
+}
+
+func (e *DuplicateEntryError) Error() string {
+	return fmt.Sprintf("duplicate entry: %s", e.Err.Error())
+}
+
+func (e *DuplicateEntryError) Is(err error) bool {
```
applike-ss, line 13: Why do you do this custom implementation? Go already has you covered for checking if an error is (errors.Is) another error or if an error can be unwrapped to another error (errors.As).

pkg/db-repo/repository.go
```diff
@@ -89,7 +90,15 @@ func (r *repository) Create(ctx context.Context, value ModelBased) error {
...
 
-	// TODO: Check for duplicate error and return a local error type instead which applications can handle themselves
+	exists := db.IsDuplicateEntryError(err)
+
+	if exists {
+		logger.Warnf("could not create model of type %v due to duplicate entry error &s", modelId, err.Error())
+		return &DuplicateEntryError{
+			Err: err,
+		}
+	}
```
applike-ss, line 100:
```suggestion
	isDuplicate := db.IsDuplicateEntryError(err)
	if isDuplicate {
		logger.Warnf("could not create model of type %s due to duplicate entry error: %s", modelId, err.Error())
		return &DuplicateEntryError{
			Err: err,
		}
	}
```

pkg/db-repo/repository.go
```diff
@@ -126,6 +135,15 @@ func (r *repository) Update(ctx context.Context, value ModelBased) error {
...
 
+	exists := db.IsDuplicateEntryError(err)
+
+	if exists {
+		logger.Warnf("could not update model of type %s with id %d due to duplicate entry error %s", modelId, *value.GetId(), err.Error())
+		return &DuplicateEntryError{
+			Err: err,
+		}
+	}
+
```
applike-ss, line 146:
```suggestion
	isDuplicate := db.IsDuplicateEntryError(err)
	if isDuplicate {
		logger.Warnf("could not update model of type %s with id %d due to duplicate entry error: %s", modelId, *value.GetId(), err.Error())
		return &DuplicateEntryError{
			Err: err,
		}
	}

```

pkg/db/client.go
```diff
@@ -198,10 +198,10 @@ func (c *ClientSqlx) execWithBackoff(query string, args ...interface{}) (res sql
 }
 
 func IsDuplicateEntryError(err error) bool {
-	if err != nil {
-		mysqlErr, ok := err.(*mysql.MySQLError)
+	mysqlErr := &mysql.MySQLError{}
 
-		return ok && mysqlErr.Number == mysqlDuplicateEntryErrorNumber
+	if errors.As(err, &mysqlErr) {
+		return mysqlErr.Number == mysqlDuplicateEntryErrorNumber
```
applike-ss, line 204:
```suggestion
		return mysqlErr.Number == mysqlerr.ER_DUP_ENTRY
```

applike-ss (changes requested):

pkg/db-repo/errors.go
ajscholl: And go needs your assistence for that as far as I read the code of the standard library.

`/usr/lib/go/src/errors/wrap.go:49`:
```go
		if x, ok := err.(interface{ Is(error) bool }); ok && x.Is(target) {
			return true
		}
```

So the library function basically walks the error chain and unwraps the errors one by one, each time checking whether this error now is the one you are looking for. The following test fails for example doesn't work anymore once you delete the implementation for `Is`:

```go
type ExampleError struct {
	Err error
}

func (f *ExampleError) Error() string {
	return fmt.Sprintf("Example: %s", f.Err.Error())
}

func (f *ExampleError) Unwrap() error {
	return f.Err
}

func (f *ExampleError) Is(err error) bool {
	_, ok := err.(*ExampleError)

	return ok
}

func TestExampleError(t *testing.T) {
	err := fmt.Errorf("an error")
	err = &ExampleError{
		Err: err,
	}
	assert.True(t, errors.Is(err, &ExampleError{}))
}
```

pkg/db/client.go
```diff
@@ -198,10 +198,10 @@ func (c *ClientSqlx) execWithBackoff(query string, args ...interface{}) (res sql
 }
 
 func IsDuplicateEntryError(err error) bool {
-	if err != nil {
-		mysqlErr, ok := err.(*mysql.MySQLError)
+	mysqlErr := &mysql.MySQLError{}
```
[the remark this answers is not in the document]
ajscholl: I think the error is represented by `*mysql.MySQLError` and you need a pointer to the error type you are looking for, so `**mysql.MySQLError` in the end is correct.

ajscholl (approved):

pkg/db-repo/repository.go
```diff
@@ -126,6 +134,14 @@ func (r *repository) Update(ctx context.Context, value ModelBased) error {
 
 	err := r.orm.Save(value).Error
 
+	isDuplicate := db.IsDuplicateEntryError(err)
+	if isDuplicate {
+		logger.Warnf("could not update model of type %s with id %d due to duplicate entry error: %s", modelId, *value.GetId(), err.Error())
```
applike-ss, line 139:
```suggestion
		logger.Warnf("could not update model of type %s with id %d due to duplicate entry error: %s", modelId, mdl.EmptyUintIfNil(value.GetId()), err.Error())
```

pkg/db-repo/repository.go
```diff
@@ -126,6 +134,14 @@ func (r *repository) Update(ctx context.Context, value ModelBased) error {
...
+		return &db.DuplicateEntryError{
+			Err: err,
+		}
+	}
+
 	if err != nil {
 		logger.Errorf(err, "could not update model of type %s with id %d", modelId, *value.GetId())
```
applike-ss, line 146:
```suggestion
		logger.Errorf(err, "could not update model of type %s with id %d", modelId, mdl.EmptyUintIfNil(value.GetId()))
```

applike-ss: maybe also check other pointer derefs for safeness. I know that there shouldn't be a model passed to Update that doesn't have a Id, but you could

ajscholl: I think you should not get past this error check should your id be nil. Either it is nil and then it is treated like a create by `Save` or it is not nil and then the problem does not exist. In case of a create, we either succeed and get an id set, or we fail and have one of the two error cases. So I agree, we should guard against nil at the two places you mentioned, but there shouldn't be any other possible places.

pkg/db-repo/repository.go
```diff
@@ -89,7 +91,14 @@ func (r *repository) Create(ctx context.Context, value ModelBased) error {
 
 	err := r.orm.Create(value).Error
 
-	// TODO: Check for duplicate error and return a local error type instead which applications can handle themselves
+	isDuplicate := db.IsDuplicateEntryError(err)
+	if isDuplicate {
```
j4k4, line 95: i would it write as `if db.IsDuplicateEntryError(err) {`  
if the condition statement gets more complex it makes sense to have it on a separate line but for simple ones i find it more readable to have it inline..

pkg/db-repo/repository.go
```diff
@@ -126,8 +135,16 @@ func (r *repository) Update(ctx context.Context, value ModelBased) error {
 
 	err := r.orm.Save(value).Error
 
+	isDuplicate := db.IsDuplicateEntryError(err)
+	if isDuplicate {
```
j4k4, line 139: same here

applike-ss (approved):

PR #385, opened 2020-12-21
apiserver, db-repo: handle record not found errors;

ajscholl:
Instead of raising an error and returning a 500 Internal Server Error,
we now return a 404 Not Found should a missing record be requested by a
client.

### DIFF
--- a/pkg/apiserver/handler.go
+++ b/pkg/apiserver/handler.go
@@ -21,6 +21,7 @@ const (
 )
 
 var ErrAccessForbidden = errors.New("cant access resource")
+var ErrRecordNotFound = errors.New("record not found")
 
 type Request struct {
 	Header   http.Header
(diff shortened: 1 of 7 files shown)


pkg/apiserver/handler.go
```diff
@@ -277,10 +278,12 @@ func handle(ginCtx *gin.Context, handler HandlerWithoutInput, input interface{},
 	resp, err := handler.Handle(reqCtx, request)
 
 	if errors.Is(err, ErrAccessForbidden) {
-		handleForbidden(ginCtx, errHandler, http.StatusForbidden, gin.Error{
-			Err:  err,
-			Type: gin.ErrorTypePrivate,
-		})
+		handleUserError(ginCtx, errHandler, http.StatusForbidden, err)
```
applike-ss, line 281: what difference does this make?

ajscholl: I renamed the function because it is basically `handleError`, but without attaching the error to the gin context (which would then later end up in the logger middleware) and therefore intended for errors not caused by us and therefore not needed to be resolved by us (how can we resolve a user requesting the wrong record / having no access? you resolve this in a dashboard you are building, which should be handling the error itself).
Pushed another commit as `handleError` could be simplified a lot after thinking about this

PR #388, opened 2020-12-23
test: added wiremock component;

applike-ss:


pkg/test/env/factory_wiremock.go
```diff
@@ -0,0 +1,85 @@
...
+	component := &wiremockComponent{
+		logger:  logger,
+		binding: container.bindings["8080/tcp"],
+	}
+
+	s := settings.(*wiremockSettings)
+	jsonStr, err := ioutil.ReadFile(s.Mocks)
+
+	if err != nil {
+		return nil, fmt.Errorf("could not read http mock configuration: %w", err)
```
ajscholl, line 70: Can we in case of an error try to get an absolute path for the file? Because if you can't read it and see the absolute path, you at least should know easily what is going wrong.

pkg/test/env/factory_wiremock.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	url := f.getUrl(container.bindings["8080/tcp"])
+	_, err = http.Post(url+"/mappings/import", "application/json", bytes.NewBuffer(jsonStr))
```
ajscholl, line 74: What happens if that request returns an error response?

pkg/test/env/factory_wiremock.go
```diff
@@ -0,0 +1,85 @@
...
+		ExpireAfter: s.ExpireAfter,
+	}
+}
+
+func (f *wiremockFactory) HealthCheck(_ interface{}) ComponentHealthCheck {
+	return func(container *container) error {
+		binding := container.bindings["8080/tcp"]
+		url := f.getUrl(binding)
+
+		_, err := http.Get(url)
```
ajscholl, line 54: So responding with 500 is a healthy container?

pkg/test/component_mysql.go
```diff
@@ -40,6 +40,10 @@ func (m *mysqlComponentLegacy) Start() error {
 		"MYSQL_ROOT_PASSWORD=gosoline",
 	}
 
+	if len(m.settings.Tmpfs) == 0 {
+		m.settings.Tmpfs["/var/lib/mysql"] = ""
```
ajscholl, line 44: This doesn't seem to have anything with the wiremock to do. Why do we need that? Can we add a comment here?

pkg/test/env/factory_wiremock.go
applike-ss: The whole component will not come up and due to that will error out here: pkg/test/mocks.go:56

pkg/test/component_mysql.go
applike-ss: you are right, it is only related to the mysql mock (the old one) and is basically making both mysql mocks act the same. The new one does also have this tmpfs by default.

pkg/test/env/factory_wiremock.go
ajscholl: I don't see how that is related. When testing it locally, my wiremock stayed healthy even though the import failed with status 422 (I did sent invalid JSON).

pkg/test/env/factory_wiremock.go
```diff
@@ -0,0 +1,97 @@
...
+		ExpireAfter: s.ExpireAfter,
+	}
+}
+
+func (f *wiremockFactory) HealthCheck(_ interface{}) ComponentHealthCheck {
+	return func(container *container) error {
+		binding := container.bindings["8080/tcp"]
+		url := f.getUrl(binding)
+
+		resp, err := http.Get(url)
```
ajscholl, line 55: For me, `/__admin` (which we get from `getUrl`) redirects to `/__admin/`, only then do I get a 200 response. Not quite sure how much we need to check the state of wiremock itself, but it sounds like it would be better to get a 200 response than a redirect.

ajscholl (approved):

PR #390, opened 2021-01-04
github: paralellize action execution;

ss89:


ss89:
@selm0 mind a quick review?

.github/workflows/go.yml
```diff
@@ -9,40 +9,97 @@ on:
...
+
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v2
+
+      - name: Get dependencies
+        run: go get -v -t -d ./...
+
       - name: Integration tests
-        run: go test -p 1 --tags='integration,fixtures' ./test/...
+        run: go test -p 3 --tags='integration,fixtures' ./test/...
```
selm0, line 105: Why 3? Maybe we can rely on GOMAXPROCS?

selm0 (changes requested):
Please double check the number of go procs for the integration test.

.github/workflows/go.yml
applike-ss: i tried, but then in fact the github runner is so busy that it can't finish the tests at all

selm0 (approved):

PR #396, opened 2021-01-11
Logger mock refactor and producer daemon fix

ajscholl:
mon/mocks: add common error detection to logger mock;

Added `NewLoggerMockedLevel` to allow specifying the maximum log level
you expect to be logged. A common use case is a test where a component
might normally swallow an error, but you need an easy way to ensure no
such error was thrown and logged by your test. Additionally the logger
mock now checks the format of the message and, if formatting marks are
detected, the use of Infof/Warnf/... is enforced. If no such marks are
detected, Info/Warn/... should be used instead.

stream: handle application termination in producer daemon;

Upon termination of the application, we start shutting down the producer
daemon and cancel all contexts. However, we have to delay cancling the
context used to write the message to the output (by one second by
default) to allow the message to still be written. Should we fail to
write the message in that timeframe, we handle the resulting cancel
error by logging a warning.

### DIFF
--- a/pkg/stream/producer_daemon.go
+++ b/pkg/stream/producer_daemon.go
@@ -6,6 +6,7 @@ import (
 	"github.com/applike/gosoline/pkg/cfg"
 	"github.com/applike/gosoline/pkg/clock"
 	"github.com/applike/gosoline/pkg/coffin"
+	"github.com/applike/gosoline/pkg/exec"
 	"github.com/applike/gosoline/pkg/kernel"
 	"github.com/applike/gosoline/pkg/mon"
 	"sync"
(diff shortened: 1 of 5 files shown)


pkg/stream/producer_daemon.go
```diff
@@ -28,6 +29,7 @@ type AggregateMarshaller func(body interface{}, attributes ...map[string]interfa
 type ProducerDaemonSettings struct {
 	Enabled           bool                   `cfg:"enabled" default:"false"`
 	Interval          time.Duration          `cfg:"interval" default:"1m"`
+	CancelDelay       time.Duration          `cfg:"cancel_delay" default:"1s"`
```
j4k4, line 32: i'm wondering if we would want to change the structure of the config here.. something like:

```
stream:
  producer:
    my-producer:
      daemon:
        enabled: true
        exec:
          cancel_delay: 1s
```

To make it more clear, that we're configuring something in the `exec` space..  
In the end this would mean, that we should rename our other `backoff` conig keys sometime in the future, too.

j4k4 (changes requested):

pkg/stream/producer_daemon.go
ajscholl: Actually I think I will just remove that again... there is already a cancel delay on the output (provided by the stream package), that should be enough. Not sure I follow you on the renaming of `backoff` to `exec`. Having `backoff` as a child key of `exec` would make more sense to me there as it describes a single executor out of the `exec` package which is easy to find... Although one could think about which fields should be at `exec` and which at `exec.backoff` then.

j4k4: okay, we don't need to have this discussion now, if you remove the cancel_delay here..

j4k4 (approved):
+1

PR #398, opened 2021-01-12
improve handling of db-repo record not found errors

erdac:


pkg/db-repo/repository.go
```diff
@@ -27,6 +27,28 @@ const (
...
+func NewRecordNotFoundError(id uint, modelId string, err error) RecordNotFoundError {
+	return RecordNotFoundError{
+		Id:      id,
+		ModelId: modelId,
+		Err:     err,
+	}
+}
+
+func (e RecordNotFoundError) Error() string {
+	return fmt.Sprintf("could not find model of type %s with id %d", e.ModelId, e.Id)
```
j4k4, line 45: you could also add info of the nested error here: `fmt.Sprintf("could not find model of type %s with id %d: %s", e.ModelId, e.Id, e.Err)`

pkg/db-repo/repository.go
```diff
@@ -27,6 +27,28 @@ const (
 var operations = []string{Create, Read, Update, Delete, Query}
 var RecordNotFound = gorm.ErrRecordNotFound
```
j4k4, line 28: we should try to remove the usage of the `RecordNotFound` error and replace it with the new custom error..

pkg/apiserver/middleware_logger.go
```diff
@@ -59,6 +61,11 @@ func LoggingMiddleware(logger mon.Logger) gin.HandlerFunc {
 		}
 
 		for _, e := range ginCtx.Errors {
+			var notFound db_repo.RecordNotFoundError
```
j4k4, line 64: the logger middleware shouldn't be aware of these special cases..  
and there is no need for it, if we handle it in the crud handler the right way..

pkg/apiserver/handler.go
```diff
@@ -284,6 +285,15 @@ func handle(ginCtx *gin.Context, handler HandlerWithoutInput, input interface{},
 		return
 	}
 
+	var notFound db_repo.RecordNotFoundError
```
j4k4, line 288: this is one step to deep down in our abstractions..  
instead it should be handled in the `pkg/apiserver/crud/read.go` handler..  
`apiserver/crud` is connected to the `db_repo` package while the underlying `apiserver` package is not. we should keep it this way.

j4k4 (changes requested):

j4k4 (approved):
+1

PR #405, opened 2021-01-15
db-repo: return custom RecordNotFoundError in query function

erdac:


pkg/db-repo/repository.go
```diff
@@ -27,12 +27,12 @@ const (
...
 type RecordNotFoundError struct {
-	Id      uint
-	ModelId string
+	Id      *uint
+	ModelId *string
 	Err     error
 }
 
-func NewRecordNotFoundError(id uint, modelId string, err error) RecordNotFoundError {
+func NewRecordNotFoundError(id *uint, modelId *string, err error) RecordNotFoundError {
```
ajscholl, line 35: I think I would either create two constructor methods, one with only a single error parameter, one with non-pointer arguments, or even think about having a wrapper type or so...

pkg/db-repo/repository.go
```diff
@@ -27,12 +27,12 @@ const (
 var operations = []string{Create, Read, Update, Delete, Query}
 
 type RecordNotFoundError struct {
-	Id      uint
-	ModelId string
+	Id      *uint
```
ajscholl, line 30: I think I would make these fields private, if you need to access them, I would provide accessor methods for them.

pkg/db-repo/repository.go
erdac: jan suggested a different custom error for this use case.
so I guess a `var NoResultError =  errors.New("no results found")` should actually be enough since I don't need to wrap the error in this case anyway?

ajscholl: ah, is `Query` method always working with slices or can we actually get a record not found error there? Because if nothing matches the filter criteria, I would expect to get an empty slice back if I can get multiple

erdac: err := db.Find(result).Error does actually return a gorm.ErrRecordNotFound in that case

erdac: at least gorm v1, which we are using does that:
https://gorm.io/docs/error_handling.html
`With the previous version, ErrRecordNotFound was returned with Find, but not for v2. And I am trying to update GORM from v1 to v2. These change makes me sick...`

j4k4: yeah, i was wondering why gorm returns a RecordNotFoundError on query..  
i would expect that only if i try to read a specific record by id..  
on query i would just expect an empty result slice if no records match the criteria..  
if we would want to return an error, i would suggest another custom error.. not RecordNotFoundError..

j4k4: @ajscholl What would your favorite? Just returning an empty slice without an error or returning a `NoResultsError`?

ajscholl: I would expect an empty slice. Looking at this, I actually found at least some users of `Query` which expect an empty slice upon no results and treat each error as "the database failed". Just consider each CRUD list handler:

```go
func (h CrudHandlerV0) List(ctx context.Context, qb *db_repo.QueryBuilder, apiView string) (interface{}, error) {
	result := make([]*Model, 0)
	err := h.repo.Query(ctx, qb, &result)

	if err != nil {
		return nil, err
	}

	out := make([]interface{}, 0)
	for _, res := range result {
		toRes, _ := h.TransformOutput(res, apiView)
		out = append(out, toRes)
	}

	return out, nil
}
```

This pattern actually makes me unsure whether we currently really return an error in that case, as this would raise an internal server error otherwise every time we get no results back. So not 100% sure why this currently works, but in my opinion it should continue to work in the future.

j4k4: yeah, i would also go with empty slice / no error on no records found by query..

j4k4 (approved):
+1

PR #406, opened 2021-01-15
errors: add retry logic for eof errors

justIrina:


pkg/exec/error.go
```diff
@@ -128,3 +128,19 @@ func IsTimeOutError(err error) bool {
 
 	return false
 }
+
+func CheckEOFError(_ interface{}, err error) ErrorType {
```
j4k4, line 132: We have here `pkg/exec/error.go:105` an EOF check already..  
Why is it not working anymore?

PR #407, opened 2021-01-18
Application add version flag

ss89:


pkg/application/options.go
```diff
@@ -305,3 +319,18 @@ func WithUTCClock(useUTC bool) Option {
 		})
 	}
 }
+
+func WithVersion(version string) Option {
+	return func(app *App) {
+		app.addSetupOption(func(config cfg.GosoConf, logger mon.GosoLog) error {
```
j4k4, line 325: feels weird to use a setup option to log something..

pkg/application/options.go
```diff
@@ -262,6 +262,20 @@ func WithLoggerTagsFromConfig(app *App) {
 	})
 }
 
+func WithLoggerVersionTag(version string) Option {
```
j4k4, line 265: where is this requirement coming from?

ss89: from me :-) private project where the user of the app should somehow be able to know whether he is using an up-to-date version without the app having to call an api endpoint to determine that.

pkg/application/options.go
ss89: Please #BeDirect and let me know where you would put it

j4k4: well, i don't have always a better idea right away, but my first impulse would be something like a `-version` flag for your command..

applike-ss:
closing this, as it will never get merged and go 1.18 does now include versioning information in go builds unless specified otherwise.

PR #409, opened 2021-01-19
stream: cleaned up kinesis output error handling;

j4k4:
improve logging for write error cases

### DIFF
--- a/pkg/stream/output_kinesis.go
+++ b/pkg/stream/output_kinesis.go
@@ -2,6 +2,7 @@ package stream
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"github.com/applike/gosoline/pkg/cfg"
 	"github.com/applike/gosoline/pkg/cloud"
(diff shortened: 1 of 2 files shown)


pkg/stream/output_kinesis.go
```diff
@@ -111,17 +109,19 @@ func (o *kinesisOutput) writeBatch(ctx context.Context, batch [][]byte) error {
...
 	backoffConfig.MaxElapsedTime = 15 * time.Minute
 
-	err := backoff.Retry(func() (err error) {
+	err := backoff.RetryNotify(func() (err error) {
 		records, err = o.putRecordsAndCollectFailed(ctx, records)
 
 		return err
-	}, backoffConfig)
+	}, backoffConfig, func(err error, duration time.Duration) {
+		o.logger.Warnf("retrying put records after facing error: %s", err)
```
ajscholl, line 117: I would like to have the following:
- a unique id so we can correlate all messages from a single batch write
- a success message after it failed once, I think `exec` already does something like this, right?

That would help us ensure all failures are matched with later successes

ajscholl (approved):

PR #412, opened 2021-01-22
ddb: implement transactions;

selm0:
AWS DynamoDB supports transactions since early 2020.
This PR adds basic support for the different methods.


pkg/ddb/builder.go
```diff
@@ -32,7 +32,7 @@ func (b *keyBuilder) buildKey(item interface{}) (KeyValues, error) {
 
 func (b *keyBuilder) fromItem(item interface{}) (KeyValues, error) {
 	if item == nil {
-		return nil, fmt.Errorf("can not build key attributes from nil item")
+		return nil, fmt.Errorf("can not build key attributes from nil Item")
```
j4k4, line 35: is there a reason why "Item" is now ucfirst everywhere? :thinking:

pkg/ddb/builder_conditioncheck.go
```diff
@@ -0,0 +1,121 @@
...
+
+type conditionCheckBuilder struct {
+	condition  *expression.ConditionBuilder
+	err        error
+	metadata   *Metadata
+	keyBuilder keyBuilder
+	returnType *string
+}
+
+func NewconditionCheckBuilder(metadata *Metadata) ConditionCheckBuilder {
```
j4k4, line 30: should be `NewConditionCheckBuilder`?

pkg/ddb/builder_putitem.go
```diff
@@ -48,7 +50,7 @@ func (b *putItemBuilder) ReturnAllOld() PutItemBuilder {
 
 func (b *putItemBuilder) Build(item interface{}) (*dynamodb.PutItemInput, error) {
 	if b.returnType != nil && *b.returnType != dynamodb.ReturnValueNone && !isPointer(item) {
-		return nil, fmt.Errorf("the provided old value has to be a pointer")
+		return nil, errors.New("the provided old value has to be a pointer")
```
j4k4, line 53: why changing this from `fmt.Errorf` to `errors.New`?  
i like having a consistent `fmt` usage. Is there a benefit despite a little performance advantage, which should be neglectable as the error should not happen in the first place? :D

pkg/ddb/builder.go
selm0: overeager refactoring of my IDE. Should be reverted

pkg/ddb/builder_putitem.go
selm0: No, it's only about using the right tool for the right purpose (and it's a little bit faster)

j4k4: for me, consistency would be the winner here.. also one less import..  
but also no hard feelings about it, if you want to keep it like this..

PR #413, opened 2021-01-22
stream: switch order of go routines started in coffin to avoid premature termination;

ajscholl:
In an integration test you could easily have an input which quickly
finishes its `Run` method. With a little bit of bad luck your pipeline
has not yet started the read routine and therefore the coffin dies
before you started all go routines. Switching the order avoids this as
`read` waits until the input produced data or stopped.


j4k4 (approved):
+1

PR #416, opened 2021-01-25
tf: add success rate alarms for consumer and gateway

justIrina:


deployments/terraform/modules/consumer/alarms.tf
```diff
@@ -0,0 +1,46 @@
+resource "aws_cloudwatch_metric_alarm" "success-rate" {
```
j4k4, line 1: the module names are quite generic right know.  
we also have to consider, that there are different consumers than sqs.  
i would suggest "grouping" the alarms with an "alarm-" prefix and to include input type.  
eg: `alarm_consumer_sqs`

deployments/terraform/modules/consumer/variables.tf
```diff
@@ -0,0 +1,40 @@
+variable "project" {}
+variable "environment" {}
+variable "family" {}
+variable "application" {}
+
+variable "alarm_create" {
```
j4k4, line 6: if the module is alarm context only, we can skip the alarm prefix for the variables i guess.

deployments/terraform/modules/consumer/variables.tf
```diff
@@ -0,0 +1,40 @@
...
+  description = "Required percentage of successful requests"
+}
+
+variable "alarm_period" {
+  type        = number
+  default     = 300
+  description = "The period in seconds over which the specified statistic is applied"
+}
+
+variable "model" {
```
j4k4, line 36: i would move the model variable up top and group it together with the other application identifiers as they are all identifying the "source" of this alarm.

deployments/terraform/modules/gateway/alarms.tf
```diff
@@ -0,0 +1,46 @@
+resource "aws_cloudwatch_metric_alarm" "success-rate" {
```
j4k4, line 1: i would suggest module name: `alarm_gateway`

deployments/terraform/modules/gateway/alarms.tf
```diff
@@ -0,0 +1,46 @@
...
+
+  metric_query {
+    id          = "requests"
+    return_data = false
+
+    metric {
+      dimensions = {
+        "path" = var.api_request_path
+      }
+      metric_name = "ApiRequestCount"
```
j4k4, line 18: i tend to go with the request count from the load balancer, as we would list all paths from a gateway here as we have quite some gateways with multiple paths.

j4k4 (changes requested):

deployments/terraform/modules/alarm_gateway/variables.tf
```diff
@@ -0,0 +1,46 @@
...
+  description = "Required percentage of successful requests"
+}
+
+variable "period" {
+  type        = number
+  default     = 300
+  description = "The period in seconds over which the specified statistic is applied"
+}
+
+variable "api_request_path" {
```
j4k4, line 36: this one isn't need anymore, right?

deployments/terraform/modules/alarm_gateway/variables.tf
```diff
@@ -0,0 +1,46 @@
...
+  type        = string
+  default     = null
+  description = "Api request path"
+}
+
+variable "aws_lb_application_arn_suffix" {
```
j4k4, line 42: is it a suffix or do we provide the full arn?

justIrina: according to https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/cloudwatch_metric_alarm examples we have to provide suffix
also here https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-cloudwatch-metrics.html when load balancer dimenssion is used it isnt' and arn

j4k4: okay, it's the "final portion of the load balancer ARN" :+1:

j4k4 (approved):

j4k4 (approved):
+1

PR #418, opened 2021-01-26
tf: add message age sqs alarm

justIrina:


deployments/terraform/modules/sqs/queue/variables.tf
```diff
@@ -37,10 +37,12 @@ variable "alarm_period" {
   default = 300
 }
 
-variable "alarm_threshold" {
+variable "alarm_visible_messages_threshold" {
   default = 200
 }
 
+variable "alarm_message_age_threshold" {}
```
j4k4, line 44: let's try a default of 60 seconds here

deployments/terraform/modules/sqs/queue/main.tf
```diff
@@ -29,7 +29,30 @@ resource "aws_cloudwatch_metric_alarm" "number-of-visible-messages" {
...
+
+  dimensions = {
+    QueueName = aws_sqs_queue.main.name
+  }
+
+  alarm_actions = ["arn:aws:sns:eu-central-1:164105964448:${var.project}-${var.environment}-${var.family}-alarm"]
+  ok_actions    = ["arn:aws:sns:eu-central-1:164105964448:${var.project}-${var.environment}-${var.family}-alarm"]
+}
+
+resource "aws_cloudwatch_metric_alarm" "oldest_message_age" {
```
j4k4, line 44: i would go without the alarm actions for the new one to run a test first..
we don't want to trigger alarms here to annoy everyone.. :D

deployments/terraform/modules/sqs/main.tf
```diff
@@ -38,9 +39,10 @@ module "dead" {
...
 
-  alarm_create              = var.alarm_dead_create
-  alarm_period              = var.alarm_dead_period
-  alarm_threshold           = var.alarm_dead_threshold
-  alarm_evaluation_periods  = var.alarm_dead_evaluation_periods
-  alarm_datapoints_to_alarm = var.alarm_dead_datapoints_to_alarm
+  alarm_create                    = var.alarm_dead_create
+  alarm_period                    = var.alarm_dead_period
+  alarm_visible_message_threshold = var.alarm_dead_threshold
+  alarm_message_age_threshold     = var.alarm_dead_message_age_threshold
```
j4k4, line 45: for the dead letter queue we don't need an age based alarm..  
message in the DLQ will always age as they are not consumed on a regular basis. it's more of a backup of failed messages.

j4k4 (changes requested):

deployments/terraform/modules/sqs/main.tf
```diff
@@ -19,11 +19,13 @@ module "main" {
...
-  alarm_threshold           = var.alarm_main_threshold
-  alarm_evaluation_periods  = var.alarm_main_evaluation_periods
-  alarm_datapoints_to_alarm = var.alarm_main_datapoints_to_alarm
+  alarm_create                    = var.alarm_main_create
+  alarm_period                    = var.alarm_main_period
+  alarm_visible_message_threshold = var.alarm_main_threshold
+  alarm_evaluation_periods        = var.alarm_main_evaluation_periods
+  alarm_datapoints_to_alarm       = var.alarm_main_datapoints_to_alarm
+  alarm_message_age_create        = var.alarm_main_message_age_create
+  alarm_message_age_threshold     = var.alarm_message_age_threshold
```
j4k4, line 28: would still use `var.alarm_main_message_age_threshold` to make it clear that the variable is for the main queue

deployments/terraform/modules/sqs/variables.tf
```diff
@@ -64,3 +64,15 @@ variable "alarm_dead_evaluation_periods" {
 variable "alarm_dead_datapoints_to_alarm" {
   default = 1
 }
+
+variable "alarm_main_message_age_create" {
+  default = 1
+}
+
+variable "alarm_dead_message_age_create" {
```
j4k4, line 72: this one is obsolete?

deployments/terraform/modules/sqs/queue/main.tf
```diff
@@ -40,3 +40,23 @@ resource "aws_cloudwatch_metric_alarm" "number-of-visible-messages" {
   alarm_actions = ["arn:aws:sns:eu-central-1:164105964448:${var.project}-${var.environment}-${var.family}-alarm"]
   ok_actions    = ["arn:aws:sns:eu-central-1:164105964448:${var.project}-${var.environment}-${var.family}-alarm"]
 }
+
+resource "aws_cloudwatch_metric_alarm" "oldest_message_age" {
+  alarm_name = "${var.family}-${var.application}-${var.queueName}-approximate-age-of-oldest-message"
```
j4k4, line 45: would go with `alarm_name = "${var.family}-${var.application}-${var.queueName}-messages-too-old"`

j4k4 (changes requested):

deployments/terraform/modules/sqs/variables.tf
justIrina: Why? by default for queue var alarm_message_age_create is 1, so if we dont overwrite it with 0 in sqs part it will be created, when actually should not.

j4k4: okay, i guess i would have set it to `0` here: https://github.com/applike/gosoline/blob/1e0f13e98e9b0e082e995b3813618251f42159f9/deployments/terraform/modules/sqs/main.tf#L47  

but we can also keep your approach

deployments/terraform/modules/sqs/queue/main.tf
```diff
@@ -40,3 +40,23 @@ resource "aws_cloudwatch_metric_alarm" "number-of-visible-messages" {
   alarm_actions = ["arn:aws:sns:eu-central-1:164105964448:${var.project}-${var.environment}-${var.family}-alarm"]
   ok_actions    = ["arn:aws:sns:eu-central-1:164105964448:${var.project}-${var.environment}-${var.family}-alarm"]
 }
+
+resource "aws_cloudwatch_metric_alarm" "oldest_message_age" {
+  alarm_name = "${var.family}-${var.application}-${var.queueName}-message-too-old"
```
j4k4, line 45: i would prefer plural here: `messages-too-old`..  
sry for the nitpicking.. ;)

j4k4 (changes requested):

j4k4 (approved):

PR #433, opened 2021-02-05
tf: use create parameter to determine alarm creation rule

justIrina:


ss89:
maybe you could rename this pr to something more generic (as this feature is)

j4k4 (approved):

PR #441, opened 2021-02-18
deployments: added family tag to ec2 instances;

### LINKED ISSUE #440: deployments: Add tag to ec2 instances: family
We should make use of the family tag that we already get when creating an ecs cluster.

Add the Tag to the ec2 instance.

applike-ss:
closes #440


mfroembgen (commented):
Go for it

PR #445, opened 2021-02-23
ddb: improved error handling on NewRepository

j4k4:
- removed call to logger.Panic
    - added error return

### DIFF
--- a/pkg/kernel/kernel.go
+++ b/pkg/kernel/kernel.go
@@ -2,7 +2,6 @@ package kernel
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"github.com/applike/gosoline/pkg/cfg"
 	"github.com/applike/gosoline/pkg/coffin"
@@ -254,8 +253,11 @@ func (k *kernel) runFactories() error {
 				}
 
 				bookLck.Lock()
-				k.addModuleToStage(container.name, module, container.opts)
-				bookLck.Unlock()
+				defer bookLck.Unlock()
+
+				if err = k.addModuleToStage(container.name, module, container.opts); err != nil {
+					return fmt.Errorf("can not add module to stage: %w", err)
+				}
 
 				return nil
 			}
@@ -267,7 +269,7 @@ func (k *kernel) runFactories() error {
 	return bootCoffin.Wait()
 }
 
-func (k *kernel) addModuleToStage(name string, module Module, opts []ModuleOption) {
+func (k *kernel) addModuleToStage(name string, module Module, opts []ModuleOption) error {
 	ms := &ModuleState{
 		Module:    module,
 		Config:    getModuleConfig(module),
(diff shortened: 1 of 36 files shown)


pkg/kernel/kernel.go
```diff
@@ -254,7 +253,9 @@ func (k *kernel) runFactories() error {
 				}
 
 				bookLck.Lock()
-				k.addModuleToStage(container.name, module, container.opts)
+				if err = k.addModuleToStage(container.name, module, container.opts); err != nil {
```
selm0, line 256: Do you need to unlock also in case of an error?

pkg/kvstore/configurable.go
```diff
@@ -75,23 +73,32 @@ func newKvStoreChainFromConfig(config cfg.Config, logger mon.Logger, name string
...
+			if err := store.Add(NewInMemoryKvStore); err != nil {
+				return nil, fmt.Errorf("can not add inMemory store: %w", err)
+			}
 		case TypeRedis:
-			store.Add(NewRedisKvStore)
+			if err := store.Add(NewRedisKvStore); err != nil {
+				return nil, fmt.Errorf("can not add redis store: %w", err)
+			}
 		default:
 			err := fmt.Errorf("invalid element type %s for kvstore chain", element)
```
selm0, line 95: Can't you pass the error back in the default cause instead of `Fatalf`ing?

pkg/mdlsub/output_ddb.go
```diff
@@ -16,33 +16,36 @@ func init() {
...
-		return ddb.NewMetricRepository(config, logger, repo)
+		return ddb.NewMetricRepository(config, logger, repo), nil
 	}
 }
 
-func outputDdbFactory(config cfg.Config, logger mon.Logger, settings *SubscriberSettings, transformers VersionedModelTransformers) map[int]Output {
+func outputDdbFactory(config cfg.Config, logger mon.Logger, settings *SubscriberSettings, transformers VersionedModelTransformers) (map[int]Output, error) {
 	outputs := make(map[int]Output)
 
 	for version := range transformers {
```
selm0, line 40: Does it make sense to already add the error to the different outputs?

selm0 (changes requested):

pkg/kvstore/configurable.go
j4k4: good catch

pkg/kernel/kernel.go
j4k4: good catch

pkg/mdlsub/output_ddb.go
j4k4: What do you mean?

selm0: You're adding the error to the outputFactory, but are not passing anything from the actual output.

selm0 (approved):

Issue #448, opened 2021-03-02
Add a fixture loader for S3


applike-ss:
For loading data e.g. for integration tests or for local development we added fixture loaders.

Currently we do have a fixture loader for redis/redis-kvstore/ddb/ddb-kvstore/mysql-orm/mysql-plain, but not for S3.

However that would be very handy for testing certain scenarios.

Check these for reference:
* pkg/fixtures:
  * writer_ddb.go
  * writer_ddb_kvstore.go
  * writer_mysql_orm.go
  * writer_mysql_plain.go
  * writer_redis.go
  * writer_redis_kvstore.go
  * purger_dynamodb.go
  * purger_mysql.go
  * purger_redis.go
* test:
  * fixtures_dynamodb_test.go
  * fixtures_mysql_test.go
  * fixtures_redis_test.go

The task is:
* Add a fixture loader for S3
* Add a purger for the S3 Fixture loader
* Add an integration test that validates the loader works
* Add an integration test that validates the purger works

ss89:
closed by #452

PR #452, opened 2021-03-12
fixtures: add fixture loader for S3;

mohamed-justdice:
This adds fixture loader for S3 and integration tests for it.


pkg/s3/client.go
```diff
@@ -0,0 +1,52 @@
...
+	instance map[string]s3iface.S3API
+}{}
+
+type Settings struct {
+	BasePath string
+	BucketName string
+	Client cloud.ClientSettings
+}
+
+func ProvideClient(config cfg.Config, logger mon.Logger, settings *Settings) s3iface.S3API {
```
j4k4, line 29: can we move the s3 client to `gosoline/pkg/cloud/aws_clients.go` ?

mohamed-justdice: Why? This is following the project layering. Basically the client here is S3 specific. So no, I would leave it here, no?

j4k4: The sns and sqs stuff will also be moved to a more aws related space in the future. Most likely be done while switching to the aws sdk v2. Furthermore i would like to keep the root packages free of any cloud provider specific packages and rather group them in the cloud package.

mohamed-justdice: okay, moving the s3 folder form `pkg` to `pkg/cloud/aws` should be right to do then right?

j4k4: yes, that would also be fine :+1:

mohamed-justdice: And that is done 👍

test/fixtures_s3_test.go
```diff
@@ -0,0 +1,165 @@
...
+
+func TestFixturesS3Suite(t *testing.T) {
+	suite.Run(t, new(FixturesS3Suite))
+}
+
+func (s FixturesS3Suite) TestS3() {
+	err := s.loader.Load(s3DisabledPurgeFixtures())
+	s.NoError(err)
+
+	goi := &aws_s3.GetObjectInput{
```
applike-ss, line 68:
```suggestion
	input := &aws_s3.GetObjectInput{
```

test/fixtures_s3_test.go
```diff
@@ -0,0 +1,165 @@
...
+		Bucket: aws.String(BucketName),
+		Key:    aws.String(fmt.Sprintf("%s/HappyFace.jpeg", BasePath)),
+	}
+	goo, err := s.client.GetObject(goi)
```
applike-ss, line 72:
```suggestion
	output, err := s.client.GetObject(goi)
```

test/fixtures_s3_test.go
```diff
@@ -0,0 +1,165 @@
...
+			Writer: fixtures.S3FixtureWriterFactory(&s3.Settings{
+				BucketName: BucketName,
+				BasePath: BasePathPurge,
+			}),
+			Fixtures: nil,
+		},
+	}
+}
+
+func (s FixturesS3Suite) s3Config() map[string]interface{} {
```
applike-ss, line 158: in fact this shouldn't be necessary, was it for you? The (new-style) integration tests usually replace endpoints properly. Maybe have a quick look for how it works for dynamodb or sqs and see if there is adjustments needed in the integration test execution code.

applike-ss (changes requested):

docs/fixtures/readme.md
```diff
@@ -71,6 +72,37 @@ func main() {
...
 ```
+or for example when using the `S3FixtureWriterFactory`:
+```
+//+build fixtures
+
+package main
+
+import (
+    ...
+    gosoline_s3 "github.com/applike/gosoline/pkg/cloud/aws/s3"
```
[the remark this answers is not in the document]
mohamed-justdice: I don't use aliases when not needed (only if IDE auto-generate it). But here, I wanted to expose the fact that this is not the standard `aws/s3` (from aws) but from our internal `pkg`. Should I keep it or remove the alias?

pkg/fixtures/purger_s3.go
```diff
@@ -0,0 +1,44 @@
...
+	return &s3Purger{
+		logger: logger,
+		bucketName:   &settings.BucketName,
+		client: client,
+	}
+}
+
+func (p *s3Purger) purge() error {
+	p.logger.Infof("purging bucket %s", *p.bucketName)
+	iter := s3manager.NewDeleteListIterator(p.client, &aws_s3.ListObjectsInput{
```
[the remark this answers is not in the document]
mohamed-justdice: Nope a bucket needs to be emptied first. Otherwise a `ErrCodeBucketNotEmpty` will arise, and there is no other way to force delete a non empty bucket.

pkg/fixtures/purger_s3.go
```diff
@@ -0,0 +1,44 @@
+package fixtures
+
+import (
+	"github.com/applike/gosoline/pkg/cfg"
+	"github.com/applike/gosoline/pkg/cloud/aws/s3"
+	"github.com/applike/gosoline/pkg/mon"
+	"github.com/aws/aws-sdk-go/aws"
+	aws_s3 "github.com/aws/aws-sdk-go/service/s3"
```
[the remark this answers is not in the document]
mohamed-justdice: noted

pkg/cloud/aws/s3/client.go
```diff
@@ -0,0 +1,55 @@
...
+	instance map[string]s3iface.S3API
+}{}
+
+type Settings struct {
+	BasePath string
+	BucketName string
+	Client cloud.ClientSettings
+}
+
+func ProvideClient(config cfg.Config, logger mon.Logger, settings *Settings) s3iface.S3API {
```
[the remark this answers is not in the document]
mohamed-justdice: Applied

pkg/cloud/aws/s3/client.go
```diff
@@ -0,0 +1,55 @@
...
+type Client interface {
+	s3iface.S3API
+}
+
+var c = struct {
+	sync.Mutex
+	instance map[string]s3iface.S3API
+}{}
+
+type Settings struct {
```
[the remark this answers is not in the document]
mohamed-justdice: I am using the `cloud.GetAwsConfig` that takes a `cloud.ClientSettings` pointer, should I implement a new `GetAwsConfigWithoutClientSettings` or just pass a pointer to an empty `cloud.ClientSettings`?

pkg/fixtures/purger_s3.go
```diff
@@ -0,0 +1,44 @@
...
+	"github.com/aws/aws-sdk-go/service/s3/s3manager"
+)
+
+type s3Purger struct {
+	logger     mon.Logger
+	client     s3.Client
+	bucketName *string
```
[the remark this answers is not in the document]
mohamed-justdice: noted

pkg/fixtures/purger_s3.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+func newS3Purger(config cfg.Config, logger mon.Logger, settings *s3.Settings) *s3Purger {
```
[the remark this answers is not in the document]
mohamed-justdice: two other pointers are in
`s3FixtureWriter` struct. A pointer for settings and another one for the purger. I think we don't need them to be pointers. Do you approve?

pkg/fixtures/purger_s3.go
```diff
@@ -0,0 +1,44 @@
...
+	err := s3manager.NewBatchDeleteWithClient(p.client).Delete(aws.BackgroundContext(), iter)
+	if err != nil {
+		return err
+	}
+	_, err = p.client.DeleteBucket(&aws_s3.DeleteBucketInput{Bucket: p.bucketName})
+	if err != nil {
+		return err
+	}
+	p.logger.Infof("purging bucket %s done", *p.bucketName)
+	return nil
```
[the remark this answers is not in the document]
mohamed-justdice: noted

pkg/fixtures/writer_s3.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+	s.logger.Infof("Bucket %s successfully created", s.settings.BucketName)
+
+	var files []string
+	err = filepath.Walk(s.settings.BasePath, func(path string, f os.FileInfo, err error) error {
+		if !f.IsDir() {
+			if !strings.HasSuffix(path, ".DS_Store") {
+				files = append(files, path)
+			}
+		}
```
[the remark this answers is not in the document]
mohamed-justdice: applied

pkg/cloud/aws/s3/client.go
```diff
@@ -0,0 +1,55 @@
+package s3
```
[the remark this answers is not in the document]
j4k4: As discussed, i try to get rid of the cloud provider specific top level packages. So `cloud/aws/s3` is good.

pkg/fixtures/purger_s3.go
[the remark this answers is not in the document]
j4k4: The when to user pointers is a tricky one.. We usually try to avoid pointers on basic types like int, string etc.. For structs we use pointers. You will find places where they are missing but usually we pass structs via reference and also have them as reference on fields.

test/fixtures_s3_test.go
mohamed-justdice: applied

test/fixtures_s3_test.go
mohamed-justdice: applied

test/fixtures_s3_test.go
mohamed-justdice: It was necessary for me. dynamodb does it the same way. I prefer keeping it this way instead of hard coding `address:port` for tests.

j4k4: new-style integration tests are replacing endpoints properly, yes, but this is an old-style integration test, so you have to set it yourself?

test/test_configs/config.fixtures_s3.test.yml
```diff
@@ -0,0 +1,6 @@
+aws_s3_autoCreate: true
+aws_sdk_retries: 30
+aws_s3_forcePathStyle: true
+
+fixtures:
+  enabled: true
```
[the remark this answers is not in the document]
mohamed-justdice: I renamed then reused the nyan cat gif I found you already used. What readme should I mention that on?

pkg/fixtures/purger_s3.go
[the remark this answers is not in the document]
mohamed-justdice: got it. that suits me, I am used to that as well.

pkg/fixtures/purger_s3.go
```diff
@@ -0,0 +1,44 @@
...
+func (p *s3Purger) purge() error {
+	p.logger.Infof("purging bucket %s", *p.bucketName)
+	iter := s3manager.NewDeleteListIterator(p.client, &aws_s3.ListObjectsInput{
+		Bucket: p.bucketName,
+	})
+	err := s3manager.NewBatchDeleteWithClient(p.client).Delete(aws.BackgroundContext(), iter)
+	if err != nil {
+		return err
+	}
+	_, err = p.client.DeleteBucket(&aws_s3.DeleteBucketInput{Bucket: p.bucketName})
```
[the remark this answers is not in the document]
mohamed-justdice: I think this should've been mentioned in the ticket (issue). That way I'd have known that there are existing "stuff" to not re-do but use them. Now i have to read the logic there and adapt, which takes time.

pkg/cloud/aws/s3/client.go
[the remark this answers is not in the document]
applike-ss: i think you won't need all of this, if you use the blobStore instead (/pkg/blob)

test/fixtures_s3_test.go
applike-ss: in that case: why not make it a new style integration test instead?

docs/fixtures/readme.md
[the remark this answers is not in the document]
applike-ss: Ok, then please make the alias camelCase

pkg/fixtures/purger_s3.go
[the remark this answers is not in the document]
applike-ss: Ahh, it seems the --force option on the cli is not an input option for the request, didn't see that.

applike-ss:
closes #448

pkg/fixtures/purger_s3.go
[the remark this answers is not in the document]
applike-ss: I did forget it myself, sorry.

test/test_configs/config.fixtures_s3.test.yml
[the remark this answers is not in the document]
applike-ss: actually i do not remember which of the thounsand entries of google image search i used :see_no_evil:

pkg/cloud/aws/s3/client.go
```diff
@@ -0,0 +1,56 @@
...
+		return c.instance[endpoint]
+	}
+
+	c.instance[endpoint] = NewClient(config, logger, settings)
+
+	return c.instance[endpoint]
+}
+
+func NewClient(config cfg.Config, logger mon.Logger, settings *Settings) *s3.S3 {
+	s3ForcePathStyle := config.GetBool("aws_s3_forcePathStyle", true)
```
applike-ss, line 49: By default we should stick to the providers default, which would be false if i'm not mistaken.

```suggestion
	s3ForcePathStyle := config.GetBool("aws_s3_forcePathStyle", false)
```

pkg/fixtures/writer_s3.go
```diff
@@ -0,0 +1,151 @@
...
+		Bucket:        aws.String(s.settings.BucketName),
+		Key:           aws.String(filePath),
+		Body:          fileBytes,
+		ContentLength: aws.Int64(size),
+		ContentType:   aws.String(fileType),
+	}
+
+	_, err = s.client.PutObject(params)
+	c <- err
+}
```
applike-ss, line 150: please use the blob store

pkg/fixtures/writer_s3.go
```diff
@@ -0,0 +1,151 @@
...
+			case awsS3.ErrCodeBucketAlreadyOwnedByYou:
+				s.logger.Info(awsS3.ErrCodeBucketAlreadyOwnedByYou)
+
+				return nil
+			}
+		}
+	}
+
+	return err
+}
```
applike-ss, line 115: shouldn't be needed when you use the blob store

test/fixtures_s3_test.go
```diff
@@ -0,0 +1,165 @@
+//+build integration fixtures
```
applike-ss, line 1: please adapt this to a new-style integration test, see `pkg/test/suite_test.go`

applike-ss (changes requested):

test/fixtures_s3_test.go
mohamed-justdice: also here, I followed the referenced files (test files) in the ticket (issue). It'd have been nice mentioning this there. Now I'll have to re-read then re-adapt

pkg/blob/store.go
```diff
@@ -62,8 +62,10 @@ type Store interface {
 	BucketName() string
 	Copy(batch CopyBatch)
 	CopyOne(obj *CopyObject) error
+	CreateBucket()
 	Delete(batch Batch)
 	DeleteOne(obj *Object) error
+	DeleteBucket() error
```
applike-ss, line 68: please respect a-z order

pkg/fixtures/writer_s3.go
```diff
@@ -0,0 +1,103 @@
...
+	ConfigName string
+}
+
+type s3FixtureWriter struct {
+	logger      mon.Logger
+	store       blob.Store
+	batchRunner *blob.BatchRunner
+
+	settings *BlobFixturesSettings
+	purger   *s3Purger
```
applike-ss, line 24: Please follow our rules in terms of sorting fields on structs:
```
Please sort properties on a struct.
1. embedded structs/interfaces
2. other structs/interfaces
3. scalars

Within each group, sort alphabetically.

The following exceptions apply:

the logger should be listed first after all embeddables, as it's the most used dependency (functionality only)
timestamps (CreatedAt, UpdatedAt) are the last properties (entities only)
```

please also adhere to that when it comes to: initalization of a struct and interfaces in general.

pkg/fixtures/purger_s3.go
```diff
@@ -0,0 +1,45 @@
...
+		bucketName:  settings.ConfigName,
+		store:       store,
+		batchRunner: br,
+	}
+}
+
+func (p *s3Purger) purge() error {
+	ctx, cancel := context.WithCancel(context.Background())
+
+	var re error
```
applike-ss, line 32: how about this:
```suggestion
	var batchRunnerErr error
```

pkg/fixtures/writer_s3.go
```diff
@@ -0,0 +1,103 @@
...
+		batchRunner: batchRunner,
+		purger:      purger,
+	}
+}
+
+func (s *s3FixtureWriter) Purge() error {
+	return s.purger.purge()
+}
+
+func (s *s3FixtureWriter) Write(*FixtureSet) error {
```
applike-ss, line 51: not sure if this is a personal flavour, but i believe we don't have a rule for that yet.
I like writing it like so:
```suggestion
func (s *s3FixtureWriter) Write(_ *FixtureSet) error {
```
@j4k4 any feelings about this?

pkg/fixtures/writer_s3.go
```diff
@@ -0,0 +1,103 @@
...
+	s.store.CreateBucket()
+
+	var files []string
+	err := filepath.Walk(s.settings.BasePath, func(path string, f os.FileInfo, err error) error {
+		if f.IsDir() {
+			return nil
+		}
```
applike-ss, line 58: I think we should handle it, if there was an issue traversing through the file tree. Also info will be nil if err != nil. See documentation for WalkFunc.
```suggestion
		if err != nil {
			return err
		}

		if f.IsDir() {
			return nil
		}
```

pkg/fixtures/writer_s3.go
```diff
@@ -0,0 +1,103 @@
...
+
+		files = append(files, path)
+
+		return nil
+	})
+	if err != nil {
+		return err
+	}
+
```
applike-ss, line 67: early return, when possible.
```suggestion
	if err != nil {
		return err
	}

	if len(files) == 0 {
		return nil
	}
```

pkg/fixtures/writer_s3.go
```diff
@@ -0,0 +1,103 @@
...
+		buffer := make([]byte, size)
+		_, err = reader.Read(buffer)
+		if err != nil {
+			return err
+		}
+
+		object := blob.Object{
+			Key:  aws.String(file),
+			Body: blob.StreamBytes(buffer),
+		}
```
applike-ss, line 86:
```suggestion
		b, err = ioutil.ReadFile(file)
		if err != nil {
			return err
		}

		object := blob.Object{
			Key:  aws.String(file),
			Body: blob.StreamBytes(b),
		}
```

test/test_configs/config.fixtures_s3.test.yml
```diff
@@ -0,0 +1,17 @@
+env: test
+
+app_project: prj
+app_family: fam
+app_name: app
+
+aws_s3_endpoint: http://localhost:4566
+aws_s3_autoCreate: false
+aws_sdk_retries: 30
+aws_s3_forcePathStyle: true
```
applike-ss, line 10:
```suggestion
aws_s3_autoCreate: false
aws_s3_endpoint: http://localhost:4566
aws_s3_forcePathStyle: true
aws_sdk_retries: 30
```

test/fixtures_s3_test.go
```diff
@@ -0,0 +1,148 @@
...
+)
+
+const (
+	BucketName    = "s3-fixtures-test-bucket"
+	BasePath      = "test_data/s3_fixtures_test_data"
+	BasePathPurge = "test_data/s3_fixtures_purge_test_data"
+)
+
+type FixturesS3Suite struct {
+	test.Suite
```
applike-ss, line 24: this should be a `suite.Suite` instead for v3 integration tests

test/fixtures_s3_test.go
```diff
@@ -0,0 +1,148 @@
...
+
```
applike-ss, line 25: no empty lines in struct definitions please

test/fixtures_s3_test.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+}
+
+func TestFixturesS3Suite(t *testing.T) {
+	s := new(FixturesS3Suite)
+	test.RunSuite(t, s)
+}
+
+func (s FixturesS3Suite) TestS3(app test.AppUnderTest) {
+	// CHECK: Alternative for the affectations below.
```
applike-ss, line 47: you can use `s.Env().Logger()` with v3 integration tests.

test/fixtures_s3_test.go
```diff
@@ -0,0 +1,148 @@
...
+	s.logger = mon.NewLogger()
+	s.loader = fixtures.NewFixtureLoader(s.Env().Config(), s.logger)
+	s.client = blob.ProvideS3Client(s.Env().Config())
+
+	err := s.loader.Load(s3DisabledPurgeFixtures())
+	s.NoError(err)
+
+	input := &s3.GetObjectInput{
+		Bucket: aws.String(BucketName),
```
applike-ss, line 56: use this to get the bucket name instead of defining it multiple times (config+test)
```
	s.Env().Config().GetString("blobstore.test.bucket")
```

test/fixtures_s3_test.go
```diff
@@ -0,0 +1,148 @@
...
+func TestFixturesS3Suite(t *testing.T) {
+	s := new(FixturesS3Suite)
+	test.RunSuite(t, s)
+}
+
+func (s FixturesS3Suite) TestS3(app test.AppUnderTest) {
+	// CHECK: Alternative for the affectations below.
+	s.logger = mon.NewLogger()
+	s.loader = fixtures.NewFixtureLoader(s.Env().Config(), s.logger)
+	s.client = blob.ProvideS3Client(s.Env().Config())
```
applike-ss, line 50: if you implement `SetupSuite()`, then you can return a slice of Options in which you can setup your environment.

test/fixtures_s3_test.go
```diff
@@ -0,0 +1,148 @@
...
+
+	err := s.loader.Load(s3DisabledPurgeFixtures())
+	s.NoError(err)
+
+	input := &s3.GetObjectInput{
+		Bucket: aws.String(BucketName),
+		Key:    aws.String(fmt.Sprintf("%s/nyan_cat2.gif", BasePath)),
```
applike-ss, line 57: `BasePath` is a local folder whichs content is uploaded to the s3, so how would you use that path to fetch a file? It can't exist in that path.
That is:
`/test_data/s3_fixtures_test_data/nyan_cat2.gif`
vs.
`/nyan_cat2.gif`

test/fixtures_s3_test.go
```diff
@@ -0,0 +1,148 @@
...
+	s.Equal(int64(28092), *output.ContentLength)
+
+	input = &s3.GetObjectInput{
+		Bucket: aws.String(BucketName),
+		Key:    aws.String(fmt.Sprintf("%s/subDir/nyan_cat.gif", BasePath)),
+	}
+	output, err = s.client.GetObject(input)
+
+	s.NoError(err)
+	s.Equal(int64(28092), *output.ContentLength)
```
applike-ss, line 80: Since you are checking all files below a certain path, you could also use filepath.Walk here to assert the exitence of your fixtures. Maybe in combination with an expected amount of files that are supposed to be in that folder.

test/fixtures_s3_test.go
```diff
@@ -0,0 +1,148 @@
...
+		Key:    aws.String(fmt.Sprintf("%s/nyan_cat2.gif", BasePathPurge)),
+	}
+	output, err = s.client.GetObject(input)
+
+	s.NoError(err)
+	s.Equal(int64(28092), *output.ContentLength)
+
+	input = &s3.GetObjectInput{
+		Bucket: aws.String(BucketName),
+		Key:    aws.String(fmt.Sprintf("%s/nyan_cat2.gif", BasePath)),
```
applike-ss, line 116:
```suggestion
		Key:    aws.String(fmt.Sprintf("%s/subDir/nyan_cat1.gif", BasePath)),
```
that?

pkg/fixtures/purger_s3.go
mohamed-justdice: Applied.

test/fixtures_s3_test.go
mohamed-justdice: noted and applied. Although I found multiple structs having empty lines separating attributes. Is it not a good idea to separate wrapped interfaces, structs (inside a struct), exported attributes from private attributes by an empty line?

test/fixtures_s3_test.go
mohamed-justdice: Applied.

test/fixtures_s3_test.go
mohamed-justdice: Applied.

test/fixtures_s3_test.go
mohamed-justdice: Applied.

test/fixtures_s3_test.go
```diff
@@ -0,0 +1,143 @@
...
+
+func TestFixturesS3Suite(t *testing.T) {
+	s := new(FixturesS3Suite)
+	suite.Run(t, s)
+}
+
+func (s FixturesS3Suite) TestS3(app suite.AppUnderTest) {
+	s.logger = s.Env().Logger()
+	s.loader = fixtures.NewFixtureLoader(s.Env().Config(), s.logger)
+	s.client = blob.ProvideS3Client(s.Env().Config())
```
applike-ss, line 47: please implement `TestingSuiteSetupTestAware` to set the test up

test/fixtures_s3_test.go
```diff
@@ -0,0 +1,143 @@
...
+
+	err := s.loader.Load(s3DisabledPurgeFixtures())
+	s.NoError(err)
+
+	input := &s3.GetObjectInput{
+		Bucket: aws.String(s.Env().Config().GetString("blobstore.test.bucket")), //BucketName),
```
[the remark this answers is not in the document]
applike-ss: Also i suggest you to save the result of something that does stay the same when querying for it into a variable. `s.Env().Config().GetString("blobstore.test.bucket")` will never change when the app is running, so why not save it into a `bucketName` variable?

pkg/blob/store.go
```diff
@@ -62,8 +62,10 @@ type Store interface {
 	BucketName() string
 	Copy(batch CopyBatch)
 	CopyOne(obj *CopyObject) error
+	CreateBucket()
```
applike-ss, line 65: This is not required. The bucket is created as part of NewStore.

applike-ss (changes requested):
Please have a closer look at **all** remaining discussions and approach me if you need help understanding it.

pkg/fixtures/writer_s3.go
```diff
@@ -0,0 +1,109 @@
...
+		files = append(files, path)
+
+		return nil
+	})
+	if err != nil {
+		return err
+	}
+
+	replaceStr := ""
+	if s.settings.BasePath != "." && s.settings.BasePath != "./" {
```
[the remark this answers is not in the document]
mohamed-justdice: This is covering the edge case where we have a relative path `.` (only `.`) as the path. In this case if I replace it then the filename will be altered. Example from `finame.ext` to `filenameext`.

test/fixtures_s3_test.go
```diff
@@ -0,0 +1,143 @@
...
+
+const (
+	configName    = "test"
+	basePath      = "test_data/s3_fixtures_test_data"
+	basePathPurge = "test_data/s3_fixtures_purge_test_data"
+)
+
+type FixturesS3Suite struct {
+	suite.Suite
+	loader fixtures.FixtureLoader
```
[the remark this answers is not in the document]
mohamed-justdice: Applied.

test/fixtures_s3_test.go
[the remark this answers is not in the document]
mohamed-justdice: Comment was going to be deleted of course. Agreed, added attribute `bucketName` to `FixturesS3Suite`.

test/fixtures_s3_test.go
mohamed-justdice: I already had `SetupSuite()` implemented. `SetupTest()` was used to init `client` and `loader` since a new container is launched for each test.

test/fixtures_s3_test.go
mohamed-justdice: Discussed and applied modifications to replicate `aws s3 cp` command line behaviour.

test/fixtures_s3_test.go
mohamed-justdice: Altered.

docs/fixtures/readme.md
```diff
@@ -23,13 +23,14 @@ type FixtureSet struct {
 ```
 You can easily define multiple fixtures to different destinations in one file and enable or disable them for each FixtureWriter.
 
-* Currently there are 4 different FixtureWriterFactories implemented to load fixtures. 
+* Currently there are 5 different FixtureWriterFactories implemented to load fixtures.
```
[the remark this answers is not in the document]
mohamed-justdice: should've been updated every-time something is added. makes more sense to increase by 1 each time than recounting every-time x)

pkg/blob/store.go
mohamed-justdice: Applied

pkg/fixtures/writer_s3.go
mohamed-justdice: Already applied.

pkg/fixtures/writer_s3.go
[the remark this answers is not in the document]
applike-ss: in that case, maybe we should filepath.Abs the folder to get an absolute path by doing so?

pkg/fixtures/writer_s3.go
mohamed-justdice: what should I answer to this? x)

docs/fixtures/readme.md
```diff
@@ -23,13 +23,14 @@ type FixtureSet struct {
...
 
-* Currently there are 4 different FixtureWriterFactories implemented to load fixtures. 
+* Currently there are 5 different FixtureWriterFactories implemented to load fixtures. 
     * `DynamoDbFixtureWriterFactory`
     * `DynamoDbKvStoreFixtureWriterFactory` 
     * `MysqlOrmFixtureWriterFactory` 
     * `MysqlPlainFixtureWriterFactory`
     * `RedisFixtureWriterFactory`
     * `RedisKvStoreFixtureWriterFactory` 
+    * `S3FixtureWriterFactory`
```
applike-ss, line 33:
```suggestion
    * `BlobFixtureWriterFactory` 
```

pkg/fixtures/writer_s3.go
mohamed-justdice: Applied.

pkg/blob/client.go
```diff
@@ -12,29 +12,30 @@ import (
 	"time"
 )
 
-var s3cl = struct {
+var s3cls = struct {
```
applike-ss, line 15:
```suggestion
var s3clients = struct {
```

pkg/fixtures/writer_s3.go
mohamed-justdice: Applied.

pkg/fixtures/purger_s3.go
```diff
@@ -0,0 +1,45 @@
+package fixtures
```
applike-ss, line 1: this file should now be called purger_blob instead, shouldn't it?

pkg/fixtures/writer_s3.go
```diff
@@ -0,0 +1,108 @@
+package fixtures
```
applike-ss, line 1: this file should now be called writer_blob instead, shouldn't it?

pkg/fixtures/writer_s3.go
mohamed-justdice: Applied

pkg/fixtures/writer_s3.go
```diff
@@ -0,0 +1,108 @@
...
+)
+
+type BlobFixturesSettings struct {
+	BasePath   string
+	ConfigName string
+}
+
+type blobFixtureWriter struct {
+	logger      mon.Logger
+	batchRunner *blob.BatchRunner
```
applike-ss, line 21: should we accept an interface here instead?

test/test_configs/config.fixtures_s3.test.yml
mohamed-justdice: Applied

pkg/fixtures/writer_s3.go
```diff
@@ -0,0 +1,108 @@
...
+	purger      *blobPurger
+	settings    *BlobFixturesSettings
```
applike-ss, line 23: actually we only need the BasePath from the settings on the writer, don't we?

pkg/fixtures/writer_s3.go
```diff
@@ -0,0 +1,108 @@
...
+	return func(config cfg.Config, logger mon.Logger) (FixtureWriter, error) {
+		store := blob.NewStore(config, logger, settings.ConfigName)
+		br := blob.NewBatchRunner(config, logger)
+		purger := newBlobPurger(config, logger, settings)
+
+		return NewBlobFixtureWriterWithInterfaces(logger, store, br, settings, purger), nil
+	}
+}
+
+func NewBlobFixtureWriterWithInterfaces(logger mon.Logger, store blob.Store, batchRunner *blob.BatchRunner, settings *BlobFixturesSettings, purger *blobPurger) FixtureWriter {
```
applike-ss, line 37: should we change the parameters into the order that is mentioned in our coding guidelines? :thinking:

applike-ss: Please read the guidelines carefully again and check your complete changeset, if you adhere to it.

pkg/fixtures/writer_s3.go
applike-ss: I can not see it on my just-now fetched/pulled branch. Please check again!

test/fixtures_s3_test.go
mohamed-justdice: If you check you'll find that I am doing that manually, checking ALL files and keys to check the wanted behaviour. Since it's a test, it's better this way. Why reproduce what it's doing already ? doesn't make sense to me.

pkg/fixtures/writer_s3.go
applike-ss: also here, i do not see it in the most recent pushed code

pkg/fixtures/writer_s3.go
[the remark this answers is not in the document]
mohamed-justdice: I don't see why do that if this is already taking care of it.

pkg/fixtures/writer_s3.go
```diff
@@ -0,0 +1,108 @@
...
+			return err
+		}
+
+		fileInfo, _ := reader.Stat()
+		size := fileInfo.Size()
+		buffer := make([]byte, size)
+		_, err = reader.Read(buffer)
+		if err != nil {
+			return err
+		}
```
applike-ss, line 86: in one of the older discussions i told you to use ioutil.ReadFile, because you don't need to check 3 errors (of which you actually ignored one)

pkg/fixtures/writer_s3.go
```diff
@@ -0,0 +1,108 @@
...
+		purger:      purger,
+	}
+}
+
+func (s *blobFixtureWriter) Purge() error {
+	return s.purger.purge()
+}
+
+func (s *blobFixtureWriter) Write(*FixtureSet) error {
+	s.store.CreateBucket()
```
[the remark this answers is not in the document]
mohamed-justdice: Applied

test/fixtures_s3_test.go
```diff
@@ -0,0 +1,153 @@
...
+	suite.Suite
+	logger     mon.Logger
+	bucketName string
+	client     s3iface.S3API
+	loader     fixtures.FixtureLoader
+}
+
+func (s *FixturesS3Suite) SetupSuite() []suite.Option {
+	return []suite.Option{
+		suite.WithLogLevel("debug"),
```
applike-ss, line 34: not needed, is it?

test/fixtures_s3_test.go
```diff
@@ -0,0 +1,153 @@
...
+
+			return nil
+		}),
+	}
+}
+
+func (s *FixturesS3Suite) SetupTest() error {
+	s.loader = fixtures.NewFixtureLoader(s.Env().Config(), s.logger)
+	s.client = blob.ProvideS3Client(s.Env().Config())
+	return nil
```
applike-ss, line 48:
```suggestion

	return nil
```

test/fixtures_s3_test.go
```diff
@@ -0,0 +1,153 @@
...
+}
+
+func TestFixturesS3Suite(t *testing.T) {
+	s := new(FixturesS3Suite)
+	suite.Run(t, s)
+}
+
+func (s *FixturesS3Suite) TestS3(app suite.AppUnderTest) {
+	err := s.loader.Load(s3DisabledPurgeFixtures())
```
applike-ss, line 57: i would separate getting the fixtures from a func and passing it to another func. Makes debugging way easier.

test/fixtures_s3_test.go
```diff
@@ -0,0 +1,153 @@
...
+	s.NoError(err)
+
+	input := &s3.GetObjectInput{
+		Bucket: aws.String(s.bucketName),
+		Key:    aws.String(fmt.Sprint("nyan_cat2.gif")),
+	}
+	output, err := s.client.GetObject(input)
```
applike-ss, line 64: error check?

test/fixtures_s3_test.go
applike-ss: please also check other occurrences

test/fixtures_s3_test.go
applike-ss: please also check other occurrences

applike-ss (changes requested):

docs/fixtures/readme.md
mohamed-justdice: Already altered

pkg/blob/client.go
mohamed-justdice: Applied

pkg/fixtures/purger_s3.go
mohamed-justdice: Agreed and Applied

pkg/fixtures/writer_s3.go
mohamed-justdice: Agreed and Applied

pkg/fixtures/writer_s3.go
mohamed-justdice: That'd be `Module` interface or `RunnableCallable` but why? I say it's better this way since it makes it clearer.

pkg/fixtures/writer_s3.go
mohamed-justdice: no we also need the `ConfigName`, passing it to the new store.

pkg/fixtures/writer_s3.go
mohamed-justdice: Guideline doesn't say anything concerning constructors params sorting. that's maybe something to include.

mohamed-justdice: Applied

pkg/fixtures/writer_s3.go
mohamed-justdice: Already applied.

test/fixtures_s3_test.go
mohamed-justdice: Yeah I added that to debug. Now it's been removed.

test/fixtures_s3_test.go
mohamed-justdice: Applied

test/fixtures_s3_test.go
mohamed-justdice: Applied for all occurrences.

test/fixtures_s3_test.go
mohamed-justdice: Was checked the line after. Altered, for all occurrences.

docs/fixtures/readme.md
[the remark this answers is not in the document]
applike-ss: not wrong, but why not fix it if you already adjust that line anyway?

pkg/fixtures/writer_s3.go
applike-ss: This might be the one discussion where i am actually waiting for a reaction for @j4k4. However not a blocker, if we don't get a response. It's not a no-no, but rather an "i'm not sure how to handle this case -> raising to project owner" :wink:

pkg/fixtures/writer_s3.go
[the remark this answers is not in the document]
applike-ss: because it is a clean approach and the sdk give you the right tools to do it clean already

pkg/fixtures/writer_s3.go
applike-ss: But it makes it untestable unit test wise. Feel free to create an interface for the BatchRunner, because we would not want any kernel.Module or stream.RunnableCallback in here (though their interface may -by pure randomness- be the same).

mohamed-justdice: Applied.

docs/fixtures/readme.md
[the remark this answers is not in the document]
mohamed-justdice: I already did after seeing it here. now it's **7**.

pkg/fixtures/writer_s3.go
j4k4: Yeah, we don't have a rule for this yet..  
I tend to write it like `func (s *s3FixtureWriter) Write(_ *FixtureSet) error {` as well, but i have no hard feelings about it..  
I guess most places have the underscore, so maybe we should stick to it?

pkg/fixtures/writer_s3.go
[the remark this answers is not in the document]
mohamed-justdice: Agreed and applied.

pkg/fixtures/writer_s3.go
mohamed-justdice: you can now

pkg/fixtures/writer_s3.go
mohamed-justdice: you can now

pkg/blob/store.go
mohamed-justdice: Required to be used by the writer. Also the writer needs to create a bucket after purge (if purge enabled).

pkg/fixtures/writer_s3.go
applike-ss: agreed, thanks :-)

pkg/fixtures/writer_blob.go
```diff
@@ -0,0 +1,112 @@
...
+type BlobFixturesSettings struct {
+	BasePath   string
+	ConfigName string
+}
+
+type blobFixtureWriter struct {
+	logger      mon.Logger
+	batchRunner blob.BatchRunner
+	purger      *blobPurger
+	settings    *BlobFixturesSettings
```
applike-ss, line 24: again, we don't need the whole settings on the struct if we only get the BasePath from it, do we? Maybe you can also adjust the purger into that direction.

test/fixtures_s3_test.go
```diff
@@ -0,0 +1,155 @@
...
+	s.NoError(err)
+
+	body, err := ioutil.ReadAll(output.Body)
```
applike-ss, line 68: so now you are not checking the error of reading the body anymore? :see_no_evil:

test/fixtures_s3_test.go
applike-ss: no, it was checked for the ReadAll before your last changeset. Now the ReadAll is not checked for NoError anymore.

pkg/fixtures/purger_blob.go
```diff
@@ -0,0 +1,45 @@
...
+	"context"
+	"github.com/applike/gosoline/pkg/blob"
+	"github.com/applike/gosoline/pkg/cfg"
+	"github.com/applike/gosoline/pkg/mon"
+)
+
+type blobPurger struct {
+	logger      mon.Logger
+	batchRunner blob.BatchRunner
+	bucketName  string
```
applike-ss, line 13: scalars always go after interfaces

test/fixtures_s3_test.go
```diff
@@ -0,0 +1,155 @@
...
+const (
+	basePath      = "test_data/s3_fixtures_test_data"
+	basePathPurge = "test_data/s3_fixtures_purge_test_data"
+	configName    = "test"
+)
+
+type FixturesS3Suite struct {
+	suite.Suite
+	logger     mon.Logger
+	bucketName string
```
applike-ss, line 27: also here: scalars go after interfaces

test/fixtures_s3_test.go
```diff
@@ -0,0 +1,155 @@
...
+	body, err = ioutil.ReadAll(output.Body)
+	s.Equal(28092, len(body))
+
+	input = &s3.GetObjectInput{
+		Bucket: aws.String(s.bucketName),
+		Key:    aws.String(fmt.Sprint("nyan_cat2.gif")),
+	}
+	output, err = s.client.GetObject(input)
+
+	s.Nil(output.Body)
```
applike-ss, line 127: should we also assert that `subDir/nyan_cat1.gif` does not exist?

applike-ss (changes requested):

pkg/fixtures/writer_s3.go
mohamed-justdice: Applied there.

pkg/fixtures/writer_blob.go
mohamed-justdice: Applied.

test/fixtures_s3_test.go
mohamed-justdice: 🙈 😂 Corrected.

pkg/fixtures/purger_blob.go
mohamed-justdice: Altered.

test/fixtures_s3_test.go
mohamed-justdice: Altered

test/fixtures_s3_test.go
mohamed-justdice: I think one is enough, since if one is purged and no error occured then all were purged.

pkg/fixtures/purger_blob.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/applike/gosoline/pkg/blob"
+	"github.com/applike/gosoline/pkg/cfg"
+	"github.com/applike/gosoline/pkg/mon"
+)
+
+type blobPurger struct {
+	logger      mon.Logger
+	batchRunner blob.BatchRunner
+	store       blob.Store
+	bucketName  string
```
applike-ss, line 14: i just realized, that you are not even using this var. In that case, please remove it completely.

test/fixtures_s3_test.go
```diff
@@ -0,0 +1,165 @@
...
+	body, err = ioutil.ReadAll(output.Body)
+
+	s.NoError(err)
+	s.Equal(28092, len(body))
+
+	input = &s3.GetObjectInput{
+		Bucket: aws.String(s.bucketName),
+		Key:    aws.String(fmt.Sprint("nyan_cat2.gif")),
+	}
+	output, err = s.client.GetObject(input)
```
applike-ss, line 135: please add an s.NoError

applike-ss (changes requested):

pkg/fixtures/purger_blob.go
mohamed-justdice: True. Done.

test/fixtures_s3_test.go
mohamed-justdice: Here actually an error is returned. So I can assert that it's the case with `s.Error()`, but I am already doing that by making sure `output.Body` is nil.

applike-ss (approved):

PR #455, opened 2021-03-17
mon: removed panic and tafal functions from logger

j4k4:


pkg/application/app.go
```diff
@@ -61,13 +62,22 @@ func Default(options ...Option) kernel.Kernel {
...
-	config := cfg.New()
-	logger := mon.NewLogger()
+	var err error
+	var config = cfg.New()
+	var logger = mon.NewLogger()
+	var ker kernel.Kernel
 
-	return NewWithInterfaces(config, logger, options...)
+	if ker, err = NewWithInterfaces(config, logger, options...); err != nil {
+		defaultErrorHandler(err, "can initialize the app")
```
selm0, line 71: Does it make sense to indicate the os.Exit behaviour in the name of the handler?
I assumed for default only the error to be logged

selm0 (approved):

PR #467, opened 2021-03-29
redis: update go-redis to v8.8.0

### LINKED ISSUE #464: Raise go-redis dependency version
We from time to time see errors from strconv.Atoi which come from go-redis.

These are already fixed since 14.08.2020, so we should upgrade to a newer version of this dependency.

The original issue is that the connection times out and that the answer is not transferred 100% (which was also not checked before the fix) so that the library tries to parse in incomplete response.

We should also add a retryable error for that case.

The fix is here: https://github.com/go-redis/redis/pull/1129

The issue is here: https://github.com/go-redis/redis/issues/988

applike-ss:
closes #464


j4k4 (commented):
How much work would it be to add the context to our interface as well?

applike-ss:
> How much work would it be to add the context to our interface as well?

i would definitely want to do that, however i wanted to have a transition phase between "we plan to break the public interface" and "we broke the public interface". I didn't check how much i'd need to adjust to properly implement the context everywhere. Will check now.

applike-ss:
I assume that it will take roughly 1-2 hours to change the interface and all gosoline parts that relate to it. This means that in the fixture loader and in the integration tests health checks we still stick to context.Background, because no kernel was started yet.

j4k4:
The dependent projects should be updated as well.  
Not sure how much redis we're using. KvStore luckily has the context already.  
I would like to add the context, but then all at once.

applike-ss:
I've added the context, now looking into the depending projects

applike-ss:
Only one depending project was affected and i raised an internal merge request to @selm0

pkg/fixtures/purger_redis.go
```diff
@@ -28,7 +29,7 @@ func newRedisPurger(config cfg.Config, logger mon.Logger, name *string) (*redisP
 
 func (p *redisPurger) purge() error {
 	p.logger.Infof("flushing redis %s", *p.name)
-	_, err := p.client.FlushDB()
+	_, err := p.client.FlushDB(context.Background())
```
selm0, line 32: Does it make sense to pass the application context also to the fixture loaders / purgers?

selm0, line 32: Does it make sense to pass the application context also to the fixture loaders / purgers?

selm0 (changes requested):

pkg/redis/dialer.go
```diff
@@ -17,15 +18,15 @@ var dialers = map[string]Dialer{
...
+type Dialer func(logger mon.Logger, settings *Settings) func(context.Context, string, string) (net.Conn, error)
 type SrvNamingFactory func(appId cfg.AppId, name string) string
 
 var srvNamingStrategy = func(appId cfg.AppId, name string) string {
 	return fmt.Sprintf("%s.%s.redis.%s.%s", name, appId.Application, appId.Environment, appId.Family)
 }
 
-func dialerSrv(logger mon.Logger, settings *Settings) func() (net.Conn, error) {
-	return func() (net.Conn, error) {
+func dialerSrv(logger mon.Logger, settings *Settings) func(ctx context.Context, network, addr string) (net.Conn, error) {
```
j4k4, line 28: if we have a context in the dialer now, does it make sense to call `DialContext` to actually use the context?

pkg/fixtures/purger_redis.go
applike-ss: If you look closer than you see that this is executed before there is an actual kernel => no context available

pkg/redis/dialer.go
applike-ss: can do

pkg/fixtures/purger_redis.go
selm0: That's not visible by only looking at your change set.

selm0 (approved):

PR #471, opened 2021-03-30
ddb: change GetItem and BatchGetItem to not return expired items;

ajscholl:
DynamoDB features a TTL field on items, but still returns expired items
from the database if you query for them before they are deleted (the
deletion will happen up to two days after an item expires). Thus we
already had logic in gosoline to filter items from a Query or Scan which
have expired but are still returned by DDB. This commit extends this
behavior such that we also filter items from GetItem or BatchGetItems if
they did expire.

This commit also updates the ddb integration test to use the new test
framework and actually adds some tests.

### DIFF
--- a/pkg/ddb/builder_query.go
+++ b/pkg/ddb/builder_query.go
@@ -2,6 +2,7 @@ package ddb
 
 import (
 	"fmt"
+	"github.com/applike/gosoline/pkg/clock"
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/dynamodb"
 	"github.com/aws/aws-sdk-go/service/dynamodb/expression"
@@ -52,7 +53,6 @@ type QueryBuilder interface {
 type queryBuilder struct {
 	filterBuilder
 
-	metadata  *Metadata
 	indexName *string
 	selected  FieldAware
 	err       error
@@ -66,9 +66,10 @@ type queryBuilder struct {
 	consistentRead   *bool
 }
 
-func NewQueryBuilder(metadata *Metadata) QueryBuilder {
+func NewQueryBuilder(metadata *Metadata, clock clock.Clock) QueryBuilder {
 	return &queryBuilder{
-		metadata: metadata,
+		filterBuilder: newFilterBuilder(metadata, clock),
+
 		selected: metadata.Main,
 	}
 }
@@ -256,7 +257,7 @@ func (b *queryBuilder) Build(result interface{}) (*QueryOperation, error) {
 
 	exprBuilder = exprBuilder.WithKeyCondition(keyCondition)
 
-	if filter := b.buildFilterCondition(b.metadata); filter != nil {
+	if filter := b.buildFilterCondition(); filter != nil {
 		exprBuilder = exprBuilder.WithFilter(*filter)
 	}
 
--- a/pkg/ddb/builder_scan.go
+++ b/pkg/ddb/builder_scan.go
@@ -2,6 +2,7 @@ package ddb
 
 import (
 	"fmt"
+	"github.com/applike/gosoline/pkg/clock"
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/dynamodb"
 	"github.com/aws/aws-sdk-go/service/dynamodb/expression"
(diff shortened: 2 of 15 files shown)


pkg/ddb/builder_query.go
```diff
@@ -70,6 +72,7 @@ func NewQueryBuilder(metadata *Metadata) QueryBuilder {
 	return &queryBuilder{
 		metadata: metadata,
 		selected: metadata.Main,
+		clock:    clock.Provider,
```
j4k4, line 75: Does it make sense to have the clock as a parameter in `NewQueryBuilder`?  
Basically deciding what clock to use on repository creation?

pkg/ddb/builder_scan.go
```diff
@@ -47,6 +49,7 @@ func NewScanBuilder(metadata *Metadata) ScanBuilder {
 	return &scanBuilder{
 		metadata: metadata,
 		selected: metadata.Main,
+		clock:    clock.Provider,
```
j4k4, line 52: Same here with the clock.

pkg/ddb/expr_filter.go
```diff
@@ -28,3 +30,32 @@ func (b *filterBuilder) buildFilterCondition(metadata *Metadata) *expression.Con
 
 	return &expr
 }
+
+type ttlStruct struct {
```
j4k4, line 34: Can we put these two type definitions on top of the file?  
I always find it confusing, if there is something between functions for another receiver..

pkg/ddb/repository.go
```diff
@@ -185,6 +188,25 @@ func (r *repository) BatchGetItems(ctx context.Context, qb BatchGetItemsBuilder,
 
 		out := outI.(*dynamodb.BatchGetItemOutput)
 		responses := out.Responses[r.metadata.TableName]
+
```
j4k4, line 191: The `BatchGetItems` is growing a little bit too large.  
Can we extract some stuff into separate functions? Eg the new ttl filter stuff?

pkg/ddb/expr_filter.go
```diff
@@ -28,3 +30,32 @@ func (b *filterBuilder) buildFilterCondition(metadata *Metadata) *expression.Con
...
+	Ttl int64 `json:"ttl"`
+}
+
+type ttlFilterer interface {
+	PerformFilterCondition(metadata *Metadata, item map[string]*dynamodb.AttributeValue, clock clock.Clock) (bool, error)
+}
+
+func (b *filterBuilder) PerformFilterCondition(metadata *Metadata, item map[string]*dynamodb.AttributeValue, clock clock.Clock) (bool, error) {
```
j4k4, line 42: Wouldn't it be nicer, leaner and cleaner if you would take metadata and clock from the builder instead of providing it as parameters?

j4k4 (changes requested):

pkg/ddb/repository.go
```diff
@@ -183,24 +187,58 @@ func (r *repository) BatchGetItems(ctx context.Context, qb BatchGetItemsBuilder,
...
+	result.ConsumedCapacity.addSlice(out.ConsumedCapacity)
+
+	if _, ok := out.UnprocessedKeys[r.metadata.TableName]; !ok {
+		return nil, nil
+	}
+
+	return out.UnprocessedKeys, nil
+}
+
+func (r *repository) filterResponses(qb BatchGetItemsBuilder, responses []map[string]*dynamodb.AttributeValue) ([]map[string]*dynamodb.AttributeValue, error) {
```
j4k4, line 222: less nesting? :thinking: 

```
func (r *repository) filterResponses(qb BatchGetItemsBuilder, responses []map[string]*dynamodb.AttributeValue) ([]map[string]*dynamodb.AttributeValue, error) {
	var ok bool
	var filterer ttlFilterer

	if filterer, ok = qb.(ttlFilterer); !ok {
		return responses, nil
	}

	filteredResponses := make([]map[string]*dynamodb.AttributeValue, 0, len(responses))

	for _, response := range responses {
		keep, err := filterer.PerformFilterCondition(response)

		if err != nil {
			return nil, fmt.Errorf("could not perform filter condition for table %s: %w", r.metadata.TableName, err)
		}

		if keep {
			filteredResponses = append(filteredResponses, response)
		}
	}

	return filteredResponses, nil
}
```

j4k4 (approved):

PR #487, opened 2021-04-26
conc: remove time component from unit test and make it use hardcoded times;

ajscholl:


pkg/conc/ddb_lock_provider_test.go
```diff
@@ -23,15 +23,34 @@ type ddbLockProviderTestSuite struct {
...
+	clock      clock.Clock
+	backOff    backoff.BackOff
 	uuidSource *uuidMocks.Uuid
 	provider   conc.DistributedLockProvider
 
 	resource string
 	token    string
 }
 
+type testBackOff struct {
```
applike-ss, line 35: please add a factory to create one

pkg/conc/ddb_lock_provider_test.go
```diff
@@ -41,16 +60,15 @@ func (s *ddbLockProviderTestSuite) SetupTest() {
...
-		RandomizationFactor: 0,
-		Multiplier:          2,
-		MaxInterval:         time.Millisecond * 4,
-		MaxElapsedTime:      time.Millisecond * 8,
-		Clock:               clockwork.NewRealClock(), // use a real clock, otherwise time does not advance without us
+	s.backOff = &testBackOff{
+		backOffs: []time.Duration{
+			time.Millisecond * 1,
+			time.Millisecond * 2,
+			time.Millisecond * 4,
```
applike-ss, line 70: any reason these should be different to one another or can we just add time.Millisecond without a multiplicator?

pkg/conc/ddb_lock_provider_test.go
ajscholl: Do I really need to add a proper factory for some one-off code in a unit test? If I make this a proper module and everything, I would need to put it somewhere other code can reuse it - do you have a place for that? If we don't want to do that, I would just keep it as it is (I am also not adding factory functions to test cases I write even though they are often also represented as a struct).

pkg/conc/ddb_lock_provider_test.go
ajscholl: Because that is what the real code would wait for and I wanted a test to resemble the real situation as closely as possible (although the real value is most likely irrelevant anyway). I can change it if you want.

applike-ss (approved):

PR #494, opened 2021-05-11
tf: merge tags into dynamodb ressource;

bogdanfinn:


deployments/terraform/modules/dynamodb/variables.tf
```diff
@@ -159,6 +159,14 @@ variable "range_key_type" {
   description = "Range Key type, which must be a scalar type: `S`, `N`, or `B` for (S)tring, (N)umber or (B)inary data"
 }
 
+variable "tags" {
+  type        = object({
+    Group = string
+  })
```
applike-ss, line 165: i think you are specifying a strict structured object here which would not be what we would want to have imho. i would like to see a object with dynamic members (map?).

applike-ss (commented):
Did you take the redis into account as well?

applike-ss (approved):

PR #496, opened 2021-05-19
db-repo: expose and config orm db settings;

mohamed-justdice:
This makes the db connection used by the ORM configurable by an explicit setting.

### DIFF
--- a/pkg/db-repo/orm.go
+++ b/pkg/db-repo/orm.go
@@ -19,14 +19,21 @@ type OrmSettings struct {
 	Application string              `cfg:"application" default:"{app_name}"`
 }
 
+type OrmConfig struct {
+	DbName string `cfg:"db" default:"default"`
+}
+
 func NewOrm(config cfg.Config, logger mon.Logger) (*gorm.DB, error) {
-	dbClient, err := db.NewClient(config, logger, "default")
+	dbName := OrmConfig{}
+	config.UnmarshalKey("orm", &dbName)
+
+	dbClient, err := db.NewClient(config, logger, dbName.DbName)
 	if err != nil {
 		return nil, fmt.Errorf("can not create dbClient: %w", err)
 	}
 
 	settings := OrmSettings{}
-	config.UnmarshalKey("db.default", &settings)
+	config.UnmarshalKey(fmt.Sprintf("db.%s", dbName.DbName), &settings)
 
 	application := settings.Application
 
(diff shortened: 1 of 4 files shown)


pkg/db-repo/orm.go
```diff
@@ -19,14 +19,21 @@ type OrmSettings struct {
 	Application string              `cfg:"application" default:"{app_name}"`
 }
 
+type OrmConfig struct {
```
j4k4, line 22: Usually we use the `*Settings`  suffix for these kind of structs.

pkg/db-repo/orm.go
```diff
@@ -19,14 +19,21 @@ type OrmSettings struct {
...
+	DbName string `cfg:"db" default:"default"`
+}
+
 func NewOrm(config cfg.Config, logger mon.Logger) (*gorm.DB, error) {
-	dbClient, err := db.NewClient(config, logger, "default")
+	dbName := OrmConfig{}
+	config.UnmarshalKey("orm", &dbName)
+
+	dbClient, err := db.NewClient(config, logger, dbName.DbName)
```
j4k4, line 30: What's the use case behind this?

pkg/db-repo/orm.go
```diff
@@ -19,14 +19,21 @@ type OrmSettings struct {
...
 }
 
+type OrmConfig struct {
+	DbName string `cfg:"db" default:"default"`
+}
+
 func NewOrm(config cfg.Config, logger mon.Logger) (*gorm.DB, error) {
-	dbClient, err := db.NewClient(config, logger, "default")
+	dbName := OrmConfig{}
+	config.UnmarshalKey("orm", &dbName)
```
j4k4, line 28: I would like a config key that is more related to the package name, like `db_repo`

pkg/db-repo/orm.go
mohamed-justdice: tThanks for the review, very much appreciated. Yep I know, but `OrmSettings` already exists, this is a higher level to configure which db is used explicitly.

pkg/db-repo/orm.go
mohamed-justdice: This is the use case behind this:
https://gitlab.applike-services.info/mcoins/backend/marketing/-/issues/1901

pkg/db-repo/orm.go
mohamed-justdice:
```
orm:
  db_repo: default

db:
  default:
    ...
```

Is this what you mean?

PR #499, opened 2021-05-27
stream, uuid: added support for protocol buffers;

ajscholl:
This commit extends the formats supported by the stream package to
protocol buffers. A properly implemented type is then asked to provide a
protobuf version of itself, that is used to parse the message, and then
the type has to transform its protobuf version to itself. For sending
messages a similar story plays out - a type converts itself to a
protobuf version and then the message is serialized.

After implementing the proper interface with your type you should only
need to configure the output correctly to use the protobuf encoding.
Inputs already receive the format from the attributes and therefore
don't need to be told by hand which format to use.


ajscholl:
TODO: The change from `string` to `Compression/EncodingType` did not work well with the config unmarshalling... need to fix that on Monday

j4k4 (approved):
+1

PR #501, opened 2021-05-28
apiserver, http: added first option to enable compression to apiserver;

ajscholl:
Imagine you want to serve your requests and responses with compression
if the client supports it. You might even want to support the client
sending you compressed requests. How do you do this? Instead of doing
`apiserver.New(api.DefineRouter)`, you do `apiserver.New(api.DefineRouter,
apiserver.WithCompression(apiserver.DefaultCompression))`. Now all
requests which can be easily compressed (i.e., no .png, .git, .jpg, .jpeg)
get compressed without you doing anything differently.

When to use it:
- You have a simple JSON api
- Your clients properly support compression or at least don't advertise
  they do and then fail if you provide compressed responses
- You want to support getting compressed requests (you need to teach
  that to your clients out of band)


pkg/apiserver/server.go
```diff
@@ -32,14 +37,23 @@ type ApiServer struct {
...
+func New(definer Definer, options ...Option) kernel.ModuleFactory {
 	return func(ctx context.Context, config cfg.Config, logger mon.Logger) (kernel.Module, error) {
-		settings := &Settings{
-			Port:         config.GetString("api_port"),
-			Mode:         config.GetString("api_mode"),
-			TimeoutRead:  config.GetDuration("api_timeout_read"),
-			TimeoutWrite: config.GetDuration("api_timeout_write"),
-			TimeoutIdle:  config.GetDuration("api_timeout_idle"),
+		var settings *Settings
+		if config.IsSet("api_port") {
```
j4k4, line 43: I really like the idea of modernizing the config structure, but do we want to invest the time to enforce the new structure?  
I have the feeling that we won't get rid of the old stuff otherwise.. 
To even push this one step further, i would like to rename the package from `apiserver` to `httpserver`.. What do you think?

pkg/apiserver/options.go
```diff
@@ -0,0 +1,66 @@
...
+type CompressionLevel int
+
+const (
+	BestCompression    CompressionLevel = ginGzip.BestCompression
+	BestSpeed          CompressionLevel = ginGzip.BestSpeed
+	DefaultCompression CompressionLevel = ginGzip.DefaultCompression
+	NoCompression      CompressionLevel = ginGzip.NoCompression
+)
+
+func WithCompression(level CompressionLevel, gzipOpts ...ginGzip.Option) Option {
```
j4k4, line 25: Do we want to have this as an option we set via code or as a config setting?  
As we have all other settings like port etc from the config, wouldn't it more consistent to configure this via config, too?  
Or am I missing something here?

ajscholl: It would be more consistent. I just have to transform all the config options then from the config to `ginGzip.Option`s, but yeah, that would be much better. And that shouldn't be too hard... Will do that tomorrow

ajscholl: Actually did this now

pkg/apiserver/server.go
ajscholl: I think I would release this, giving the option to switch to the new structure. Then, in a later version I would drop the old structure and replace the check with a failure (i.e., if you still use the old format, the apiserver doesn't start so you see it isn't working anymore instead of silently have your settings ignored).
I really don't have a string opinion about the naming, I think `apiserver` is okay and sounds like quite some changes needed... So I wouldn't change it, but it wouldn't be too hard to actually change it and if you think this makes sense, I also wouldn't oppose that

j4k4: yeah, this is how you would do it with a library which is used a lot by external projects.. give them some time to migrate.. in our case i still like the "move fast" approach a lot and just do the hard cut.. than it's done and of the table..  
But i also don't want to force you to do it.. :grimacing:

j4k4 (approved):

PR #506, opened 2021-06-04
currency: EnsureHistoricalExchangeRates;

techmo-io:
-kvStore: ProvideConfigurableKvStore;
-currency: EnsureHistoricalExchangeRates;


pkg/currency/constants.go
```diff
@@ -0,0 +1,3 @@
+package currency
+
+var kvStoreName = "currency"
```
ajscholl, line 3: If you don't plan to change that name, you could make it a `const`.

pkg/kvstore/configurable.go
```diff
@@ -103,3 +104,31 @@ func newKvStoreChainFromConfig(config cfg.Config, logger mon.Logger, name string
 func GetConfigurableKey(name string) string {
 	return fmt.Sprintf("kvstore.%s", name)
 }
+
+var configurableKvStoreLock = sync.Mutex{}
+var ConfigurableKvStores = map[string]KvStore{}
```
ajscholl, line 109: Why is this map public if you protect it by a mutex? We wouldn't want other people do something with that map without being able to lock it, right? So it would be better if it wouldn't be exported, right?

pkg/kvstore/redis.go
```diff
@@ -134,25 +134,32 @@ func (s *redisKvStore) getChunk(ctx context.Context, resultMap *refl.Map, keys [
...
 	if err != nil {
-		return fmt.Errorf("can not get key to write value to redis: %w", err)
+		return fmt.Errorf("can not set value in redis store: %w", err)
 	}
 
-	err = s.client.Set(ctx, keyStr, bytes, s.settings.Ttl)
+	return nil
+}
 
+func (s *redisKvStore) MarshalKeyValue(key interface{}, value interface{}) (string, []byte, error) {
```
ajscholl, line 151: I think this also should be private, right?

pkg/kvstore/redis.go
```diff
@@ -162,13 +169,50 @@ func (s *redisKvStore) PutBatch(ctx context.Context, values interface{}) error {
...
+	chunkSize := s.settings.BatchSize
+	pairs := make([]interface{}, 0, chunkSize)
 	for k, v := range mii {
-		if err = s.Put(ctx, k, v); err != nil {
-			return fmt.Errorf("failed to write batch to redis: %w", err)
+		key, value, err := s.MarshalKeyValue(k, v)
+		if err != nil {
+			return fmt.Errorf("PutBatch could not marshal key/value: %w", err)
+		}
+		pairs = append(pairs, key)
```
ajscholl, line 179: I think this can be written as `pairs = append(pairs, key, value)`, making it a little bit more compact

pkg/kvstore/redis.go
```diff
@@ -162,13 +169,50 @@ func (s *redisKvStore) PutBatch(ctx context.Context, values interface{}) error {
 		return fmt.Errorf("could not convert values from %T to map[interface{}]interface{}", values)
 	}
 
+	chunkSize := s.settings.BatchSize
+	pairs := make([]interface{}, 0, chunkSize)
```
ajscholl, line 173: It is nice that you try to already reserve enough capacity, but then you should reserve twice that much because you add keys and values.

pkg/kvstore/redis.go
```diff
@@ -162,13 +169,50 @@ func (s *redisKvStore) PutBatch(ctx context.Context, values interface{}) error {
...
+
+	err := s.client.MSet(ctx, pairs)
+	if err != nil {
+		return fmt.Errorf("failed to flush batch to redis: %w", err)
+	}
+
+	// setting ttl
+	pipe := s.client.Pipeline()
+	for i := 0; i < len(pairs); i += 2 {
+		keyStr, ok := pairs[i].(string)
```
ajscholl, line 207: This is code controlled completely by you, so this should never fail. Therefore I would not check the assertion, but simply panic if it is not the case.

pkg/kvstore/redis.go
```diff
@@ -162,13 +169,50 @@ func (s *redisKvStore) PutBatch(ctx context.Context, values interface{}) error {
...
-	return nil
+	return s.flushChunk(ctx, pairs)
+}
+
+func (s *redisKvStore) flushChunk(ctx context.Context, pairs []interface{}) error {
+	if len(pairs) < 1 {
+		return nil
+	}
+
+	err := s.client.MSet(ctx, pairs)
```
ajscholl, line 199: Hmm, can we also do the mset in the pipeline? Is Pipeline the go equivalent of `MULTI` in redis? Because otherwise we have a strange state visible where the data was set, but has no ttl set.

pkg/redis/client.go
```diff
@@ -18,6 +18,10 @@ type ErrCmder interface {
 	Err() error
 }
 
+type Pipeliner interface {
```
ajscholl, line 21: If you generate a mock for this interface, please also add the go:generate comment to it

ajscholl (changes requested):

pkg/kvstore/redis.go
techmo-io: umm, pipelining by itself, is just saving on network roundtrip, so no transaction behaviour,
but we can easily make the whole thing a transaction.
on it.

pkg/kvstore/redis.go
j4k4: actually, i would prefer this assertion check instead of the panic..  
feels just more clean.. you're right, that this can fail only by changing some code, but if some other developer changes something somewhere, he would get a nice hint about what went wrong? :thinking:

techmo-io: reverted to the assertion check.

pkg/currency/constants.go
techmo-io: done

pkg/kvstore/configurable.go
techmo-io: done

pkg/kvstore/redis.go
techmo-io: done

pkg/kvstore/redis.go
techmo-io: done

pkg/kvstore/redis.go
techmo-io: done

pkg/redis/client.go
techmo-io: done

pkg/kvstore/redis.go
```diff
@@ -162,13 +169,45 @@ func (s *redisKvStore) PutBatch(ctx context.Context, values interface{}) error {
...
+	pipe := s.client.Pipeline().TxPipeline()
+
+	pipe.MSet(ctx, pairs)
+	// setting ttl
+	for i := 0; i < len(pairs); i += 2 {
+		keyStr, ok := pairs[i].(string)
+		if !ok {
+			return fmt.Errorf("setting ttl, failed to cast key to string: %v", pairs[i])
+		}
+		pipe.Expire(ctx, keyStr, s.settings.Ttl)
```
j4k4, line 207: Do we have a ttl in every case? Or is the ttl optional?  
So we need a call for every single pair? I guess no way to do this in batch as well?

pkg/kvstore/redis.go
ajscholl: I would argue the hint would be the failing test the dev should then get. And at least for me it is much easier to debug a panic than an error. Why? Because I can set a breakpoint for a panic (I think it is somewhere in the runtime in fatalpanic or something like that for me) and then immediately get the state at the panic - with an error, I always have to trace that. But if you would like to keep the error, that is also fine with me, in the end it should be covered by the test anyway.

pkg/kvstore/redis_test.go
```diff
@@ -90,8 +91,20 @@ func TestRedisKvStore_Put(t *testing.T) {
 
 func TestRedisKvStore_PutBatch(t *testing.T) {
 	store, client := buildTestableRedisStore()
-	client.On("Set", mock.AnythingOfType("*context.emptyCtx"), "applike-gosoline-kvstore-kvstore-test-foo", []byte(`{"id":"foo","body":"bar"}`), time.Duration(0)).Return(nil)
-	client.On("Set", mock.AnythingOfType("*context.emptyCtx"), "applike-gosoline-kvstore-kvstore-test-fuu", []byte(`{"id":"fuu","body":"baz"}`), time.Duration(0)).Return(nil)
+
+	pipe := &redisMocks.Pipeliner{}
+	pipe.On("MSet", mock.AnythingOfType("*context.emptyCtx"), mock.MatchedBy(func(input []interface{}) bool {
+		possibleInput1 := `[applike-gosoline-kvstore-kvstore-test-foo {"id":"foo","body":"bar"} applike-gosoline-kvstore-kvstore-test-fuu {"id":"fuu","body":"baz"}]`
+		possibleInput2 := `[applike-gosoline-kvstore-kvstore-test-fuu {"id":"fuu","body":"baz"} applike-gosoline-kvstore-kvstore-test-foo {"id":"foo","body":"bar"}`
```
ajscholl, line 98: Is this missing a `]` at the end?

techmo-io: it was.

pkg/kvstore/redis.go
techmo-io: don't know of a batch TTL command,
pipelining was the next best thing.

ajscholl (approved):

PR #508, opened 2021-06-04
apiserver, test: added struct validation and other options to apiserver and dropped support for api_port, api_mode, etc.

ajscholl:
This commit exposes the struct validation next to the field validation
from validator/v10 as well as type mappers and aliases. It also switches
the api server completely to the new format as supporting both formats
was getting more complex than expected. If you are using the old format,
you will get an error when starting the api servier.

This commit also extends the ApiTestSuite to allow returning a slice of
test cases instead of a single test and adds a second validation
function which gets the response as a parameter.


j4k4 (approved):
+1

PR #514, opened 2021-06-10
stream: add options to the producer daemon for the maximum size of each message and the total batch of those;

ajscholl:
The producer daemon now gains a lot more intelligence (for good or bad) about how to batch messages. Each aggregate and batch can grow to a maximum number of bytes (and we also know about how many bytes a compressed message roughly consumes). By default we use the numbers for SNS/SQS there.


pkg/stream/output_in_memory.go
```diff
@@ -2,6 +2,7 @@ package stream
 
 import (
 	"context"
+	"encoding/json"
```
j4k4, line 5: json package again?

pkg/stream/producer_daemon_aggregator.go
```diff
@@ -0,0 +1,237 @@
+package stream
+
+import (
+	"bytes"
+	"compress/gzip"
+	"encoding/json"
```
j4k4, line 6: here again the json package?

pkg/stream/producer_daemon_aggregator_test.go
```diff
@@ -0,0 +1,352 @@
...
+type aggregatorTestCase struct {
+	aggregationSize    int
+	aggregationMaxSize int
+	compression        stream.CompressionType
+
+	messages []*stream.Message
+	flushes  []expectedFlush
+}
+
+type expectedFlush struct {
```
j4k4, line 25: can we group types and receiver functions?

j4k4 (approved):

PR #515, opened 2021-06-11
error logging with fmt.Errorf

j4k4:
Since golang 1.13 there is fmt.Errorf to create and wrap errors which allows the usual msg and args pattern to format and create errors.  
We use this functionality now to create the final error which is used now on logging any error.


pkg/cfg/config.go
```diff
@@ -177,8 +177,8 @@ func (c *config) GetMsiSlice(key string, optionalDefault ...[]map[string]interfa
 	var reflectValue = reflect.ValueOf(data)
 
 	if reflectValue.Kind() != reflect.Slice {
-		err = fmt.Errorf("can not cast value %v[%T] of key %s to []map[string]interface{}", data, data, key)
-		c.err(err, err.Error())
+		err = fmt.Errorf("can not cast value %v[%T] of key %s to []map[string]interface{}: %w", data, data, key, err)
+		c.err(err.Error())
```
ajscholl, line 181: Shouldn't this be `c.err("can not cast value %v[%T] of key %s to []map[string]interface{}: %w", data, data, key, err)`?

pkg/cfg/config.go
```diff
@@ -620,7 +619,8 @@ func (c *config) unmarshalStruct(key string, output interface{}, additionalDefau
 		settings, err := c.settings.Get(key).Map()
 
 		if err != nil {
-			c.err(fmt.Errorf("value is not of type map[string]interface{}: %w", err), "can not get settings for key: %s", key)
+			err = fmt.Errorf("value is not of type map[string]interface{}: %w", err)
```
ajscholl, line 622: Is this line correct? What is `value` in this context?

pkg/cfg/error.go
```diff
@@ -1,37 +1,37 @@
...
-	err = errors.Wrapf(err, msg, args...)
+func PanicErrorHandler(msg string, args ...interface{}) {
+	err := fmt.Errorf(msg, args...)
 	panic(err)
 }
 
 func LoggerErrorHandler(logger Logger) ErrorHandler {
-	return func(err error, msg string, args ...interface{}) {
-		logger.Error(err, msg, args...)
+	return func(msg string, args ...interface{}) {
```
ajscholl, line 27: So this whole could be replaced by `return logger.Error`?

pkg/cfg/error.go
```diff
@@ -1,37 +1,37 @@
...
+		logger.Error(msg, args...)
 	}
 }
 
 func SentryErrorHandler(sentry Sentry) ErrorHandler {
-	return func(err error, msg string, args ...interface{}) {
-		err = errors.Wrapf(err, msg, args...)
+	return func(msg string, args ...interface{}) {
+		err := fmt.Errorf(msg, args...)
```
ajscholl, line 34: I wonder... `errors.Wrapf` adds a stack trace somehow... do we need this? Does sentry add its own stack trace?

pkg/db-repo/change_history_manager.go
```diff
@@ -224,7 +224,7 @@ func (c *changeHistoryManager) execute() error {
 		if err != nil {
 			c.logger.WithFields(mon.Fields{
 				"sql": statement,
-			}).Error(err, "")
+			}).Error(": %w", err)
```
ajscholl, line 227: This looks like it could need a proper error message

pkg/db-repo/notification_repo.go
```diff
@@ -77,10 +77,7 @@ func (r *notifyingRepository) doCallback(ctx context.Context, callbackType strin
 	}
 
 	if len(errors) > 0 {
-		err := fmt.Errorf("there were %v errors during execution of the callbacks for create", len(errors))
-		logger.Error(err, err.Error())
-
-		return err
+		return fmt.Errorf("there were %v errors during execution of the callbacks for create", len(errors))
```
ajscholl, line 80: Maybe we should replace the whole error by a multierror? I for my part hate it if I know there were 7 errors, but don't know what the errors were...

pkg/dispatcher/repository_db.go
```diff
@@ -36,7 +36,7 @@ func (r Repository) Create(ctx context.Context, value db_repo.ModelBased) error
 
 	for _, err := range errs {
 		if err != nil {
-			r.logger.Error(err, "error on "+db_repo.Create+" for event "+eventName)
+			r.logger.Error("error on "+db_repo.Create+" for event : %w"+eventName, err)
```
ajscholl, line 39: This file seems to mix format strings and string concatenation

pkg/mon/mocks/factory.go
```diff
@@ -38,38 +38,32 @@ func NewLoggerMockedUntilLevel(level string) *Logger {
...
+	mockLoggerMethod(logger, "Info", mon.Info, levelIndex >= levelMap[mon.Info])
+	mockLoggerMethod(logger, "Warn", mon.Warn, levelIndex >= levelMap[mon.Warn])
+	mockLoggerMethod(logger, "Error", mon.Error, levelIndex >= levelMap[mon.Error])
 
 	return logger
 }
 
 func inspectLogFunction(level string, allowed bool) func(args mock.Arguments) {
 	return func(args mock.Arguments) {
 		// ensure we use formatting markers exactly when using a method with formatting
```
ajscholl, line 51: I think we can delete this comment, it is no longer valid

pkg/cfg/config.go
j4k4: indeed

pkg/cfg/config.go
j4k4: that wrapping can be dropped, as the `Map()` call already returns an appropriate error

pkg/cfg/error.go
j4k4: Luckily the `SentryErrorHandler` is not used anywhere, so we don't have to care about this here..  
But in general the sentry stuff needs some overhaul, as it uses quite some old error packages.  
This is planned for the incoming new sentry handler for the incoming new logger package..

pkg/db-repo/change_history_manager.go
j4k4: added the message from the error return..

pkg/db-repo/notification_repo.go
j4k4: using multierror now

pkg/dispatcher/repository_db.go
j4k4: fixed

pkg/mon/mocks/factory.go
j4k4: done

pkg/db-repo/notification_repo.go
```diff
@@ -65,22 +66,19 @@ func (r *notifyingRepository) doCallback(ctx context.Context, callbackType strin
...
 		}
 	}
 
-	if len(errors) > 0 {
-		err := fmt.Errorf("there were %v errors during execution of the callbacks for create", len(errors))
-		logger.Error(err, err.Error())
-
-		return err
+	if errors != nil {
+		return fmt.Errorf("there were errors during execution of the callbacks for create: %w", errors)
```
ajscholl, line 81: Actually they are not the callbacks for create, but for `callbackType`, right?

ajscholl (approved):

PR #525, opened 2021-06-30
db_repo: add query crossing model verification and error;

mohamed-justdice:
This adds a verification in db_repo.Query and triggers an error if the model being queried is different than the repo metadat.

### DIFF
--- a/pkg/db-repo/repository.go
+++ b/pkg/db-repo/repository.go
@@ -224,11 +224,7 @@ func (r *repository) isQueryableModel(model interface{}) bool {
 		modelType = modelType.Elem()
 	}
 
-	if strings.ToLower(modelType.Name())  != strings.ToLower(r.GetMetadata().ModelId.Name) {
-		return false
-	}
-
-	return true
+	return strings.EqualFold(modelType.Name(), r.GetMetadata().ModelId.Name)
 }
 
 func (r *repository) Query(ctx context.Context, qb *QueryBuilder, result interface{}) error {
@@ -236,7 +232,7 @@ func (r *repository) Query(ctx context.Context, qb *QueryBuilder, result interfa
 	defer span.Finish()
 
 	db := r.orm.New()
-	
+
 	db = db.Table(r.GetMetadata().TableName)
 
 	for _, j := range qb.joins {
--- a/pkg/db-repo/repository.go
+++ b/pkg/db-repo/repository.go
@@ -27,10 +27,10 @@ const (
 )
 
 var (
-	operations = []string{Create, Read, Update, Delete, Query}
-	ErrCrossQuery = fmt.Errorf("cross querying wrong model from repo")
+	operations     = []string{Create, Read, Update, Delete, Query}
+	ErrCrossQuery  = fmt.Errorf("cross querying wrong model from repo")
 	ErrCrossCreate = fmt.Errorf("cross creating wrong model from repo")
-	ErrCrossRead = fmt.Errorf("cross reading wrong model from repo")
+	ErrCrossRead   = fmt.Errorf("cross reading wrong model from repo")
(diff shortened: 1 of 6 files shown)


pkg/db-repo/repository.go
```diff
@@ -217,6 +217,16 @@ func (r *repository) Delete(ctx context.Context, value ModelBased) error {
 	return err
 }
 
+func (r *repository) isQueryableModel(model interface{}) bool {
+	modelType := reflect.TypeOf(model)
```
selm0, line 221: Solving this via reflection requires the model to always match the name on the metadata. I wouldn't rely on this assumption.

Can you find a way to e.g. compare the table names in the from clause?
Maybe the ORM provides a way to get the table name from the model

pkg/db-repo/repository.go
```diff
@@ -230,6 +240,15 @@ func (r *repository) Query(ctx context.Context, qb *QueryBuilder, result interfa
 	}
 
 	for i := range qb.where {
+		currentWhere := qb.where[i]
```
selm0, line 243: Will this break any joins with conditions applied on the joined entity, e.g.
select* from user u join device d on u.device_id = d.id where d.type = 'phone'?

Maybe it also makes sense to add the check to the other methods as well to detect any cross-entity calls?

test/db_repo_query_test.go
```diff
@@ -0,0 +1,149 @@
...
+
+	s.NoError(err)
+}
+
+func (s *DbRepoQueryTestSuite) TearDownSuite() {
+	s.mocks.Shutdown()
+}
+
+func (s *DbRepoQueryTestSuite) TestQueryCorrectModel() {
+	ctx := context.Background()
```
selm0, line 81: What's the use case of an integration test here? What do you need the running DB instance for?

selm0 (changes requested):

pkg/db-repo/repository.go
mohamed-justdice: No, it shouldn't break the Joins. since they are done separately.

pkg/db-repo/repository.go
mohamed-justdice: Applied

test/db_repo_query_test.go
mohamed-justdice: making sure that the correct query works (when using the correct model)

pkg/db-repo/repository.go
```diff
@@ -217,20 +217,33 @@ func (r *repository) Delete(ctx context.Context, value ModelBased) error {
 	return err
 }
 
+func (r *repository) isQueryableModel(db *gorm.DB, model interface{}) bool {
+	tableName := db.NewScope(model).TableName()
```
selm0, line 221: why don't you use the orm on the repository?
	tableName := r.orm.NewScope(model).TableName()

selm0 (changes requested):

pkg/db-repo/repository.go
mohamed-justdice: Applied

pkg/db-repo/repository.go
```diff
@@ -246,7 +299,14 @@ func (r *repository) Query(ctx context.Context, qb *QueryBuilder, result interfa
 		db = db.Limit(qb.page.limit)
 	}
 
-	err := db.Find(result).Error
+	err := r.checkResultModel(result)
```
selm0, line 302: I suggest moving this check to the beginning of the file, like for the other operations.
Prevents doing unnecessary stuff :D

selm0 (changes requested):

pkg/db-repo/repository.go
mohamed-justdice: Done

PR #528, opened 2021-07-08
guard: implement GetAll;

selm0:
The guard package missed the implementation for `GetAll` and `GetById` of the ladon Manager.
Both are added in this PR


pkg/guard/guard.go
```diff
@@ -26,22 +30,44 @@ func NewGuard(config cfg.Config, logger mon.Logger) (*LadonGuard, error) {
...
 	return g.warden.IsAllowed(request)
 }
 
-func (g LadonGuard) GetPolicesBySubject(subject string) (ladon.Policies, error) {
-	pol, err := g.warden.Manager.FindPoliciesForSubject(subject)
+func (g LadonGuard) GetPolicies() (ladon.Policies, error) {
+	policies := make(ladon.Policies, 0)
+
+	for offset := int64(0); true; offset += fetchLimit {
+		pols, err := g.warden.Manager.GetAll(fetchLimit, offset)
```
mohamed-justdice, line 54: A personal preference is that we shouldn't use an infinite loop here, even though `.GetAll()` will return len==0 at some point. But local checks are more efficient.

mohamed-justdice (changes requested):

pkg/guard/guard_test.go
```diff
@@ -0,0 +1,38 @@
...
+
+	pol2 := &ladon.DefaultPolicy{
+		ID: "200",
+	}
+
+	manager.On("GetAll", int64(100), int64(0)).Return(ladon.Policies{pol1}, nil).Once()
+	manager.On("GetAll", int64(100), int64(100)).Return(ladon.Policies{pol2}, nil).Once()
+	manager.On("GetAll", int64(100), int64(200)).Return(ladon.Policies{}, nil).Once()
+
+	pols, err := g.GetPolicies()
```
mohamed-justdice, line 31: `GetPoliciesBySubject` is not tested here

mohamed-justdice (approved):

PR #529, opened 2021-07-12
validation: log validation failure;

mohamed-justdice:
log the validation `validation.repository_db` err if it returns one.

### DIFF
--- a/pkg/validation/repository_db.go
+++ b/pkg/validation/repository_db.go
@@ -3,24 +3,30 @@ package validation
 import (
 	"context"
 	"github.com/applike/gosoline/pkg/db-repo"
+	"github.com/applike/gosoline/pkg/mon"
 )
 
 type Repository struct {
+	logger mon.Logger
 	db_repo.Repository
 	validator Validator
 }
 
-func NewRepository(validator Validator, repo db_repo.Repository) db_repo.Repository {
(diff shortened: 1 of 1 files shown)


selm0 (approved):
Don't forget to squash your commits

j4k4:
I'm not a fan of this. Why do we need this?  
Usually the caller should decide what should happen on error. I did this error by myself more than once..  
Even logging this as a warning will hurt you if you run into an use case where you don't want to have a warning..

selm0:
> I'm not a fan of this. Why do we need this?
> Usually the caller should decide what should happen on error. I did this error by myself more than once..
> Even logging this as a warning will hurt you if you run into an use case where you don't want to have a warning..

Do you refer to e.g. the logs in db.repository?
It seems to me that you're advocating to remove any logs from gosoline :thinking:

j4k4:
> Do you refer to e.g. the logs in db.repository?
Yes

> It seems to me that you're advocating to remove any logs from gosoline
That's not what I said

But a lot of the gosoline packages is utility code. Especially stuff like validation or repositories..  
The result of fetching or validating a record by itself has no real meaning. How should the validation code know how severe a failing validation is? Only your business layer has the knowledge about that. So the logging should happen there.

PR #530, opened 2021-07-14
apiserver: add profiling option

lantosgyuri:


pkg/apiserver/server.go
```diff
@@ -89,7 +89,7 @@ func NewWithInterfaces(logger mon.Logger, router *gin.Engine, tracer tracing.Tra
 	}
 
 	var err error
-	var address = server.Addr
+	address := server.Addr
```
selm0, line 92: please order the variables here :)

pkg/application/options.go
```diff
@@ -287,6 +289,15 @@ func WithProducerDaemon(app *App) {
 	})
 }
 
+func WithProfiling() Option {
```
selm0, line 292: And redo the commit message, it's not anymore only about the streams

selm0 (changes requested):

selm0 (approved):

PR #532, opened 2021-07-19
log: implemented new logger architecture

j4k4:
Removed the old `mon` package and did a split into a new `log` and `metric` package.  
`log` contains a new logger implementation based on the existing interface with a more flexible 'handler' architecture.  
`metric` is a copy of the metric related files of the old `mon` package with stripped 'metric' prefixes.


pkg/log/testdata/config.yml
```diff
@@ -0,0 +1,6 @@
+log:
+  level: warn
```
ajscholl, line 2: What does this level mean? I can't find a use for that

pkg/cli/logger.go
```diff
@@ -1,99 +1,21 @@
...
+	var err error
+	var writer io.Writer
 
-func (r *fingersCrossedOutput) Write(line []byte) (n int, err error) {
-	if !r.crossed {
-		return os.Stdout.Write(line)
-	}
-
-	r.buffer = append(r.buffer, line)
+	if writer, err = log.NewIoWriterFile("logs.log"); err != nil {
```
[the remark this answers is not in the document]
selm0: Isn't this behaviour also what you critized on the other PR, that you're requiring a file logger for the cli?
Shouldn't this be configurable via config file / flags / env parameters, and defaulting to stdout / stderr

pkg/log/formatter.go
```diff
@@ -0,0 +1,76 @@
...
+		return nil, fmt.Errorf("failed to marshal fields to JSON, %v", err)
+	}
+
+	return append(serialized, '\n'), nil
+}
+
+func getFieldsAsString(fields map[string]interface{}) string {
+	fieldParts := make([]string, 0, len(fields))
+
+	for k, v := range fields {
```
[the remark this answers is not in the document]
j4k4: sounds good :+1:

pkg/cli/logger.go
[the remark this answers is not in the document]
j4k4: Yes, it should be possible :+1:   
But I would like to tackle an improved `clui` package in another PR  
So this code is just to have something runnable.. :)

pkg/log/handler_iowriter.go
```diff
@@ -0,0 +1,84 @@
...
+type HandlerIoWriterSettings struct {
+	Level           string   `cfg:"level" default:"info"`
+	Channels        []string `cfg:"channels"`
+	Formatter       string   `cfg:"formatter" default:"console"`
+	TimestampFormat string   `cfg:"timestamp_format" default:"15:04:05.000"`
+	Writer          string   `cfg:"writer" default:"stdout"`
+}
+
+func handlerIoWriterFactory(config cfg.Config, handlerIndex int) (Handler, error) {
+	key := fmt.Sprintf("log.handlers[%d]", handlerIndex)
```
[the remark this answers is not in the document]
j4k4: yeah, makes sense.. changed it to having the key as argument..

pkg/log/fields.go
```diff
@@ -0,0 +1,91 @@
...
+func prepareForLog(v interface{}) interface{} {
+	switch t := v.(type) {
+	case error:
+		// Otherwise errors are ignored by `encoding/json`
+		return t.Error()
+	case time.Time:
+		return v
+	case map[string]interface{}:
+		// perform a deep copy of any maps contained in this map element
+		// to ensure we own the object completely
```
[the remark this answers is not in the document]
selm0: I'm not sure if I get your point right; but why would you want to have error fields inside a nested map?
Isn't a top-level (multi) error field enough, and anything nested might confuse you?

pkg/log/fields.go
```diff
@@ -0,0 +1,91 @@
...
+	for k, v := range input {
+		newMap[k] = prepareForLog(v)
+	}
+
+	return newMap
+}
+
+func prepareForLog(v interface{}) interface{} {
+	switch t := v.(type) {
+	case error:
```
selm0, line 25: Does this also work with the multi-error package?

pkg/log/testdata/config.yml
j4k4: It's used in: 

`config.UnmarshalKey(handlerConfigKey, settings, cfg.UnmarshalWithDefaultsFromKey("log.level", "level"))`  
`gosoline/pkg/log/handler_iowriter.go:21`  

Enables one to configure the default level for all handlers in a single place.

pkg/log/formatter.go
```diff
@@ -0,0 +1,76 @@
...
+var formatters = map[string]Formatter{
+	"console": FormatterConsole,
+	"json":    FormatterJson,
+}
+
+func FormatterConsole(timestamp string, level int, format string, args []interface{}, err error, data Data) ([]byte, error) {
+	fieldString := getFieldsAsString(data.Fields)
+	contextString := getFieldsAsString(data.ContextFields)
+
+	levelStr := fmt.Sprintf("%-7v", LevelName(level))
```
selm0, line 21: Why do you need to use `%v` here instead of `%d`?

pkg/log/handler.go
```diff
@@ -0,0 +1,45 @@
...
+type Handler interface {
+	Channels() []string
+	Level() int
+	Log(timestamp time.Time, level int, msg string, args []interface{}, err error, data Data) error
+}
+
+type HandlerFactory func(config cfg.Config, handlerIndex int) (Handler, error)
+
+var handlerFactories = map[string]HandlerFactory{
+	"iowriter": handlerIoWriterFactory,
```
selm0, line 18: What other handler factories do you imagine to be implemented? Having the factory pattern for only one type looks a bit boilerplate-ish

pkg/log/handler.go
```diff
@@ -0,0 +1,45 @@
...
+}
+
+func AddHandlerFactory(name string, factory HandlerFactory) {
+	handlerFactories[name] = factory
```
selm0, line 22: Also, why isn't the handlerIoWriterFactory using the `AddHandlerFactory` method?

pkg/log/handler.go
```diff
@@ -0,0 +1,45 @@
...
+	var handlerFactory HandlerFactory
+	var handlers = make([]Handler, len(settings.Handlers))
+
+	for i, handlerSettings := range settings.Handlers {
+		if handlerFactory, ok = handlerFactories[handlerSettings.Type]; !ok {
+			return nil, fmt.Errorf("there is no logging handler of type %s", handlerSettings.Type)
+		}
+
+		if handlers[i], err = handlerFactory(config, i); err != nil {
+			return nil, fmt.Errorf("can not create logging handler of type %s on index %d", handlerSettings.Type, i)
```
selm0, line 40: Can you please include the err returned by the handler factory? 
Or is it already logged in a different place?

selm0: When including the error, you should also be able to drop the index from the error string as I expect the error to be handler specific enough to identify where it originated (or using the stack trace)

pkg/log/handler_iowriter.go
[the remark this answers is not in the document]
selm0: I would also support a different mechanism, e.g. a named config as we do have for other repositories? Or won't that fit in the use cases?

pkg/log/handler.go
```diff
@@ -0,0 +1,45 @@
...
+	settings := &LoggerSettings{}
+	config.UnmarshalKey("log", settings)
+
+	var ok bool
+	var err error
+	var handlerFactory HandlerFactory
+	var handlers = make([]Handler, len(settings.Handlers))
+
+	for i, handlerSettings := range settings.Handlers {
+		if handlerFactory, ok = handlerFactories[handlerSettings.Type]; !ok {
```
selm0, line 35: Personally, I favour not to define the variables separately:
```
handlerFactory, ok := handlerFactories...
if !ok {
}```

Only saves a few lines, I know :D

File: pkg/log/handler_sentry.go [hunk 12]
```diff
@@ -0,0 +1,79 @@
...
+	"time"
+)
+
+//go:generate mockery -name Sentry
+type Sentry interface {
+	CaptureException(exception error, hint *sentry.EventHint, scope sentry.EventModifier) *sentry.EventID
+}
+
+func init() {
+	AddHandlerFactory("metric", LoggerHandlerFactory)
```
selm0, line 15: Is `metric` the correct key here?

pkg/log/handler_sentry.go
```diff
@@ -0,0 +1,79 @@
...
+
+func (h HandlerSentry) Channels() []string {
+	return []string{}
+}
+
+func (h HandlerSentry) Level() int {
+	return PriorityError
+}
+
+func (h HandlerSentry) Log(_ time.Time, _ int, _ string, _ []interface{}, err error, data Data) error {
```
selm0, line 64: Those are a lot of unused parameters; does it make sense to split the use cases of logging errors vs. logging the rest on handler level?

pkg/log/logger.go
```diff
@@ -0,0 +1,222 @@
...
+		}
+	}
+}
+
+func (l *gosoLogger) shouldLogToChannels(current string, channels []string) bool {
+	if len(channels) == 0 {
+		return true
+	}
+
+	for _, channel := range channels {
```
selm0, line 199: Just an idea: would it be better to read and understand the different cases when using `funk.ContainsString()` with either `*`, `channel` or `!channel`?

pkg/mapx/mapnode.go
```diff
@@ -31,6 +31,13 @@ func interfaceToMapNode(val interface{}) *MapXNode {
 
 		return &MapXNode{value: slice}
 
+	case []interface{}:
+		for i, v := range val {
+			val[i] = interfaceToMapNode(v).value
```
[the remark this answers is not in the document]
selm0: +1, you shouldn't modify arguments

selm0 (changes requested):

pkg/log/testdata/config.yml
ajscholl: I see. Yeah, that makes sense, just looked a little bit confusing to have two different levels there without a direct explanation.

pkg/log/fields.go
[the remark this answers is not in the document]
ajscholl: In general yes, but if someone would do `map[string]interface{}{ "myError": err }`, that would be excluded from the JSON serializer, right? But this method here nicely converts them to strings by calling `Error`, so that should no longer a problem. Just thought we could also add that to the comment

pkg/log/formatter.go
ajscholl: Do you mean `%s`? I think `LevelName(level)` returns a string, not a number. But yeah, I guess that could be done like for the channel in the next line

pkg/mapx/map.go
```diff
@@ -88,12 +88,16 @@ func (m *MapX) Set(key string, value interface{}, options ...MapOption) {
...
 	case Msier:
-		msi := msier.Msi()
+		msi := t.Msi()
 		value = msiToMsn(msi)
 	case map[string]interface{}:
-		value = msiToMsn(msier)
+		value = msiToMsn(t)
+	case []interface{}:
+		for i, val := range t {
+			t[i] = interfaceToMapNode(val).value
```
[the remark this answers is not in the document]
j4k4: fixed

pkg/mapx/mapnode.go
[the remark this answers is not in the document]
j4k4: fixed, too

pkg/log/formatter.go
j4k4: switched to `%s`

pkg/log/handler_sentry.go
j4k4: changed to `sentry`

pkg/log/handler.go
j4k4: changed

pkg/log/handler.go
j4k4: well, there are already the sentry and metric handlers..  
currently I have no plan for additional handlers but with this pattern it is open for extension..

pkg/log/handler.go
j4k4: Personally, I like defining the variables separately cause I find it more readable :)

pkg/log/logger.go
j4k4: would be an alternative, yes

pkg/log/handler_sentry.go
j4k4: I don't think so.. yes, the handler don't uses the parameter but i could imagine a handler which actually uses all of these..  
Isn't the io writer handler such a case? It uses all arguments on error..

pkg/log/handler.go
j4k4: added the error but kept the index.. just an additional info that shouldn't hurt..

pkg/log/fields.go
j4k4: As the multi-error implements the `error` interface it should work, yes..  
And this code is just copied from the old `mon` package, so it is somewhat battle tested already.. :grimacing:

pkg/log/fields.go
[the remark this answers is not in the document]
j4k4: added a note to the comment

pkg/mapx/struct.go
```diff
@@ -433,7 +511,7 @@ func (m *MapXStruct) doWriteMap(cfg string, targetValue reflect.Value, sourceMap
 	return nil
 }
 
-func (m *MapXStruct) doWriteSlice(cfg string, targetValue reflect.Value, sourceValues *MapX) error {
+func (m *Struct) doWriteSlice(cfg string, targetValue reflect.Value, sourceValues *MapX) error {
```
selm0, line 514: Maybe you want to switch the receiver name to `s` instead of `m`

j4k4: done

pkg/application/options.go
```diff
@@ -158,122 +152,89 @@ func WithKernelSettingsFromConfig(app *App) {
...
+		if handlers, err = log.NewHandlersFromConfig(config); err != nil {
+			return fmt.Errorf("can not create handlers from config: %w", err)
+		}
 
-func WithLoggerMetricHook(app *App) {
-	app.addLoggerOption(func(config cfg.GosoConf, logger mon.GosoLog) error {
-		metricHook := mon.NewMetricHook()
-		return logger.Option(mon.WithHook(metricHook))
+		if len(handlers) == 0 {
+			handlers = append(handlers, log.NewCliHandler())
```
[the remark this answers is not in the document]
j4k4: It's the old and new default logger behavior, yes

pkg/cfg/environment.go
```diff
@@ -0,0 +1,62 @@
+package cfg
+
+import (
+	"os"
+	"strings"
+)
+
+type EnvProvider interface {
+	Environ() []string
```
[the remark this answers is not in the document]
j4k4: removed it as it is not used anywhere anymore

pkg/application/options.go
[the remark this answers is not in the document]
selm0: Can we name it as such? :D

pkg/application/options.go
```diff
@@ -288,18 +249,18 @@ func WithProducerDaemon(app *App) {
 }
 
 func WithTracing(app *App) {
-	app.addLoggerOption(func(config cfg.GosoConf, logger mon.GosoLog) error {
-		tracingHook := tracing.NewLoggerErrorHook()
+	app.addLoggerOption(func(config cfg.GosoConf, logger log.GosoLogger) error {
+		tracingHook := tracing.NewLoggerErrorHandler()
```
ajscholl, line 253: Should we rename this from hook to handler in the variable?

pkg/cfg/config.go
```diff
@@ -54,15 +53,16 @@ type config struct {
...
+var templateRegexp = regexp.MustCompile(`{([\w.\-]+)}`)
+var keyToEnvRegexp = regexp.MustCompile(`\[(\d)\]`)
 
 func New() GosoConf {
-	return NewWithInterfaces(os.LookupEnv)
+	return NewWithInterfaces(NewOsEnvProvider())
 }
 
-func NewWithInterfaces(lookupEnv LookupEnv) GosoConf {
+func NewWithInterfaces(lookupEnv EnvProvider) GosoConf {
```
ajscholl, line 63: Shouldn't the argument now be called envProvider?

pkg/cfg/config.go
```diff
@@ -504,20 +504,68 @@ func (c *config) mergeStruct(prefix string, settings interface{}, options ...Mer
 	return c.mergeMsi(prefix, msi, options...)
 }
 
-func (c *config) readEnvironment(prefix string, input *mapx.MapX) *mapx.MapX {
+func (c *config) readEnvironmentFromStructKeys(prefix string, structKeys []mapx.StructKey) *mapx.MapX {
+	environment := mapx.NewMapX()
+
+	var key, envKey string
```
ajscholl, line 510: Why are these outside of the loop? They seem to be used only at some places and are always assigned before they are used

pkg/cfg/config.go
```diff
@@ -531,8 +579,7 @@ func (c *config) resolveEnvKey(prefix string, key string) string {
 		key = strings.Join([]string{prefix, key}, ".")
 	}
 
-	rp := regexp.MustCompile(`\[(\d)\]`)
-	matches := rp.FindAllStringSubmatch(key, -1)
+	matches := keyToEnvRegexp.FindAllStringSubmatch(key, -1)
```
ajscholl, line 582: So we are transforming foo[3] to foo.3 here (or in the loop following this)? What happens if we get foo[32]? That doesn't match with a single digit, right?

pkg/log/formatter.go
```diff
@@ -0,0 +1,84 @@
...
+
+func FormatterJson(timestamp string, level int, format string, args []interface{}, err error, data Data) ([]byte, error) {
+	msg := fmt.Sprintf(format, args...)
+	jsn := make(map[string]interface{}, 8)
+
+	if err != nil {
+		jsn["err"] = err.Error()
+	}
+
+	jsn["channel"] = data.Channel
```
ajscholl, line 53: Wouldn't it be nicer if we would be using a go struct with the correct annotations (e.g., omit_empty for errors) and nice types?

pkg/log/handler_iowriter_writer_file.go
```diff
@@ -0,0 +1,30 @@
...
+	config.UnmarshalKey(configKey, settings)
+
+	return NewIoWriterFile(settings.Path)
+}
+
+func NewIoWriterFile(path string) (io.Writer, error) {
+	var err error
+	var file *os.File
+
+	if file, err = os.OpenFile(path, os.O_APPEND|os.O_WRONLY|os.O_CREATE, 0600); err != nil {
```
ajscholl, line 25: What do we think about buffering the writes to a file? Does the os.File struct already take care of that?

pkg/log/handler_sentry.go
```diff
@@ -0,0 +1,82 @@
...
+	return NewHandlerSentry(config)
+}
+
+type HandlerSentry struct {
+	hub *sentry.Hub
+}
+
+func NewHandlerSentry(config cfg.Config) (*HandlerSentry, error) {
+	env := config.GetString("env")
+	appName := config.GetString("app_name")
```
ajscholl, line 24: At some point we should add methods for these operations to the config package, otherwise we will never be able to rename `app_name` should we eventually want to do that

pkg/log/logger.go
```diff
@@ -0,0 +1,224 @@
...
+		},
+		ctxResolvers: nil,
+		handlers:     handlers,
+	}
+}
+
+func (l *gosoLogger) Option(options ...Option) error {
+	for _, opt := range options {
+		if err := opt(l); err != nil {
+			return fmt.Errorf("can not apply option %T", err)
```
ajscholl, line 111: Why %T? Shouldn't it be %w to embed the error?

pkg/application/options.go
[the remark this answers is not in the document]
j4k4: I would rather keep it like it is..  
Imagine a second builtin handler like `NewJsonHandler`..  
Just because one of these is used as the default, I don't have to name it DefaultHandler..  
For me it's more obvious, that the cli format is the default format..

pkg/cfg/environment.go
```diff
@@ -0,0 +1,46 @@
+package cfg
+
+import (
+	"os"
+)
+
+type EnvProvider interface {
+	LookupEnv(key string) (string, bool)
+	Setenv(key string, value string) error
```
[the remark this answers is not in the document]
j4k4: renamed

pkg/application/options.go
j4k4: renamed

pkg/cfg/config.go
j4k4: renamed

pkg/cfg/config.go
j4k4: true, not needed anymore after the latest changes

pkg/log/logger.go
j4k4: fixed

pkg/log/formatter.go
j4k4: Is it more performant using a map?  
Struct would involve costly reflection usage with every log message? :thinking:

pkg/log/handler_sentry.go
j4k4: Yes, this has to be improved somewhen.. but's that's a big change by itself..

pkg/log/handler_iowriter_writer_file.go
j4k4: Don't know to be honest..  
But I don't expect this to be an issue currently..

pkg/cfg/config.go
j4k4: added a `+`

pkg/log/formatter.go
ajscholl: Not sure what is going on in go here... but for me a struct is actually twice as fast as a map (3us instead of 6us/iteration). I am just using 

```go
type JsonStruct struct {
	Error     string                 `json:"err,omitempty"`
	Channel   string                 `json:"channel"`
	Level     int                    `json:"level"`
	LevelName string                 `json:"level_name"`
	Timestamp string                 `json:"timestamp"`
	Message   string                 `json:"message"`
	Fields    map[string]interface{} `json:"fields"`
	Context   map[string]interface{} `json:"context"`
}
```

so there isn't going anything fancy on here... maybe it can allocate the struct on the stack or something like that? Or reflection isn't as expensive here?

pkg/log/formatter.go
```diff
@@ -0,0 +1,84 @@
...
+	jsn["level"] = level
+	jsn["level_name"] = LevelName(level)
+	jsn["timestamp"] = timestamp
+	jsn["message"] = msg
+	jsn["fields"] = data.Fields
+	jsn["context"] = data.ContextFields
+
+	serialized, err := json.Marshal(jsn)
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal fields to JSON, %v", err)
```
ajscholl, line 63: Just realized that you want %w here and not %v

pkg/log/formatter.go
j4k4: okay, switched to struct

pkg/log/formatter.go
j4k4: fixed

ajscholl (approved):

pkg/application/options.go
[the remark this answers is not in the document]
j4k4: obsolete due to named handlers

selm0 (approved):

pkg/application/app.go
```diff
@@ -94,14 +93,18 @@ func NewWithInterfaces(config cfg.GosoConf, logger mon.GosoLog, options ...Optio
...
+
 	for _, opt := range app.loggerOptions {
 		if err = opt(config, logger); err != nil {
 			return nil, fmt.Errorf("can not apply logger options on application: %w", err)
 		}
 	}
 
-	if err = cfg.ApplyPostProcessors(config, logger); err != nil {
-		return nil, fmt.Errorf("can not apply post processor on config: %w", err)
+	for name, priority := range cfgPostProcessors {
```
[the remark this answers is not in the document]
j4k4: yeah..  but needed to  change the order.. otherwise the processor for the logger wasn't applied on time..

PR #539, opened 2021-07-30
deps: updated gin-gonic from 1.6.3 to 1.7.7;

ss89:


j4k4 (approved):
+1

applike-ss:
We can't merge this yet, because i found an issue that was introduced in gin 1.7.0 which does add a constraint under which circumstances the headers for client ip determination are trusted. Gin supposes one to use the .Run func to start it, however we currently don't make use of that.

applike-ss:
This issue is about the breaking change they did: https://github.com/gin-gonic/gin/issues/2697

Their project management seems a bit strange, because their master does contain the fix and on top of it is version v1.7.3, however the tag v1.7.3 does NOT include the fix

applike-ss:
finally gin-gonic 1.7.7 release did fix our issue, so checking it with our apps and will merge afterwards

PR #542, opened 2021-08-03
apiserver: improve bind error message if the client lost the connection;

ajscholl:
If a client sends a request, but then loses the connection, we would log `bind error - EOF` without really making clear what happened. How we instead log `network error - client has gone away - EOF`, which is hopefully easier to understand.


j4k4 (approved):
+1

PR #548, opened 2021-08-05
ddb: Removed the `WithRange` method from the `QueryBuilder` interface;

ajscholl:
`WithRange` was a dangerous method - if you had code which queried using `GetItem` and changed that to `Query` and your range query used a string as the range value, your code would still compile - but fail at runtime. Instead, you should've been using `WithRangeEq`. Now your code will fail to compile and you can immediately fix the problem instead of seeing it much later.


j4k4 (approved):
+1

PR #558, opened 2021-08-16
log: fix potential buffer overflow;

ss89:


applike-ss:
Instead of simply fixing an unlikely to happen overflow issue, we should rather limit the amount of fields and context fields on the logger.
We can add a `log.max_fields` and `log.max_context_fields` with a sensible default of something like 1024.
Additionally we should log a warning, when someone tries to add more fields than the current limit onto a separate channel `logger_max_fields` so that it can be filtered easily.

ss89:
i will close this down in favour of the suggested solution

PR #570, opened 2021-08-30
create new response with body, status-code, content-type, header

SamuelTheJackson:
- WithBody and WithContentType are chainable functions for easier usage

### DIFF
--- a/pkg/apiserver/handler.go
+++ b/pkg/apiserver/handler.go
@@ -3,15 +3,16 @@ package apiserver
 import (
 	"context"
 	"fmt"
+	"io/ioutil"
+	"net/http"
+	"net/url"
+
 	"github.com/applike/gosoline/pkg/coffin"
 	"github.com/applike/gosoline/pkg/mdl"
 	"github.com/gin-gonic/gin"
 	"github.com/gin-gonic/gin/binding"
 	"github.com/pkg/errors"
 	"github.com/spf13/cast"
-	"io/ioutil"
-	"net/http"
-	"net/url"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


pkg/apiserver/handler.go
```diff
@@ -22,20 +23,22 @@ const (
 
 var ErrAccessForbidden = errors.New("cant access resource")
 
+type Body interface{}
```
applike-ss, line 26: doesn't have value, since it doesn't add any type-safety here

pkg/apiserver/handler.go
```diff
@@ -22,20 +23,22 @@ const (
...
+
 type Request struct {
+	Body     interface{}
+	ClientIp string
```
applike-ss, line 30: doesn't apply to our coding guidelines anymore

pkg/apiserver/handler.go
```diff
@@ -22,20 +23,22 @@ const (
...
 
 // Don't create a response directly, use New*Response instead
 type Response struct {
-	StatusCode  int
+	Body        interface{}
 	ContentType *string // might be nil
 	Header      http.Header
-	Body        interface{}
+	StatusCode  int
 }
```
applike-ss, line 42:
```suggestion
type Response struct {
	Header      http.Header
	Body        interface{}
	ContentType *string // might be nil
	StatusCode  int
}
```

pkg/apiserver/handler.go
```diff
@@ -90,6 +93,18 @@ func (r *Response) AddHeader(key string, value string) {
...
 
+func (r *Response) WithBody(body Body) *Response {
+	r.Body = body
+
+	return r
+}
+
+func (r *Response) WithContentType(contentType string) *Response {
+	r.ContentType = mdl.String(contentType)
```
applike-ss, line 103:
```suggestion
	r.ContentType = &contentType
```

applike-ss (changes requested):

applike-ss (approved):

PR #579, opened 2021-09-16
unique-id: add module and generator;

selm0:


pkg/log/handler_sentry_context.go
```diff
@@ -18,8 +20,7 @@ func SentryContextConfigProvider(config ConfigProvider, handler *HandlerSentry)
 }
 
 func SentryContextEcsMetadataProvider(_ ConfigProvider, handler *HandlerSentry) error {
-	ecsMetadata, err := ReadEcsMetadata()
-
+	ecsMetadata, err := ecs.ReadEcsMetadata()
```
ajscholl, line 23: If the package is already called `ecs`, the method could just be called `ReadMetadata`

pkg/unique-id/instance_registerer_test.go
```diff
@@ -0,0 +1,109 @@
...
+		MachineId: 0,
+		CreatedAt: s.clock.Now(),
+	}
+
+	s.repository.
+		On("Persist", ctx, instance).
+		Return(nil)
+
+	registeredInstance, err := s.registerer.Register(ctx)
+	require.NoError(s.T(), err)
```
ajscholl, line 62: What is the difference between `require.NoError`, `assert.NoError` and `s.NoError`? They all ensure there is no error, right?

pkg/unique-id/instance_repository.go
```diff
@@ -0,0 +1,103 @@
...
+	GetAll(ctx context.Context) ([]Instance, error)
+}
+
+type instanceRepository struct {
+	ddbRepo ddb.Repository
+}
+
+type Instance struct {
+	MachineId uint      `json:"machineId" ddb:"key=hash"`
+	TaskArn   string    `json:"ip"`
```
ajscholl, line 35: `ip`?

pkg/unique-id/module.go
```diff
@@ -0,0 +1,126 @@
...
+	if um.instance, err = NewGenerator(config, logger); err != nil {
+		return nil, err
+	}
+
+	return um.instance, nil
+}
+
+func ProvideModule() kernel.ModuleFactory {
+	return func(_ context.Context, config cfg.Config, logger log.Logger) (kernel.Module, error) {
+		if um.instance != nil {
```
ajscholl, line 47: Can't this be replaced with a call to `ProvideGenerator`?

pkg/unique-id/module.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+}
+
+func (m *uniqueIdGenerator) Run(ctx context.Context) error {
+	<-ctx.Done()
+
+	// prevent further generations
+	m.generator = nil
+
+	// need a new context here, because the run context is already finished
```
ajscholl, line 109: Hmm, should your context eventually cancel after a timeout? Otherwise you could block here if something needs a lot of retries... and the kernel would kill itself. However, not sure if that really makes a difference

pkg/unique-id/module.go
```diff
@@ -0,0 +1,126 @@
...
+	"sync"
+
+	"github.com/applike/gosoline/pkg/cfg"
+	"github.com/applike/gosoline/pkg/kernel"
+	"github.com/applike/gosoline/pkg/log"
+)
+
+type uniqueIdGenerator struct {
+	kernel.EssentialBackgroundModule
+	kernel.EssentialStage
```
ajscholl, line 15: So your module should not write metrics because the metric daemon is also in this stage

pkg/unique-id/module.go
```diff
@@ -0,0 +1,126 @@
...
+
+	return nil
+}
+
+func (m *uniqueIdGenerator) NextID() (uint64, error) {
+	if m.generator == nil {
+		return 0, fmt.Errorf("generator not initialized, you need to use the module")
+	}
+
+	return m.generator.NextID()
```
ajscholl, line 125: This could still crash because another thread could just have set generator to nil if you are unlucky. If you want to be sure, copy it to a local variable and check that

ajscholl (changes requested):

pkg/apiserver/param_reader.go
```diff
@@ -13,14 +14,28 @@ func GetUintFromRequest(request *Request, name string) (*uint, bool) {
...
 }
 
+func GetUint64FromRequest(request *Request, name string) (*uint64, bool) {
+	paramString, found := request.Params.Get(name)
+
+	if !found {
+		return mdl.Uint64(0), false
+	}
+
+	param, err := strconv.ParseInt(paramString, 10, 0)
```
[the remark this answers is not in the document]
selm0: As discussed, I switched to int64 and adapted accordingly

pkg/unique-id/module.go
selm0: Indeed, and luckily it's not

pkg/unique-id/module.go
selm0: Only partially, as `Generator` and `kernel.Module` don't share any interfaces. I extracted it into a separate function

pkg/unique-id/module.go
selm0: Always depends on the shutdown scenario. I don't think that will really happen, but I added a 5 second timeout to the context

pkg/unique-id/module.go
ajscholl: So ddb writes no metrics? Okay, that is good, wasn't sure about that

pkg/apiserver/param_reader.go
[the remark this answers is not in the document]
ajscholl: Your code still fails on 32 bit platforms because specifying a size of 0 will cause it to fail on values with more than 32 bits needed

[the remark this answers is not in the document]
selm0: I didn't push yet.. ^^

pkg/uniqueid/instance_registerer.go
```diff
@@ -0,0 +1,96 @@
...
+func NewInstanceRegisterWithInterfaces(clk clock.Clock, repository InstanceRepository, settings Settings) *instanceRegisterer {
+	return &instanceRegisterer{
+		clock:      clk,
+		repository: repository,
+		settings:   settings,
+	}
+}
+
+func (r *instanceRegisterer) Register(ctx context.Context) (*Instance, error) {
+	for i := uint16(0); i < r.settings.MaxInstances; i++ {
```
ajscholl, line 69: So how does this work in detail if I have a service with 500 instances running? Will this mean I have to make 500 calls?

pkg/uniqueid/module.go
```diff
@@ -0,0 +1,123 @@
...
+
+	logger = logger.WithChannel("unique-id")
+
+	registerer, err := NewInstanceRegisterer(ctx, config, logger, settings)
+	if err != nil {
+		return nil, fmt.Errorf("could not create instance registerer: %w", err)
+	}
+
+	// pass application context once available
+	instance, err := registerer.Register(context.Background())
```
ajscholl, line 67: Can this now be `ctx` and the comment resolved?

pkg/uniqueid/module.go
```diff
@@ -0,0 +1,123 @@
...
+	m.logger.Info("unregistered instance")
+
+	return nil
+}
+
+func (m *uniqueIdGenerator) NextId() (*int64, error) {
+	generator := m.generator
+
+	if generator == nil {
+		return nil, fmt.Errorf("generator not initialized, you need to use the module")
```
ajscholl, line 119: In what situation does this happen when I forget to use the module? `Run` only sets it to nil after it is done, if I don't use the module, doesn't it just work (but fail to unregister at the end)?

selm0:
Closing PR due to different approach

PR #599, opened 2021-09-27
parquet: read parquet file into pointer to slice target;

bogdanfinn:


pkg/parquet/reader.go
```diff
@@ -37,6 +37,7 @@ type (
 type Reader interface {
 	ReadDate(ctx context.Context, datetime time.Time, target interface{}) error
 	ReadDateAsync(ctx context.Context, datetime time.Time, target interface{}, callback ResultCallback) error
+	ReadFileIntoTarget(ctx context.Context, file string, target interface{}) error
```
applike-ss, line 40: Not part of this PR, but we should refactor the naming here. File is technically incorrect, because it is an s3 key that we use.

applike-ss, line 40: We might also want to be able to define a batch size and an offset if that is possible. That is to not use a GB of memory just because we are deserializing all data of a file at once.

applike-ss: i created #600 for this

pkg/parquet/reader.go
```diff
@@ -37,8 +37,8 @@ type (
 type Reader interface {
 	ReadDate(ctx context.Context, datetime time.Time, target interface{}) error
 	ReadDateAsync(ctx context.Context, datetime time.Time, target interface{}, callback ResultCallback) error
-	ReadFileIntoTarget(ctx context.Context, file string, target interface{}) error
-	ReadFile(ctx context.Context, file string) (ReadResults, error)
+	ReadFileIntoTarget(ctx context.Context, file string, target interface{}, batchSize *int, offset int64) error
+	ReadFile(ctx context.Context, file string, batchSize *int, offset int64) (ReadResults, error)
```
applike-ss, line 41: i think if we want to make it the way it is done in the go sdk, then we'd not make it pointers but rather check for a -1 value => infinite

pkg/parquet/reader.go
```diff
@@ -220,6 +220,17 @@ func (r *s3Reader) ReadFile(ctx context.Context, file string) (ReadResults, erro
 	}
 
 	size := int(pr.GetNumRows())
+
+	if batchSize != nil {
+		err := pr.SkipRows(offset)
+
+		if err != nil {
```
applike-ss, line 227:
```suggestion
		err = pr.SkipRows(offset)
		if err != nil {
```

pkg/parquet/reader.go
```diff
@@ -220,6 +220,17 @@ func (r *s3Reader) ReadFile(ctx context.Context, file string) (ReadResults, erro
...
+			return nil, err
+		}
+
+		size = mdl.EmptyIntIfNil(batchSize)
```
applike-ss, line 231: imagine we have a parquet file of 99 entries and a batch size of 10, would we need to ensure that size=9 for the last batch?

pkg/parquet/reader.go
bogdanfinn: thank you

pkg/parquet/reader.go
```diff
@@ -183,7 +184,29 @@ func (r *s3Reader) ReadDateAsync(ctx context.Context, datetime time.Time, target
...
+
+	result, err := r.ReadFile(ctx, file, batchSize, offset)
+	if err != nil {
+		return fmt.Errorf("can not read file %s: %w", file, err)
+	}
+
+	decoded := refl.CreatePointerToSliceOfTypeAndSize(target, len(result))
+	err = r.decode(result, decoded)
+
+	if err != nil {
```
applike-ss, line 200:
```suggestion
	decoded := refl.CreatePointerToSliceOfTypeAndSize(target, len(result))

	err = r.decode(result, decoded)
	if err != nil {
```

pkg/parquet/reader.go
```diff
@@ -197,6 +220,23 @@ func (r *s3Reader) ReadFile(ctx context.Context, file string) (ReadResults, erro
...
+		}
+
+		leftRows := size - int(offset)
+
+		size = batchSize
+
+		if batchSize > leftRows {
+			size = leftRows
+		}
+	}
```
applike-ss, line 238:
```suggestion

	if offset > 0 {
		err = pr.SkipRows(offset)
		if err != nil {
			return nil, err
		}
	}

	if batchSize > 0 {
		remainingRows := size - int(offset)

		size = batchSize

		if batchSize > remainingRows {
			size = remainingRows
		}
	}
```

applike-ss: i can't add a suggestion below changes code, but i'd also move `results := make(ReadResults, size)` up after this block and return it `if size == 0`

PR #601, opened 2021-09-27
dx: added post processors for health check and config ports as well as automatically enabled path style for dev;

ajscholl:
Configure api.health.port to `"0"`, cfg.server.port to `"0"`, and `cloud.aws.s3.clients.default.usePathStyle` to `true` so you don't have to set them by hand.


j4k4 (changes requested):
Can we get this somehow more in line with the open closed principle?

j4k4 (approved):
+1

PR #608, opened 2021-10-11
kernel middleware and metadata additions

j4k4:


test/fixtures_s3_test.go
```diff
@@ -27,25 +26,19 @@ const (
...
-
 	return []suite.Option{
 		suite.WithLogLevel("debug"),
 		suite.WithConfigFile("test_configs/config.fixtures_s3.test.yml"),
 	}
 }
 
 func (s *FixturesS3Suite) SetupTest() (err error) {
-	s.logger = s.Env().Logger()
+	s.ctx = context.Background()
```
selm0, line 40: Why not use the env ctx here?

selm0 (approved):

test/fixtures_s3_test.go
j4k4: the context is not used in any "boot" functionality, so there is no need to use the environment/application context

PR #619, opened 2021-11-10
parquet: add ReadFileColumns function to reader

nikplxjt:
Currently our `parquet.Reader` only provides the capability to read a whole file into a target. In that process all columns of the parquet schema are read and tried to parse into the given schema. As parquet is a column based file format, it is easy for the format to just read specific columns. Therefore it would be nice to have a method to just read some columns and not the whole file.

### DIFF
--- a/pkg/parquet/reader.go
+++ b/pkg/parquet/reader.go
@@ -38,7 +38,6 @@ type Reader interface {
 	ReadDate(ctx context.Context, datetime time.Time, target interface{}) error
 	ReadDateAsync(ctx context.Context, datetime time.Time, target interface{}, callback ResultCallback) error
 	ReadFileIntoTarget(ctx context.Context, file string, target interface{}, batchSize int, offset int64) error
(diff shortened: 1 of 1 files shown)


pkg/parquet/reader.go
```diff
@@ -206,6 +221,87 @@ func (r *s3Reader) ReadFileIntoTarget(ctx context.Context, file string, target i
 	return nil
 }
 
+func (r *s3Reader) ReadFileColumns(ctx context.Context, columnNames []string, file string, batchSize int, offset int64) (ReadResults, error) {
+	columnNameSet := make(map[string]struct{})
```
selm0, line 225: I think we should make this the default behaviour, otherwise we only support this for a few use cases which is not really consistent (and leads to quite some almost-duplicates)

selm0 (changes requested):

selm0 (approved):

PR #621, opened 2021-11-11
Distinct entity metric changes

ajscholl:
Changes to allow us to track how many users performed some action (using HyperLogLogs in redis & writing the maximum number as a metric)


j4k4 (approved):
+1

PR #634, opened 2021-11-30
stream: forward appctx container to consumer

SamuelTheJackson:
Add a copyContainer function to copy a container from one context to an other

fix the stream consumer to work with the appctx

### DIFF
--- a/pkg/appctx/container.go
+++ b/pkg/appctx/container.go
@@ -57,3 +57,13 @@ func Provide(ctx context.Context, key interface{}, factory func() (interface{}, 
 
 	return val, nil
 }
+
+func CopyContainer(from, to context.Context) (context.Context, error) {
+	var contI interface{}
+
+	if contI = from.Value(containerKey); contI == nil {
+		return nil, &ErrNoApplicationContainerFound{}
+	}
+
+	return context.WithValue(to, containerKey, contI), nil
+}
--- a/pkg/stream/consumer_base.go
+++ b/pkg/stream/consumer_base.go
@@ -7,6 +7,7 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/justtrackio/gosoline/pkg/appctx"
 	"github.com/justtrackio/gosoline/pkg/cfg"
 	"github.com/justtrackio/gosoline/pkg/clock"
 	"github.com/justtrackio/gosoline/pkg/coffin"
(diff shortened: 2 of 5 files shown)


pkg/appctx/container.go
```diff
@@ -57,3 +57,13 @@ func Provide(ctx context.Context, key interface{}, factory func() (interface{},
 
 	return val, nil
 }
+
+func CopyContainer(from, to context.Context) (context.Context, error) {
```
ajscholl, line 61: Copy implies... well a copy. You are however sharing the container, so if I provide something in one container, it is also available in the other. I guess this is intended, but then the method should be called `ShareContainer`

pkg/stream/consumer_base.go
```diff
@@ -115,11 +116,20 @@ func (c *baseConsumer) run(kernelCtx context.Context, inputRunner func(ctx conte
...
+	dyingBackgroundCtxWithContainer, err := appctx.CopyContainer(kernelCtx, context.Background())
+	if err != nil {
+		return err
+	}
+
 	// create ctx whose done channel is closed on dying coffin
-	cfn, dyingCtx := coffin.WithContext(context.Background())
+	cfn, dyingCtx := coffin.WithContext(dyingBackgroundCtxWithContainer)
 
+	manualBackgroundCtxWithContainer, err := appctx.CopyContainer(kernelCtx, context.Background())
```
ajscholl, line 127: You only need a single base context here - a value context on top of a background context will never close their done channel, only the later wrapping (using `coffin.WithContext` and `cfn.Context`) allows them to die

pkg/stream/consumer_batch.go
```diff
@@ -83,11 +84,20 @@ func (c *BatchConsumer) Run(kernelCtx context.Context) error {
...
+func (c *BatchConsumer) run(ctx context.Context) (err error) {
 	logger := c.logger.WithContext(ctx)
+
 	defer logger.Debug("run is ending")
 	defer c.wg.Done()
-	defer c.processBatch(context.Background())
+	defer func() {
+		var ctxWithContainer context.Context
+
+		if ctxWithContainer, err = appctx.CopyContainer(ctx, context.Background()); err != nil {
```
ajscholl, line 95: Not sure what I think about overwriting the previous error, doesn't matter I guess because this error here should never happen...

ajscholl (changes requested):

PR #640, opened 2021-12-09
clock: added `NewTimer` function to allow mocking more functionality;

ajscholl:
When you need to wait for some time OR until the context expires, you can't use `clock.After` (that would leak the timer created in the background as you can't call `Stop` on a channel). Thus, you need to use a `Timer`, but those are currently not mocked. This commit allows you to mock those timers and control them from your unit test as needed.


j4k4 (approved):
+1

PR #645, opened 2021-12-18
fixtures: adjusted fixture writer/purger structure;

### LINKED ISSUE #644: S3 integration component is always loaded
When i write an integration test component the s3 component is always loaded even though my config that is used does not have any cloud configuration.

This seems to come from the relatively new PostProcessor `UseLocalstackDefaultsConfigPostProcessor`.

For whatever reason the init of the `pkg/cloud/aws/s3/client.go` is executed when i run my integration test.

When i set a breakpoint to that init func, there is no stacktrace that indicates that the source would be either in my or gosolines code (though that has to be the case).

See attached picture.

image

ss89:
fixes #644


ss89:
this might be one for you @j4k4

ss89:
@j4k4 i think you wanted to change something in terms of fixture loading. Do you already have a plan what should be changed? I think we didn't talk about that in detail. Maybe i could implement that in this PR or in a separate one, so i can get this merged?

PR #647, opened 2021-12-18
test: raised dynamodb,localstack,mysql,toxiproxy,wiremock version and…

ss89:
… adjusted readme;

PR #648, opened 2021-12-18
Ensure test component images are pulled prior test execution

### LINKED ISSUE #646: Integration test execution should ensure image is pulled prior test execution
Currently when starting from a freshly installed docker instance all images are pulled on-demand, rather then prior the test execution.

On slower internet connections this can lead to test failure because container 1 was already shut down due to expiry while container 2's image was still being pulled from the registry.

This (ignoring go's time limit) could be overcome if we ensured the images were pulled before trying to start all containers that are needed for the test execution.

to test this locally follow this:
```shell
go clean -testcache
docker prune system -a #hit y and enter to let docker prune all images and non-running containers off of your system
cd test
go test -tags integration,fixtures ./...
```

ss89:
fixes #646 and includes #647 for testability

### DIFF
--- a/pkg/test/component_cloudwatch.go
+++ b/pkg/test/component_cloudwatch.go
@@ -29,10 +29,8 @@ func (c *cloudwatchComponent) Boot(config cfg.Config, _ log.Logger, runner *dock
 	config.UnmarshalKey(key, c.settings)
 }
 
-func (c *cloudwatchComponent) Start() error {
-	containerName := fmt.Sprintf("gosoline_test_cloudwatch_%s", c.name)
-
-	return c.runner.Run(containerName, &containerConfigLegacy{
+func (c *cloudwatchComponent) getContainerConfig() *containerConfigLegacy {
+	return &containerConfigLegacy{
 		Repository: "localstack/localstack",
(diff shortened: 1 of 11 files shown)


ss89:
this might be one for you @j4k4

pkg/test/env/environment.go
```diff
@@ -91,6 +91,10 @@ func NewEnvironment(t *testing.T, options ...Option) (*Environment, error) {
 		return env, fmt.Errorf("can not create container runner: %w", err)
 	}
 
+	if err = env.runner.PullContainerImages(skeletons); err != nil {
```
j4k4, line 94: Does it make sense to integrate this call into `RunContainers`?  
Would it make a little bit more convenient maybe.

applike-ss: yeah i thought about that too. Will adjust that, because it's unlikely that someone would want to only use the container runner without the test system around it

applike-ss: adjusted that

j4k4 (approved):
+1

applike-ss:
rebased and squashed, will merge after successful test execution

PR #665, opened 2022-01-06
kernel: return an appropriate exit code on leaving the kernel;

j4k4:
Before this release, the kernel run function was just returning to the golang main function, which resulted in an exit code of 0 in normal cases. An exception was the forced stop which exited with code 1.

With this release, the kernel always exits with an appropriate exit code:  
- Code 0 on successful runs  
- Code 1 on errors during module boot or run
- Code 10 if there are no modules to run
- Code 11 if there are no foreground modules
- Code 12 if the kernel is forced to stop due to the kill timeout

### DIFF
--- a/pkg/kernel/kernel.go
+++ b/pkg/kernel/kernel.go
@@ -17,6 +17,16 @@ import (
 	"golang.org/x/sys/unix"
 )
 
+const (
+	ExitCodeOk           = 0
+	ExitCodeErr          = 1
+	ExitCodeNothingToRun = 10
+	ExitCodeNoForeground = 11
+	ExitCodeForced       = 12
+)
+
+type ExitHandler func(code int)
+
 //go:generate mockery --name Kernel
 type Kernel interface {
 	Add(name string, moduleFactory ModuleFactory, opts ...ModuleOption)
@@ -51,7 +61,9 @@ type kernel struct {
 	foregroundModules int32
 
 	killTimeout time.Duration
-	forceExit   func(code int)
+	exitCode    int
+	exitOnce    sync.Once
+	exitHandler ExitHandler
 }
 
 func New(ctx context.Context, config cfg.Config, logger log.Logger, options ...Option) (*kernel, error) {
@@ -69,7 +81,8 @@ func New(ctx context.Context, config cfg.Config, logger log.Logger, options ...O
 		started:   conc.NewPoisonedLock(),
 
 		killTimeout: time.Second * 10,
-		forceExit:   os.Exit,
+		exitCode:    ExitCodeErr,
+		exitHandler: os.Exit,
 	}
 
 	if err := k.Option(options...); err != nil {
@@ -116,6 +129,9 @@ func (k *kernel) AddMiddleware(middleware Middleware, position Position) {
 	}
 }
 
+// Run will boot and run the modules added to the kernel.
+// By default, os.Exit will get called if an error occurs or after the modules have stopped running,
+// which means that there will be no return out of this call.
 func (k *kernel) Run() {
 	// do not allow config changes anymore
 	if err := k.started.Poison(); err != nil {
@@ -124,24 +140,27 @@ func (k *kernel) Run() {
 		return
 	}
 
-	defer k.logger.Info("leaving kernel")
+	defer k.exit()
 	k.logger.Info("starting kernel")
 
 	if err := k.runMultiFactories(k.ctx); err != nil {
 		k.logger.Error("error building additional modules by multiFactories: %w", err)
 		close(k.running)
+		k.exitCode = ExitCodeErr
 		return
 	}
 
 	if len(k.moduleSetupContainers) == 0 {
 		k.logger.Warn("nothing to run")
 		close(k.running)
+		k.exitCode = ExitCodeNothingToRun
 		return
 	}
 
 	if err := k.runFactories(k.ctx); err != nil {
 		k.logger.Error("error building modules: %w", err)
 		close(k.running)
+		k.exitCode = ExitCodeErr
 		return
 	}
 
@@ -156,15 +175,16 @@ func (k *kernel) Run() {
 	}
 
 	if !k.hasModules() {
-		k.logger.Info("nothing to run")
 		close(k.running)
+		k.logger.Warn("nothing to run")
+		k.exitCode = ExitCodeNothingToRun
 		return
 	}
 
 	k.foregroundModules = int32(k.countForegroundModules())
 	if k.foregroundModules == 0 {
-		k.logger.Info("no foreground modules")
-		close(k.running)
+		k.logger.Warn("no foreground modules")
+		k.exitCode = ExitCodeNoForeground
 		return
 	}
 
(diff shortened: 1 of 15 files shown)


pkg/kernel/kernel.go
```diff
@@ -69,7 +80,8 @@ func New(ctx context.Context, config cfg.Config, logger log.Logger, options ...O
 		started:   conc.NewPoisonedLock(),
 
 		killTimeout: time.Second * 10,
-		forceExit:   os.Exit,
+		exitCode:    ExitCodeOk,
```
ajscholl, line 83: By default the exit code should be an error in my opinion and only be changed to success once we complete running the app - otherwise, if you panic, you would return an exit success, right?

pkg/kernel/kernel.go
```diff
@@ -120,24 +132,27 @@ func (k *kernel) Run() {
 	// do not allow config changes anymore
 	k.started.Poison()
 
-	defer k.logger.Info("leaving kernel")
+	defer k.exit()
```
ajscholl, line 135: Should we document that this function never returns normally (unless configured otherwise)?

pkg/kernel/kernel.go
j4k4: changed

pkg/kernel/kernel.go
j4k4: do you mean comment in the code or some explanation in the kernel docs?

ajscholl: Comment in the code

j4k4: added

ajscholl (approved):

pkg/kernel/kernel.go
```diff
@@ -182,6 +203,17 @@ func (k *kernel) Run() {
 		}
 
 		k.waitStopped()
+
+		hasErr := false
+		for _, stage := range k.stages {
+			if stage.err != nil && stage.err != ErrKernelStopping {
```
applike-ss, line 209: i wonder if it is ok for some modules to now properly shutdown.
Let's imagine there are metrics that weren't flushed to e.g. cloudwatch.
While the application did a great job and everything was good, we would still return an error code because metrics weren't flushed.
This is something we might want to know about, however we'd never check the exit code of an app for that - would we?
I think, we should only consider the application specific modules and not the whole module universe here.

pkg/kernel/kernel_test.go
```diff
@@ -459,7 +465,7 @@ func (s *slowExitModule) Run(_ context.Context) error {
 func TestModuleFastShutdown(t *testing.T) {
```
applike-ss, line 465: should we have a test for the ExitCodeNoForeground case?

pkg/log/status/module_test.go
```diff
@@ -115,7 +115,7 @@ func (m *testModule) WorkWithContext(ctx context.Context) error {
 }
 
 func TestModuleExample(t *testing.T) {
-	app := application.New()
+	app := application.New(application.WithKernelExitHandler(func(code int) {}))
```
applike-ss, line 118:
```suggestion
		exitCodeHandler := application.WithKernelExitHandler(func(code int) {})
		app := application.New(exitCodeHandler)
```

pkg/test/application.go
```diff
@@ -8,6 +8,7 @@ import (
 func Application() kernel.Kernel {
 	options := []application.Option{
 		application.WithConfigFile("./config.dist.yml", "yml"),
+		application.WithKernelExitHandler(func(code int) {}),
```
applike-ss, line 11: why do we have to overwrite this if there is a default with os.Exit?

test/stream/producer_daemon_test.go
```diff
@@ -115,7 +115,9 @@ func (s *ProducerDaemonTestSuite) TestWriteData() {
 		t: s.T(),
 	}
 
-	app := application.Default(application.WithLoggerHandlers(handler))
+	app := application.Default(application.WithLoggerHandlers(handler), application.WithKernelExitHandler(func(code int) {
+		assert.Equal(s.T(), kernel.ExitCodeOk, code, "exit code should be %d", kernel.ExitCodeOk)
+	}))
```
applike-ss, line 120:
```suggestion
	exitCodeHandler := func(code int) {
		assert.Equal(s.T(), kernel.ExitCodeOk, code, "exit code should be %d", kernel.ExitCodeOk)
	}
	handlers := []application.Option{
		application.WithLoggerHandlers(handler),
		application.WithKernelExitHandler(exitCodeHandler),
	}
	
	app := application.Default(handlers...)
```

applike-ss (changes requested):

pkg/kernel/kernel.go
```diff
@@ -211,6 +243,13 @@ func (k *kernel) Running() <-chan struct{} {
 	return k.running
 }
 
+func (k *kernel) exit() {
+	k.exitOnce.Do(func() {
+		k.logger.Info("leaving kernel with exit code %d", k.exitCode)
+		k.exitHandler(k.exitCode)
```
selm0, line 249: If I set an empty exit handler, will this then automatically always return an exit code of 0?

test/stream/producer_daemon_test.go
```diff
@@ -115,7 +115,9 @@ func (s *ProducerDaemonTestSuite) TestWriteData() {
 		t: s.T(),
 	}
 
-	app := application.Default(application.WithLoggerHandlers(handler))
+	app := application.Default(application.WithLoggerHandlers(handler), application.WithKernelExitHandler(func(code int) {
+		assert.Equal(s.T(), kernel.ExitCodeOk, code, "exit code should be %d", kernel.ExitCodeOk)
```
selm0, line 119: Maybe a short-hand exit code assertion helps to not duplicate this line over and over again :D

selm0 (commented):
Left some comments

pkg/kernel/kernel.go
j4k4: Leaving an empty exit handler will give back control to the caller..  
In our case this would lead to an exit code of 0, yes

test/stream/producer_daemon_test.go
j4k4: Could be, but as this code is already being refactored, I will leave it like this for now :)

pkg/test/application.go
j4k4: This one is used in tests where we can't use os.Exit..  
Furhtermore, this code will hopefully be gone soon..

pkg/kernel/kernel.go
j4k4: For now would I like to start with the whole module universe..  
But in general you could be right, that some "debugging" modules should not affect the exit code..

test/stream/producer_daemon_test.go
j4k4: As this test is already in progress of being refactored, I will leave it like this..

pkg/kernel/kernel_test.go
applike-ss: @j4k4

j4k4: added

PR #692, opened 2022-01-13
mdlsub: make publisher write to any output;

erdac:


j4k4:
not needed anymore due to v0.3.58

PR #710, opened 2022-01-18
stream: producer output loop uses empty context

j4k4:
use an empty ctx instead of the kernel one in the producer output loop


pkg/stream/producer_daemon.go
```diff
@@ -336,13 +338,7 @@ func (d *producerDaemon) outputLoop(ctx context.Context) error {
 
 		// no need to have some delayed cancel context or so here - if you need this, your output should've already provided that
```
ajscholl, line 339: I guess we can delete this comment then, too?

ajscholl (approved):

PR #737, opened 2022-02-01
stream: retry handler

j4k4:
This pr implements a retry handler for the consumer to add the retry mechanism into the stream framework instead of solely relying on the retry mechanism of the sqs system.  
For the mvp we attach a dedicated sqs queue to the consumer which is used as a retry queue.  
If a message isn't acknowledged it get's pushed into the retry queue with a delay and gets consumed from there.


pkg/stream/retry_io.go
```diff
@@ -0,0 +1,100 @@
...
+		ClientName:   "",
+		Unmarshaller: UnmarshallerMsg,
+	}
+
+	if input, err = NewSqsInput(ctx, config, logger, inputSettings); err != nil {
+		return nil, fmt.Errorf("can not create input: %w", err)
+	}
+
+	outputSettings := &SqsOutputSettings{
+		AppId:             settings.AppId,
```
ajscholl, line 43: I think I would initialize all fields here from the inputSettings, makes it clearer that the two are related

pkg/stream/retry_io.go
```diff
@@ -0,0 +1,100 @@
...
+	"context"
+	"fmt"
+
+	"github.com/justtrackio/gosoline/pkg/cfg"
+	"github.com/justtrackio/gosoline/pkg/cloud/aws/sqs"
+	"github.com/justtrackio/gosoline/pkg/log"
+)
+
+type RetryHandlerIO struct {
+	input    AcknowledgeableInput
```
ajscholl, line 13: We could embed the input, then we don't have to implement all the methods for the input to forward the calls to the input

pkg/stream/retry_io.go
```diff
@@ -0,0 +1,100 @@
...
+	output   Output
+	settings *RetrySettings
+}
+
+func NewRetryHandlerIO(ctx context.Context, config cfg.Config, logger log.Logger, settings *RetrySettings) (RetryHandler, error) {
```
ajscholl, line 18: I would make the type private and return an interface if possible

pkg/stream/retry_io.go
j4k4: done

pkg/stream/retry_io.go
j4k4: done

pkg/stream/retry_io.go
j4k4: it is returning an interface already? :thinking:

pkg/stream/input.go
```diff
@@ -30,3 +30,7 @@ type AcknowledgeableInput interface {
 	// service.
 	AckBatch(ctx context.Context, msgs []*Message) error
 }
+
+type RetryableInput interface {
```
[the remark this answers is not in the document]
j4k4: the plan is to use the mock in the unit tests

pkg/stream/consumer_base.go
```diff
@@ -194,16 +233,83 @@ func (c *baseConsumer) runConsumerCallback(ctx context.Context) error {
...
+				}
+			}
+		}
+	}
+}
+
+func (c *baseConsumer) shovelDataFromRetry(ctx context.Context) error {
+	defer c.logger.Debug("shovelDataFromRetry is ending")
+
+	for {
```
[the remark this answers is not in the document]
j4k4: done

pkg/stream/retry_io.go
ajscholl: Seems like I can't read, you are right

pkg/stream/input.go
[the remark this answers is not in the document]
ajscholl: So is this code currently missing the //go:generate comment? If so, I think the generated mock is still missing from the commit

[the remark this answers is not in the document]
j4k4: ah, this one.. for now I wasn't in need for an mock here..

test/stream/retry_handler/config.dist.yml
```diff
@@ -0,0 +1,34 @@
...
+    default:
+      retry:
+        enabled: true
+        after: 1s
+
+  input:
+    consumer:
+      type: kinesis
+      target_queue_id: postback-request
+      target_application: postback-service
```
ajscholl, line 23: Should it be named like this?

j4k4: removed

pkg/stream/consumer_base.go
```diff
@@ -83,11 +97,31 @@ func NewBaseConsumer(ctx context.Context, config cfg.Config, logger log.Logger,
...
 	})
 
+	retryHandler := NewRetryHandlerNoop()
+
+	// disable the retry handler for inputs which have a retry mechanism on its own
+	if retryAware, ok := input.(RetryableInput); ok {
+		settings.Retry.Enabled = settings.Retry.Enabled && !retryAware.HasRetry()
+	}
+
+	if settings.Retry.Enabled {
```
selm0, line 107: Should the retry resources also show up on the metadata server?

pkg/stream/consumer_base.go
```diff
@@ -194,16 +233,61 @@ func (c *baseConsumer) runConsumerCallback(ctx context.Context) error {
...
+	cfn := coffin.New()
+	cfn.GoWithContextf(ctx, c.shovelDataFromSource(c.input, "input"), "panic during shoveling data from input")
+	cfn.GoWithContextf(ctx, c.shovelDataFromSource(c.retryHandler, "retry handler"), "panic during shoveling data from retry")
+
+	return cfn.Wait()
+}
+
+func (c *baseConsumer) shovelDataFromSource(input Input, src string) func(ctx context.Context) error {
+	return func(ctx context.Context) error {
+		defer c.logger.Debug("shovelDataFromSource %s is ending", src)
```
selm0, line 249: Should we use the context here for the logger?

pkg/stream/consumer_base.go
```diff
@@ -194,16 +233,61 @@ func (c *baseConsumer) runConsumerCallback(ctx context.Context) error {
 	return nil
 }
 
+func (c *baseConsumer) shovelData(ctx context.Context) error {
+	defer c.logger.Debug("shovelData is ending")
```
selm0, line 237: Logger + Ctx?

pkg/stream/consumer_base.go
```diff
@@ -194,16 +233,61 @@ func (c *baseConsumer) runConsumerCallback(ctx context.Context) error {
 	return nil
 }
 
+func (c *baseConsumer) shovelData(ctx context.Context) error {
```
selm0, line 236: I'm not a big fan of using `shovel`, sounds a bit unprofessional to me :D

pkg/stream/input_sqs.go
```diff
@@ -212,6 +212,10 @@ func (i *sqsInput) AckBatch(ctx context.Context, msgs []*Message) error {
 	return multiError.ErrorOrNil()
 }
 
+func (i *sqsInput) HasRetry() bool {
```
selm0, line 215: Is there a designed way to overwrite this? In case of batch consumption, it's beneficial to not use the queue's retry mechanism but a custom one, so we can acknowledge all messages about reading.

In automation we sometimes have the problem that processing a message takes such a long time that it's already redelivered by the primary queue as we acknowledge it only after successful processing.

pkg/stream/consumer_acknowledge.go
```diff
@@ -18,28 +18,40 @@ func NewConsumerAcknowledgeWithInterfaces(logger log.Logger, input Input) Consum
...
-		return
+	ackInputs := make(map[AcknowledgeableInput][]*Message)
+
+	for _, data := range cdata {
+		if ackInput, ok = data.input.(AcknowledgeableInput); !ok {
+			continue
+		}
+
+		if _, ok = ackInputs[ackInput]; !ok {
+			ackInputs[ackInput] = make([]*Message, 0)
```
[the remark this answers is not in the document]
j4k4: changed

pkg/stream/consumer.go
```diff
@@ -62,43 +62,43 @@ func NewConsumerWithInterfaces(base *baseConsumer, callback ConsumerCallback) *C
...
-	defer c.logger.Debug("runConsuming is ending")
+func (c *Consumer) readData(ctx context.Context) error {
+	defer c.logger.Debug("read from input is ending")
 	defer c.wg.Done()
 
 	for {
 		select {
 		case <-ctx.Done():
-			return fmt.Errorf("return from consuming as the coffin is dying")
+			return nil
```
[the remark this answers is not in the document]
j4k4: the done channel is closed if another routine of the coffin had an error and that error is returned on cfn.Wait

pkg/stream/consumer.go
```diff
@@ -150,5 +148,18 @@ func (c *Consumer) process(ctx context.Context, msg *Message) bool {
 		c.handleError(ctx, err, "an error occurred during the consume operation")
 	}
 
+	if ack {
+		return true
+	}
+
+	// do not put it back into retry if it's already in there
+	if _, ok := msg.Attributes[AttributeRetry]; ok {
```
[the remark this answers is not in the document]
j4k4: As this is retry handler specific, it is handled in the sqs retry handler implementation now

pkg/stream/consumer_base.go
j4k4: the context shouldn't have any additional information here..

pkg/stream/consumer_base.go
j4k4: same

pkg/stream/input_sqs.go
j4k4: This sounds for me like an additional PR make the ack behaviour configurable?

pkg/stream/consumer_base.go
j4k4: renamed :D

pkg/stream/consumer_base.go
j4k4: Yes, makes sense, will add that..

pkg/stream/consumer_acknowledge.go
```diff
@@ -18,28 +18,40 @@ func NewConsumerAcknowledgeWithInterfaces(logger log.Logger, input Input) Consum
...
+func (c *ConsumerAcknowledge) AcknowledgeBatch(ctx context.Context, cdata []*consumerData) {
 	var ok bool
 	var ackInput AcknowledgeableInput
 
-	if ackInput, ok = c.input.(AcknowledgeableInput); !ok {
-		return
+	ackInputs := make(map[AcknowledgeableInput][]*Message)
+
+	for _, data := range cdata {
+		if ackInput, ok = data.input.(AcknowledgeableInput); !ok {
```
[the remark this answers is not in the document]
j4k4: I think it makes sense, to have the default ack mechanism as it is now, as it just makes sense to ack a message after processing it.. but like I stated already in the other comment I'm open to make that configurable, to always ack a message immediately..

PR #770, opened 2022-02-22
kinsumer delay

j4k4:
Enabled the kinsumer to process records with a configured delay.

### DIFF
--- a/pkg/cloud/aws/kinesis/kinsumer.go
+++ b/pkg/cloud/aws/kinesis/kinsumer.go
@@ -56,7 +56,7 @@ type Settings struct {
 	// Name of the stream (before expanding with project, env, family & application prefix)
 	StreamName string `cfg:"stream_name" validate:"required"`
 	// The shard reader will sleep until the age of the record is older than this delay
-	ConsumeDelay time.Duration `cfg:"consumer_delay" default:"0"`
+	ConsumeDelay time.Duration `cfg:"consume_delay" default:"0"`
 	// InitialPosition of a new kinsumer. Defines the starting position on the stream if no metadata is present.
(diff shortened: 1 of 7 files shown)


pkg/cloud/aws/kinesis/kinsumer.go
```diff
@@ -135,6 +138,7 @@ func NewKinsumer(ctx context.Context, config cfg.Config, logger log.Logger, sett
 	}
 
 	shardReaderFactory := func(logger log.Logger, shardId ShardId) ShardReader {
+		logger = logger.WithChannel("kinsumer-read")
```
ajscholl, line 141: I would argue this call belongs to the `NewShardReaderWithInterfaces` call and not here

pkg/cloud/aws/kinesis/shard_reader.go
```diff
@@ -463,5 +475,14 @@ func getShardReaderDefaultMetrics(stream Stream) metric.Data {
 			Unit:  metric.UnitCount,
 			Value: 0.0,
 		},
+		{
+			Priority:   metric.PriorityHigh,
+			MetricName: metricNameWaitDuration,
+			Dimensions: map[string]string{
+				"StreamName": string(stream),
+			},
+			Unit:  metric.UnitMillisecondsAverage,
```
ajscholl, line 484: An average metric is problematic with a default - you get wrong results I think. Think about what happens if you have a single call to write 100 to that metric:

- you start with an initialization of 0
- you add the 100
- you now have a sum of 100 from 2 datapoints
- the result we write to cw is 50

ajscholl (approved):

PR #860, opened 2022-04-05
sqs: return the error when GetQueueUrl error is not QueueDoesNotExist…

netsi:
When the `GetQueueUrl` fails we are not returning the error if it is not `QueueDoesNotExist` which makes it hard to find out what why the sqs consumer is not working.


pkg/cloud/aws/sqs/service.go
```diff
@@ -201,7 +201,7 @@ func (s *service) GetUrl(ctx context.Context, name string) (string, error) {
 			return "", nil
 		}
 
-		return "", nil
+		return "", err
```
j4k4, line 204: Do you mind adding an error wrap here? To add some context to the error?  
e.g. `fmt.Errorf("can not request queue url: %w", err)`  
I know it seems redundant from time to time, but out of experience it's better to have one information to much than to less..

netsi: sure

j4k4:
rebasing should fix the failing static code analysis job

j4k4 (approved):

PR #861, opened 2022-04-05
introduce grpc server pkg

netsi:


pkg/grpcserver/stats_logger.go
```diff
@@ -0,0 +1,114 @@
...
+const (
+	contextKey key = 0
+)
+
+type statsLogger struct {
+	logger   log.Logger
+	settings *Settings
+}
+
+func NewStatsLogger(settings *Settings, logger log.Logger) *statsLogger {
```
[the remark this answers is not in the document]
netsi: good catch, will reorder the args

examples/grpcserver/config.dist.yml
```diff
@@ -0,0 +1,10 @@
+env: dev
+app_project: project
+app_family: family
+app_name: grpc_server
+
+# grpcserver module configuration
```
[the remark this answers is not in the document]
netsi: good idea, changed it in a way to support multiple grpc servers and updated the example to actually have multiple servers running (one with health checks and one without)

pkg/grpcserver/server.go
```diff
@@ -0,0 +1,154 @@
...
+func New(name string, definer ServiceDefiner, middlewares ...MiddlewareFactory) kernel.ModuleFactory {
+	return func(ctx context.Context, config cfg.Config, logger log.Logger) (kernel.Module, error) {
+		var (
+			err         error
+			definitions *Definitions
+		)
+		settings := &Settings{}
+		config.UnmarshalKey(fmt.Sprintf("%s.%s", grpcServerConfigKey, name), settings)
+
+		logger = logger.WithChannel(grpcServiceChannel)
```
j4k4, line 52: Do you want to have the server name in the channel?  
Or add the server name as a field?  
To be able to distinguish between different servers?

pkg/grpcserver/server.go
```diff
@@ -0,0 +1,154 @@
+package grpcserver
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"net"
+	"net/http"
+
+	grpc_middleware "github.com/grpc-ecosystem/go-grpc-middleware"
```
j4k4, line 10: this looks like too many import groups..  
usually we have 2 groups only.. stdlib and all the rest..

j4k4, line 10: this looks like too many import groups..  
usually we have 2 groups only.. stdlib and all the rest..

pkg/grpcserver/server_test.go
```diff
@@ -0,0 +1,174 @@
+package grpcserver_test
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
```
j4k4, line 9: Again the thing with the import groups

j4k4, line 9: Again the thing with the import groups

pkg/grpcserver/stats_logger.go
```diff
@@ -0,0 +1,114 @@
+package grpcserver
+
+import (
+	"context"
+	"sync"
+
+	"google.golang.org/grpc/stats"
```
j4k4, line 7: import groups

j4k4, line 7: import groups

pkg/grpcserver/server.go
netsi: my bad was using `goimports`, will fix it for all the files I introduce

pkg/grpcserver/server.go
netsi: I think it would be better to have it as a field, will add it

j4k4 (approved):

PR #871, opened 2022-04-13
fixed data race in clock package

Miraddo:
You had a data race in your clock package.

`go test -race -v .\pkg\clock\`

image
image

### DIFF
--- a/pkg/clock/fake_ticker.go
+++ b/pkg/clock/fake_ticker.go
@@ -50,11 +50,11 @@ func (f *fakeTicker) Reset(d time.Duration) {
 		panic(errors.New("non-positive interval for Reset"))
 	}
 
-	f.remaining = d
-	f.duration = d
-
 	f.clock.lck.Lock()
 	defer f.clock.lck.Unlock()
+
+	f.remaining = d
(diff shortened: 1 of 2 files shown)


ajscholl:
Thanks for finding this, but I merged #809 instead (which also should fix this data race along many others found by the race detector). I guess we can close this PR then?

Miraddo:
I rechecked the package. It still has the data race, 
It's not about the test. 
The data race will happen in the package clock in the fake_ticker line 53 and fake_timer line 47.
If you find out that has the data race is good in some of your scenarios so you can add `// +build !race`

ajscholl:
Are you sure this is still a problem? The lines you mentioned now look like this since I merged #809:
```golang
f.lck.Lock()
f.remaining = d
f.duration = d
f.lck.Unlock()
```
and this:
```golang
f.lck.Lock()
f.remaining = d
f.lck.Unlock()
```
So now there should be a synchronization point preventing the race, right?

Miraddo:
> So now there should be a synchronization point preventing the race, right?

Sorry, you are right.
It was my fault, It's work now without data race.

PR #872, opened 2022-04-13
Introduce Kafka

abdelrahman-t:
Introduce Kafka Consumer/Producer

will be handled in a different ticket:
-  Deadlettering and retries.


j4k4:
Hey,

thanks for contributing your Kafka implementation.
To kick things off, I would like to suggest some structural changes to the packaging and config.

As the stream package usually is an additional layer above the "transportation" layer, I would like to move the kafka package out of there and have it more like the redis one as a top level package for now.
In addition I think it would be nice to have the connection config handled like the other clients, e.g. the aws clients:

```yaml
kafka:
  connections:
    default:
      bootstrap_servers: ["bootstrap:9094"]
      insecure_skip_verify: true
      username: username
      password: password

stream:
  output:
    datalake:
      type: kafka
      topic: "topic"
      async_write: true
      connection: default

  input:
    datalake:
      type: kafka
      topic: "topic"
      connection: default
```

This way it should be quite easy to have a central pool of kafka connections which can be reused in multiple inputs, outputs and even applications..

pkg/stream/input.go
```diff
@@ -29,6 +29,9 @@ type AcknowledgeableInput interface {
 	// AckBatch does the same as calling Ack for every single message would, but it might use fewer calls to an external
 	// service.
 	AckBatch(ctx context.Context, msgs []*Message) error
+	// Nack does a negative acknowledgement of a message. If possible,
+	// prefer calling Nack with a batch as it is more efficient.
+	Nack(ctx context.Context, msg ...*Message) error
```
j4k4, line 34: Do I understand correctly, that calling Nack will be used to send messages into a dead letter queue?  
As the stream.consumer has already implemented a retry logic, I'm not sure if it makes sense to have something similar on the inputs.  
Do you plan to use the inputs independently or in conjunction with a consumer?  
If you just want to send messages into the dead letter queue instead of retrying them, we could think about improving the consumer retry handler.

pkg/stream/kafka/connection/config.go
```diff
@@ -0,0 +1,11 @@
+package connection
+
+type Conf struct {
```
j4k4, line 3: In the gosoline ecosystem, we call these structs just `Settings`.

pkg/stream/input.go
abdelrahman-t: 1. Yes, Nack will be used to send messages into a dead letter queue, since there are no per-message acks in kafka, but offsets, usually the way to do **non-blocking** retries is to dead-letter the message and have another consumer retry the message (from the dead-letter/retry topic), similar to approach explained here: https://eng.uber.com/reliable-reprocessing/

2. In conjunction with a consumer.

3. Plan would be similar to the approach in the article above, to achieve non-blocking retries by pushing to a retry/dead-letter topic and have another consumer (maybe parallel consumer) consume from there.

we could maybe provide multiple retry handler implementations? what do you think?

j4k4: Gosoline has already something in place to support this retry design. We're using it with our Kinesis consumers as Kinesis don't have built-in retry behaviour, too. You can configure it like this for a consumer:
```yaml
stream:
  consumer:
    default:
      input: kinesis-input
      retry:
        enabled: true
        type: sqs
        after: 5m
        max_attempts: 6
        queue_id: triggers-retry
```
If retry is enabled and the message isn't acked, it is send into the retry queue and gets reprocessed from there by the same consumer. After the max attempts it will end up in a DLQ. SQS is the default retry handler which provides this behaviour. If you want to stick with Kafka for retry, too, you can add a second implementation for the retry handler, but could also start with SQS first. Have a look into `gosoline/pkg/stream/retry.go` about this.

abdelrahman-t: thanks for the insights :) I will be working on this today.

abdelrahman-t: Pushed.

pkg/stream/kafka/connection/config.go
abdelrahman-t: Pushed.

pkg/stream/consumer.go
```diff
@@ -148,11 +148,9 @@ func (c *Consumer) process(ctx context.Context, msg *Message) bool {
...
-	if ack {
-		return true
+	if !ack {
+		c.retry(ctx, msg)
 	}
 
-	c.retry(ctx, msg)
-
-	return ack
+	return ack || isHeadOfLineBlocking(c.input)
```
j4k4, line 155: I'm wondering if the consumer should be aware of these "implementation" details. Meaning that the consumer has to know about how the input is handling acknowledgement. If the next input has again a different acknowledgement pattern, it will get quite ugly here.

Maybe we should let the input decide what to do with a positive or negative ack. If we would change the interface to:
```golang
type AcknowledgeableInput interface {
	Input
	Ack(ctx context.Context, msg *Message, ack bool) error
	AckBatch(ctx context.Context, msgs []*Message, acks []bool) error
}
```
The SQS input can send the acknowledgement based on the ack value and the kafka input can just always send the commit. So the consumer would call Ack every time on the input not just in the success case. I guess this would make it easier, leaner and cleaner..

abdelrahman-t: makes sense, pushed :)

pkg/kafka/topic.go
```diff
@@ -0,0 +1,31 @@
...
+func DLTopicName(topic, groupID string) string {
+	return NormalizeTopicName(
+		fmt.Sprintf(
+			"%s.%s.dl",
+			topic,
+			groupID,
+		))
+}
+
+func namingStrategy(appId cfg.AppId, topic string) string {
```
j4k4, line 29: Is there any way to overwrite the naming strategy?  
That would be a good option to have.

abdelrahman-t:
Hi @j4k4 

Thanks for comments, will take a look later today, btw here are some of the issues I had while integrating the Kafka input/output, maybe we can think of somethings to make such use cases much easier in the future.

## Types
if stream.Message would be an interface, that would allow inputs to return arbitrary message types.
Problem: Currently I had to embed the original Kafka message as an attribute of stream.Message and at the boundaries (Kafka impl. <> Stream impl.), I convert back and forth between internal representation of the message and stream.Message.  

## Batching
Currently the Kafka input/output does batching internally because the Kafka lib has batching builtin.
Problem: I have to specify batching config twice, in both consumer and input/output level.

## Streams Vs Queues (head of line blocking inputs)
The latest changes to acking logic made this much easier :)

pkg/kafka/consumer/consumer.go
```diff
@@ -0,0 +1,196 @@
...
+	)
+
+	for j := 0; j < c.Settings.BatchSize; j++ {
+		c.Logger.Debug("fetching a message")
+
+		msg, err := c.Reader.FetchMessage(ctx)
+
+		if err != nil {
+			if !errors.Is(err, context.DeadlineExceeded) {
+				c.Logger.WithFields(log.Fields{"Error": err}).Error("failed to fetch message")
```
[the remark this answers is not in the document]
abdelrahman-t: lib does backoff, now error is returned and should bubble up to the kernel.

pkg/kafka/consumer/reader.go
```diff
@@ -0,0 +1,72 @@
...
+		// Explicit commits.
+		CommitInterval: CommitOffsetsSync,
+
+		// Safe defaults.
+		RetentionTime:  DefaultConsumerGroupRetentionTime,
+		MaxAttempts:    DefaultMaxRetryAttempts,
+		IsolationLevel: kafka.ReadCommitted,
+
+		Logger:      logging.NewKafkaLogger(logger).DebugLogger(),
+		ErrorLogger: logging.NewKafkaLogger(logger).ErrorLogger(),
```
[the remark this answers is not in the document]
abdelrahman-t: so you mean an error handler should be configurable? something like:

```
error_handler:
    level: "warning"|"error"|"fatal" # fatal would bubble up the error to the kernel.
```

pkg/stream/consumer.go
```diff
@@ -98,12 +98,15 @@ func (c *Consumer) processAggregateMessage(ctx context.Context, cdata *consumerD
...
-
+	ack := true
 	for _, m := range batch {
-		c.process(ctx, m)
+		if !c.process(ctx, m) {
+			ack = false
+		}
 	}
 
+	c.Acknowledge(ctx, cdata, ack)
```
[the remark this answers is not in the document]
abdelrahman-t: I assumed ack should be done on success only, restored original behaviour.

pkg/kafka/topic.go
abdelrahman-t: done.

pkg/kafka/consumer/reader.go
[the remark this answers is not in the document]
j4k4: I've just checked the existing inputs again and how they are handling internal errors..  
SQS basically does the same and directly logs it as an error.. So let's just keep it as it is for now..

PR #887, opened 2022-04-21
introduce prometheus metrics

netsi:
- apiserver: expose metrics endpoint
    - metric: add prometheus metric writer
    - examples: add prometheus metrics example


j4k4:
Hey, thank you for contributing a prometheus metric writer.

As far as I can see, you relying on many of the convenience functions which are provided by the prometheus library, right?  
Meaning, you're using the `DefaultRegisterer` from the global scope of the package. This limits us to have 1 prometheus metric writer per execution context only. As there is usually only 1 application running most of the time, this won't cause problems, but for example our integration tests could execute multiple applications in parallel which will result in unpredictable behaviour if these global scope variables are used. I would recommend to not use the convenience functions but instead create a dedicated `Registerer` and `Gatherer` for every application.

To expose the metrics via http you're relying on the apiserver package? I'm not sure if we should create a dependency here between these packages. As far as I can see, it's just a single handler? Would it be possible to use the stdlib net/http package to create a server to host this handler? Using the gin library for this seems a little bit heavy.

Are the cloudwatch and prometheus writers swappable? Meaning, can I just switch from `cw` to `prom` in the config and everything works? How is it working in conjunction with the metric daemon? I guess it is not needed for the prometheus writer? I'm just wondering how much sense it does make to try to put these things into the same package or if we should handle them more separately..

netsi:
@j4k4 

* I pushed a new commit that uses a custom `Registerer` and `Gatherer` and should allow us to have different writers per execution context

* removed the metrics endpoint from the apiserver (wouldn't the same point apply for the ApiHealthCheck?). Now it is using the `net/http` pkg

* The writters should be swappable for example you can get the `ApiRequestCount` metric produced by the `apiserver` handlers in prometheus 

example output from the application in the `examples/metrics/prometheus`
```
# HELP gosoline:dev:metrics_ApiRequestCount unit: Count
# TYPE gosoline:dev:metrics_ApiRequestCount counter
gosoline:dev:metrics_ApiRequestCount{path="/current-value"} 4
gosoline:dev:metrics_ApiRequestCount{path="/decrease"} 1
gosoline:dev:metrics_ApiRequestCount{path="/increase"} 7
# HELP gosoline:dev:metrics_ApiRequestResponseTime unit: Milliseconds
# TYPE gosoline:dev:metrics_ApiRequestResponseTime summary
gosoline:dev:metrics_ApiRequestResponseTime_sum{path="/current-value"} 0.2171715
gosoline:dev:metrics_ApiRequestResponseTime_count{path="/current-value"} 1
gosoline:dev:metrics_ApiRequestResponseTime_sum{path="/decrease"} 0.098895
gosoline:dev:metrics_ApiRequestResponseTime_count{path="/decrease"} 1
gosoline:dev:metrics_ApiRequestResponseTime_sum{path="/increase"} 0.2140137142857143
gosoline:dev:metrics_ApiRequestResponseTime_count{path="/increase"} 1
# HELP gosoline:dev:metrics_ApiStatus2XX unit: Count
# TYPE gosoline:dev:metrics_ApiStatus2XX counter
gosoline:dev:metrics_ApiStatus2XX{path="/current-value"} 4
gosoline:dev:metrics_ApiStatus2XX{path="/decrease"} 1
gosoline:dev:metrics_ApiStatus2XX{path="/increase"} 7
# HELP gosoline:dev:metrics_api_request unit: prom-counter
# TYPE gosoline:dev:metrics_api_request counter
gosoline:dev:metrics_api_request{handler="cur"} 4
gosoline:dev:metrics_api_request{handler="decr"} 1
gosoline:dev:metrics_api_request{handler="incr"} 7
```

I think that is the advantage of having the prometheus writter on the same package, we could simply swap the writer in the configs and have all the metrics we had in Cloudwatch in prometheus. There is no benefit of using the metric daemon though because there are no delays caused by pushing metrics somewhere.

j4k4:
> * wouldn't the same point apply for the ApiHealthCheck?

Yes, it does. Improving the applicatin health check is on the roadmap. :+1:

pkg/metric/writer_prometheus.go
```diff
@@ -0,0 +1,221 @@
...
+	metricLimit int64
+	metrics     *int64
+}
+
+func NewPromWriter(ctx context.Context, config cfg.Config, logger log.Logger) (*promWriter, error) {
+	settings := &promSettings{}
+	config.UnmarshalKey("metric.prometheus", settings)
+
+	appId := cfg.GetAppIdFromConfig(config)
+	namespace := fmt.Sprintf("%s:%s:%s", appId.Project, appId.Environment, appId.Family)
```
j4k4, line 55: So the namespace is on family level? How do you distinguish between applications? I would have expected the app name here as well.

pkg/metric/prometheus/http/metric_server.go
```diff
@@ -0,0 +1,142 @@
+package http
```
j4k4, line 1: Do we need an extra package for this? Maybe move it up into the prometheus package and name the factory `NewMetricsServer`? :thinking:

pkg/metric/prometheus/registry.go
```diff
@@ -0,0 +1,27 @@
...
+import (
+	"context"
+	"sync"
+
+	"github.com/justtrackio/gosoline/pkg/appctx"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/collectors"
+)
+
+var registryLock sync.Mutex
```
j4k4, line 12: I can see why one would need a lock here but ideally this would be handled by the appctx.Provide function, right? I was a little bit surprised that that one has no locking yet..  
If you feel like it, you could try to add a locking mechanism there which locks based on the key? :thinking:

j4k4:
> There is no benefit of using the metric daemon though because there are no delays caused by pushing metrics somewhere.

I'm wondering if we would want to change the writer config.. Do one really needs multiple metric writers in one app? Does it make sense to write to multiple metric systems at once? :thinking:   
Maybe we should change it to allow one metric writer only. With that we could enable the daemon based on which writer is selected.. eg cw uses the daemon and prom does not..

pkg/metric/writer_prometheus.go
```diff
@@ -0,0 +1,221 @@
...
+
+type promSettings struct {
+	// MetricLimit is used to avoid having metrics for which the name is programmatically generated (or have large number
+	// of possible dimensions) which could lead in a memory leak.
+	MetricLimit int64 `cfg:"metric_limit" default:"10000"`
+}
+
+type promDatum struct {
+	*Datum
+	namespace string
```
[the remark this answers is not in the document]
netsi: makes sense, removed it

pkg/metric/writer_prometheus.go
netsi: added the application too, but now anyone could have it's own namespace naming strategy for prometheus by using the `metric.WithPromNSNamingStrategy()` func

pkg/metric/prometheus/http/metric_server.go
netsi: moved it up to prometheus, it was causing import cycle because the metric_server is using the kernel pkg

pkg/metric/prometheus/registry.go
netsi: makes sense, could solve potential future issues, in my case because I want to access the registry almost at the same time I had issues. Moved the lock to appctx. 

I had one issue though, I wanted to create a KeyLock in `conc` pkg, but because that pkg has dependency on `ddb` which uses the `metric` pkg I had an import cycle, I would like probably to move all the ddb related locks inside a sub pkg, but that would cause breaking changes, what's your opinion? currently this lock implementation is in appctx

netsi:
@j4k4 regarding the change to a single writer, maybe we could do that in a second step? I would like to avoid introducing breaking changes atm. 

But to answer your question the only two reasons I can think of having two metric writers at the same time is during a migration from one system to another or during an experimentation with a new metric solution. I don't think though that is a reason to support multiple writers as it is not a really common case.

pkg/application/options.go
```diff
@@ -271,6 +272,10 @@ func WithProfiling() Option {
 	return WithModule("profiling", apiserver.NewProfiling())
 }
 
+func WithPrometheusMetricServer(app *App) {
```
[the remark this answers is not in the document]
netsi: Yeah it seems a bit inconvenient to have to specify it, I'll try to come up with something

pkg/appctx/container.go
```diff
@@ -24,6 +27,18 @@ func (e ErrNoItemFound) Error() string {
 	return fmt.Sprintf("no item with key %v found", e.Key)
 }
 
+type KeyLock struct {
+	locks sync.Map
+}
+
+func (l *KeyLock) Lock(key any) *sync.Mutex {
+	lockItem, _ := l.locks.LoadOrStore(key, &sync.Mutex{})
```
j4k4, line 35: We shouldn't ignore errors, even if its very unlikely or impossible to get them. Let's just handle them to write clean code. :+1:

pkg/appctx/container.go
```diff
@@ -24,6 +27,18 @@ func (e ErrNoItemFound) Error() string {
 	return fmt.Sprintf("no item with key %v found", e.Key)
 }
 
+type KeyLock struct {
```
j4k4, line 30: I'm wondering if it would be possible to delete the mutex on unlocking to free up the memory again?

pkg/metric/prometheus/registry.go
j4k4: We can keep the lock implementation here for now..  
Maybe just move it into a dedicated file to make it more prominent?  
I'm not a big fan of mixing to many "complex" types in a single file..

j4k4:
> @j4k4 regarding the change to a single writer, maybe we could do that in a second step? I would like to avoid introducing breaking changes atm.

Yes, for the sake of moving forward, we can keep it as is for now..  
But i would like to see this change somewhat in the future even if it is a bigger breaking change again..

pkg/appctx/container.go
netsi: LoadOrStore returns the item and a boolean flag if it was actually loaded (aka cash-hit).  I don't think we care at this point if the item was retrieved from the cache or not, but maybe if we also want to delete the mutex as you suggested after unlocking we could use this flag.

pkg/metric/prometheus/registry.go
netsi: I actually want it to be in `conc` pkg but to do so I'll have to move the locks that have the `ddb` dependency somewhere else, what do you think?

j4k4: Where do you would put it?

pkg/appctx/container.go
```diff
@@ -4,11 +4,16 @@ import (
...
+
+	"github.com/justtrackio/gosoline/pkg/conc"
 )
 
 type containerKeyType int
 
-var containerKey containerKeyType = 1
+var (
+	containerKey containerKeyType = 1
+	lock                          = conc.NewKeyLock()
```
j4k4, line 15: The lock shouldn't be global but bound to the container, right? :thinking:

netsi: hmm the container itself uses sync.Map so that should not be a problem. But I tried removing the lock completely and I can't seem to reproduce the race condition I had in the past.

netsi: but you are right there seem to not be a reason to have a global lock at this place

netsi: ok scratch the comment before :D , so the issue does not occur anymore because of the changes I did to the metric factory.  So it doesn't add 2 modules 1 for Prometheus API server and one for the Metric daemon. But it could be in some edge cases that we'll need the lock, so I'll add it to the container

j4k4: :+1:

j4k4 (approved):

PR #891, opened 2022-04-29
log: implemented sentry settings via config;

bogdanfinn:


j4k4:
This is a complicated one..  
Currently, the sentry config would be on root level, right?  As  there could be in theory multiple sentry handlers added to the logger, so the sentry config should be on handler level like it is already with the iowriter handler. Now, the difficulty is, that the sentry handler isn't added via config but instead as an application option. That's messed up and has to be changed to have all the logger handlers handled all the same. We would have to set this then via env variables for the prod deployments but this isn't working yet, as for this the config pkg needs some changes I would like to do for quite some time already.. 

What's the requirement behind this? We could merge it like this, if it's important and urgent, but it will 100% not stay like this, once there is time to do it in a proper way..

applike-ss:
Why would one need more than one sentry configuration in 1 app? I don't see why an application should send 1 error event to multiple sentries or why it should devide them e.g. by log level/handler. Isn't that over-engineered to have this capability to split by log handler?

j4k4:
Yes, it doesn't make sense to have more than 1 sentry handler, but we're using sentry as a log handler, so it should be configured as a handler. For me, it's more important here, that all of them are following the same pattern.  

By the way, it's the same in java with log4j.. Sentry is used here as an appender and configured on appender level. You could also add multiple sentry appenders. Nobody would do it but it's about following the pattern.
https://docs.sentry.io/platforms/java/guides/log4j2/#configure

In our case it's just named handler instead of appender.. Everything else is basically the same,,

PR #892, opened 2022-05-02
db_repo: added too many results error;

bogdanfinn:


pkg/db-repo/errors.go
```diff
@@ -54,3 +54,27 @@ func (e NoQueryResultsError) Unwrap() error {
...
+
+type TooManyResultsError struct {
+	modelId string
+	err     error
+}
+
+func NewTooManyResultsError(found int, expected int, modelId string) TooManyResultsError {
+	return TooManyResultsError{
+		modelId: modelId,
+		err:     fmt.Errorf("found %d result(s), expected %d result(s)", found, expected),
```
selm0, line 66: What do you need this additional error for? Isn't it sufficient to store `found` and `expected`? The Unwrap also doesn't make much sense, as you'll never be able to assert for this specific error as it's untyped

selm0 (changes requested):

selm0 (approved):

PR #894, opened 2022-05-04
moved dispatcher into appctx

nikplxjt:
Currently we can not use listeners properly in our integration tests. For each test apiserver test case, the callbacks of the dispatcher are added multiple times. I moved the dispatcher into the appctx so that it can be retrieved via a `Provide...` function and added a `Reset` method which can be used to reset the callbacks. This is now done by the testsuite together with the cleanup of the sqs queues and kvstores after each test case. 

Also I exposed the `Get` method of the ladon warden on the guard, because we need it in some other repository now.


pkg/dispatcher/dispatcher.go
```diff
@@ -1,41 +1,80 @@
...
+type dispatcher struct {
+	logger    log.Logger
+	mx        sync.RWMutex
+	listeners map[string][]Callback
+}
 
-var listeners = map[string][]Callback{}
-var d = &dispatcher{}
+func ProvideDispatcher(ctx context.Context, config cfg.Config, logger log.Logger) Dispatcher {
+	dispatcher, _ := appctx.Provide(ctx, dispatcherCtxKey("DispatcherRepository"), func() (Dispatcher, error) {
```
bogdanfinn, line 31: dispatcherCtxKey("Dispatcher")

pkg/guard/guard.go
```diff
@@ -12,12 +12,13 @@ const fetchLimit = 100
 
 //go:generate mockery --name Guard
 type Guard interface {
-	IsAllowed(request *ladon.Request) error
+	CreatePolicy(pol ladon.Policy) error
+	DeletePolicy(pol ladon.Policy) error
+	Get(id string) (ladon.Policy, error)
```
bogdanfinn, line 17: just to make it consistent maybe `GetPolicy()` ?

pkg/test/suite/run.go
```diff
@@ -156,6 +157,9 @@ func runTestCaseWithSharedEnvironment(t *testing.T, suite TestingSuite, suiteOpt
 		stream.ResetInMemoryOutputs()
 		stream.ResetProducerDaemons()
 		kvstore.ResetConfigurableKvStores()
+
+		disp := dispatcher.ProvideDispatcher(suite.Env().Context(), suite.Env().Config(), suite.Env().Logger())
```
j4k4, line 161: Why do we need the reset here? Shared environment doesn't mean shared application, so every testcase should be executed in it's own application.. Since the dispatcher is scoped to the application now, there shouldn't be a need to reset it, right?

pkg/guard/guard.go
j4k4: Agree

pkg/dispatcher/dispatcher.go
j4k4, line 31: We shouldn't ignore errors and instead handle them in a proper way.

pkg/test/suite/run.go
nikplxjt: Actually if we use that together with the `WithSharedEnvironment` option, the suites env is not recreated which means that the dispatcher is still in the context. So when we run an `ApiServerTestCase` we run the `SetupApiDefinitions` function  on each test case which will run all the `AddListener` functions which will add multiple listeners to the dispatcher.

j4k4: Mh, okay.. so the api definitions get created with the env context and not the app context? Or is it the same for the integrations test.. Sry, haven't looked at the code for some time.. :grimacing:   
Each app should have it's own context.. If that's not the case, we have to fix that issue..

pkg/dispatcher/repository_db.go
```diff
@@ -15,14 +15,17 @@ type Repository struct {
 	logger     log.Logger
 }
 
-func NewRepository(config cfg.Config, logger log.Logger, repo db_repo.Repository) db_repo.Repository {
-	disp := Get()
+func NewRepository(ctx context.Context, config cfg.Config, logger log.Logger, repo db_repo.Repository) (db_repo.Repository, error) {
+	disp, err := ProvideDispatcher(ctx, config, logger)
+	if err != nil {
+		return nil, err
```
j4k4, line 21: We should wrap the error here to add some context like "can not get dispatcher" or something like that..

pkg/test/suite/run.go
```diff
@@ -2,6 +2,7 @@ package suite
 
 import (
 	"fmt"
+	"github.com/justtrackio/gosoline/pkg/dispatcher"
```
j4k4, line 5: This import line should go down to the others.. :)

pkg/dispatcher/repository_db.go
nikplxjt: done

pkg/test/suite/run.go
nikplxjt: done

pkg/test/suite/run.go
nikplxjt: I added a new method `ResetContext` on the test env, an made it just switch the context with a fresh one. As far as I can see, all test cases still run with no problem, so I removed the `Reset` method on the dispatcher again. Thanks for pointing it out.

nikplxjt: After some discussions with @selm0 , I made the testcases just always inject a new context in each test. This also passes all test cases.

selm0 (approved):

PR #895, opened 2022-05-11
Entity sharing package

nikplxjt:
In the future  we want to have the capability to share certain entities with users that do not have full read/write access to parts of the used software. With "sharing" we mean having users with restricted access to specific entities, enforced by guard policies. This PR adds the `share` package which provides a `crud`-like HTTP handler as well as a kernel option and a wrapper for a `db_repo.Repository`. It also defines a `Share` entity, which consists of:
* The ID of the targeted entity
* The ID of the owning entity (i.e. the user)
* The ID of the guard policy (given the SqlWarden is used)

When using this new package to create entity shares there one wants to do the following things:
1. Add the new kernel option the the target applications options - this will create a table for the new shares (similar to the change history option)
2. Configure the table names of the share feature in the application config
3. Create a new  create-handler for shares to a specific entity (it will add a `/share` path to the entity path) and add it in the applications `apiserver.Definer`
4. Wrap the repository of the entity with the new share repository wrapper (so that the shares and policies of an entity will be updated/deleted if the entity itself is updates/deleted), you will need to provide a repository that can access the table of the shares
5. Implement the `Shareable` interface for the entity one wants to share
6. (optional) Implement a list and delete handler for the shares in the new table

### DIFF
--- a/pkg/application/options.go
+++ b/pkg/application/options.go
@@ -7,6 +7,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/justtrackio/gosoline/pkg/share"
+
 	"github.com/justtrackio/gosoline/pkg/apiserver"
 	"github.com/justtrackio/gosoline/pkg/cfg"
 	"github.com/justtrackio/gosoline/pkg/clock"
@@ -142,6 +144,12 @@ func WithExecBackoffSettings(settings *exec.BackoffSettings) Option {
 func WithDbRepoChangeHistory(app *App) {
 	app.addKernelOption(func(config cfg.GosoConf) kernelPkg.Option {
 		return kernelPkg.WithMiddlewareFactory(db_repo.KernelMiddlewareChangeHistory, kernelPkg.PositionEnd)
+	})
+}
+
+func WithApiServerShares(app *App) {
+	app.addKernelOption(func(config cfg.GosoConf) kernelPkg.Option {
+		return kernelPkg.WithMiddlewareFactory(share.KernelMiddlewareShares, kernelPkg.PositionEnd)
 	})
 }
 
--- a/pkg/share/manager.go
+++ b/pkg/share/manager.go
@@ -1,0 +1,92 @@
+package share
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/jinzhu/gorm"
+	"github.com/justtrackio/gosoline/pkg/appctx"
+	"github.com/justtrackio/gosoline/pkg/cfg"
+	"github.com/justtrackio/gosoline/pkg/db-repo"
+	"github.com/justtrackio/gosoline/pkg/log"
+)
+
+type Settings struct {
+	TableName struct {
+		Owner  string `cfg:"owner" default:"owners"`
+		Policy string `cfg:"policy" default:"policies"`
+		Share  string `cfg:"share" default:"shares"`
+	} `cfg:"table_name"`
+}
+
+type shareManagerAppctxKey string
+
+type Share struct {
+	db_repo.Model
+	EntityId   uint
+	EntityType string
+	OwnerId    uint
+	PolicyId   string
+}
+
+func (s *Share) GetPolicyId() string {
+	return s.PolicyId
+}
+
+type shareManager struct {
+	orm      *gorm.DB
+	logger   log.Logger
+	settings Settings
+}
+
+func ProvideShareManager(ctx context.Context, config cfg.Config, logger log.Logger) (*shareManager, error) {
+	return appctx.Provide(ctx, shareManagerAppctxKey("shareManager"), func() (*shareManager, error) {
+		return NewShareManager(config, logger)
+	})
+}
+
+func NewShareManager(config cfg.Config, logger log.Logger) (*shareManager, error) {
+	orm, err := db_repo.NewOrm(config, logger)
+	if err != nil {
+		return nil, fmt.Errorf("can not create orm: %w", err)
+	}
+
+	settings := Settings{}
+	config.UnmarshalKey("share", &settings)
+
+	return &shareManager{
+		logger:   logger.WithChannel("share_manager"),
+		orm:      orm,
+		settings: settings,
+	}, nil
+}
+
+func (m *shareManager) SetupShareTable() error {
+	tn := m.settings.TableName
+
+	scope := m.orm.NewScope(Share{})
+	if scope.Dialect().HasTable(tn.Share) {
+		return nil
+	}
+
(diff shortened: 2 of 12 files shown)


pkg/share/manager.go
```diff
@@ -0,0 +1,92 @@
+package share
+
+import (
+	"context"
+	"fmt"
+	"github.com/jinzhu/gorm"
+	"github.com/justtrackio/gosoline/pkg/appctx"
+	"github.com/justtrackio/gosoline/pkg/cfg"
+	db_repo "github.com/justtrackio/gosoline/pkg/db-repo"
```
applike-ss, line 9:
```suggestion
	"github.com/justtrackio/gosoline/pkg/db-repo"
```

pkg/share/manager.go
```diff
@@ -0,0 +1,92 @@
...
+	"github.com/justtrackio/gosoline/pkg/log"
+)
+
+type Settings struct {
+	TableNames struct {
+		Share  string `cfg:"share" default:"shares"`
+		Owner  string `cfg:"owner" default:"owners"`
+		Policy string `cfg:"policy" default:"policies"`
```
applike-ss, line 17:
```suggestion
		Share  string `cfg:"share" default:"share"`
		Owner  string `cfg:"owner" default:"owner"`
		Policy string `cfg:"policy" default:"policy"`
```

pkg/share/manager.go
```diff
@@ -0,0 +1,92 @@
...
+	PolicyId   string
+}
+
+func (s *Share) GetPolicyId() string {
+	return s.PolicyId
+}
+
+type shareManager struct {
+	orm      *gorm.DB
+	logger   log.Logger
```
applike-ss, line 37: please ensure our code style

pkg/share/manager.go
```diff
@@ -0,0 +1,92 @@
...
+			entity_id        int unsigned not null,
+			entity_type		 varchar(255) not null,
+			owner_id       	 int unsigned not null,
+			policy_id        varchar(255) not null,
+			created_at       timestamp    not null,
+			updated_at       timestamp    not null,
+
+			constraint shares_owner_id_fk foreign key (owner_id) references %s (id),
+			constraint shares_policy_id_fk foreign key (policy_id) references %s (id)
+		)
```
applike-ss, line 84: should we make this a const to not bloat the function?

pkg/share/options.go
```diff
@@ -0,0 +1,25 @@
...
+		}
+
+		if err := manager.SetupShareTable(); err != nil {
+			logger.Error(err.Error())
+			return
+		}
+
+		next()
+	}
+}
```
applike-ss, line 25: I'm not up-to-date RN with gosoline and middlewares. Is returning in case of an error the right thing to do or should the next middleware be executed?

pkg/share/manager.go
```diff
@@ -0,0 +1,92 @@
...
+	"fmt"
+	"github.com/jinzhu/gorm"
+	"github.com/justtrackio/gosoline/pkg/appctx"
+	"github.com/justtrackio/gosoline/pkg/cfg"
+	db_repo "github.com/justtrackio/gosoline/pkg/db-repo"
+	"github.com/justtrackio/gosoline/pkg/log"
+)
+
+type Settings struct {
+	TableNames struct {
```
applike-ss, line 14:
```suggestion
	TableName struct {
```

pkg/share/manager.go
```diff
@@ -0,0 +1,92 @@
...
+		Share  string `cfg:"share" default:"shares"`
+		Owner  string `cfg:"owner" default:"owners"`
+		Policy string `cfg:"policy" default:"policies"`
+	} `cfg:"table_names"`
```
applike-ss, line 18:
```suggestion
	} `cfg:"table_name"`
```

pkg/share/shareable_repository.go
```diff
@@ -0,0 +1,109 @@
...
+	"github.com/justtrackio/gosoline/pkg/log"
+	"github.com/justtrackio/gosoline/pkg/mdl"
+	"github.com/ory/ladon"
+)
+
+type shareRepository struct {
+	db_repo.Repository
+	shareRepository db_repo.Repository
+	guard           guard.Guard
+	logger          log.Logger
```
applike-ss, line 18: is the struct order correct?

pkg/test/suite/run.go
```diff
@@ -2,6 +2,7 @@ package suite
 
 import (
 	"fmt"
+	"github.com/justtrackio/gosoline/pkg/dispatcher"
```
applike-ss, line 5: shouldn't internal and external packages be grouped by that?

pkg/test/suite/run.go
```diff
@@ -156,6 +157,13 @@ func runTestCaseWithSharedEnvironment(t *testing.T, suite TestingSuite, suiteOpt
 		stream.ResetInMemoryOutputs()
 		stream.ResetProducerDaemons()
 		kvstore.ResetConfigurableKvStores()
+
+		disp, err := dispatcher.ProvideDispatcher(suite.Env().Context(), suite.Env().Config(), suite.Env().Logger())
```
applike-ss, line 161: maybe extract `suite.Env()` into its own variable if you call it multiple times?

pkg/db-repo/errors.go
```diff
@@ -54,3 +54,23 @@ func (e NoQueryResultsError) Unwrap() error {
...
+type TooManyResultsError struct {
+	modelId  string
+	found    int
+	expected int
+}
+
+func NewTooManyResultsError(found int, expected int, modelId string) TooManyResultsError {
+	return TooManyResultsError{
+		modelId: modelId,
+	}
```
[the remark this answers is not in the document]
nikplxjt: @bogdanfinn

pkg/application/options.go
```diff
@@ -148,6 +149,13 @@ func WithDbRepoChangeHistory(app *App) {
 	})
 }
 
+func WithApiServerShares(app *App) {
+	app.addKernelOption(func(config cfg.GosoConf, kernel kernelPkg.GosoKernel) error {
```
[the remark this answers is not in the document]
nikplxjt: hmmm Goland does not highlight  this ... probably the usage detection does not work for closures

pkg/db-repo/errors.go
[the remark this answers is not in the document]
bogdanfinn: resolved in my branch

selm0 (approved):

PR #901, opened 2022-05-24
Add functional helpers package

nikplxjt:
As Go 1.18 introduced generics, it is now possible to write some functional code in a typesafe manner. This can replace packages like `thoas/go-funk`, which uses reflection pretty heavily to implement functions that work for a given subset of types. Using reflection has a pretty heavy performance impact, which can be mitigated by using generics. The other non-generic approach of monomorphizing "by hand" (i.e. writing a function for each type) on the other hand causes heavy code duplication, while allowing for optimal performance. Using generics is a good middle ground here.

This PR implements a minimal set of useful functional helpers and benchmarks them against `thoas/go-funk`. The set of functions is kept minimal, because of the possibility that some third party library solves them even more efficiently in the future (`go-funk` does not yet have generics in place).

### DIFF
--- a/docs/fixtures/readme.md
+++ b/docs/fixtures/readme.md
@@ -72,9 +72,9 @@ func createFixtures() []*fixtures.FixtureSet {
 			Fixtures: []interface{}{
 				&OrmFixtureExample{
 					Model: db_repo.Model{
-						Id: mdl.Uint(1),
+						Id: mdl.Box(uint(1)),
 					},
-					Name: mdl.String("example"),
+					Name: mdl.Box("example"),
 				},
 			},
 		},
--- a/docs/more_details/integration_tests.md
+++ b/docs/more_details/integration_tests.md
@@ -32,8 +32,8 @@ func (s *ConsumerTestSuite) TestSuccessTwice(app suite.AppUnderTest) {
 	consumer := s.Env().StreamInput("consumerInput")
 	s.NotNil(consumer)
 
-	consumer.Publish(mdl.Uint(2), nil)
-	consumer.Publish(mdl.Uint(3), nil)
+	consumer.Publish(mdl.Box(uint(2)), nil)
+	consumer.Publish(mdl.Box(uint(3)), nil)
 
 	app.Stop()
 	app.WaitDone()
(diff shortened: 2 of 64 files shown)


selm0 (approved):

nikplxjt:
### Implemented funk package

This change introduces a new package `funk` which works similar to `github.com/thoas/go-funk` which is used throughout gosoline. As Go introduces generics with v1.18 and `go-funk` did not provide any generic versions of their functional helpers yet, we decided to implement our own version.

In this package you can find generic versions of often used functional helpers for slices like `Filter`, `Map`, `Reduce` and `Filter`. It also provides some helpful data structures like a generic `Set` type and the means to easily convert between data structures like `SliceToSet` or `SliceToMap`.

The package also includes benchmark tests (you can find them in the package README). Performance-wise the functions perform constantly better then `go-funk`, because it does not need to rely on reflection. If you encounter problems with highlighting, please make sure that you have the latest version of GoLand installed, or if the problems persist please install the GoLand Early Access Program version and see if the highlighting problems go away.

Please note that `funk` does not provide map and slice helpers like `map.Values`, `map.Keys` or `slice.Contains` as they are now already generically implemented withing `golang.org/x/exp/slices` and `golang.org/x/exp/maps`. Please use these packages, when possible.

Also be aware that functions like `funk.Uniq` and `slices.Contains` can work slightly different from their `go-funk` counterparts, as the generic versions use the `comparable` constraint and the `==` operator to compare elements, which will do a shallow equality check, while `go-funk` usually uses `reflect.DeepEqual`. Usually doing a `DeepEqual` delivers worse performance and is not needed, therefore if you really need it, please implement your own function or use some helper with a customizable predicate like `funk.FindFirstFunc`.

### BREAKING: made mdl helpers generic

The second change in this version is the removal of all explicitely typed functions in the `mdl` package (i.e. `mdl.Uint`, `mdl.EmptyStringIfNil` ...) and their replacement with `mdl.Box` and `mdl.EmptyIfNil`. 

Please note that `mdl.Box` is a generic function that take an object and returns a pointer to it. So `mdl.Int(6)` becomes `mdl.Box(6)`. The type will be inferred in most cases. Only for instanciating types like `uint` which is initialized like another default type (`6` in Go will assumed to be `int` by default) you will need to use a typecast like this `mdl.Box(uint(6))`. 

**If you upgrade to this version, you will need to replace most of your calls to the mdl package**

PR #904, opened 2022-06-10
kernel refactorings

j4k4:
With this change we're splitting the creation and execution of a kernel into multiple parts.

This introduces mainly two advantages over the old approach:
    - once the kernel is created, the kernel gets immutable, 
      which reduces some of the complexity which was caused due to a mutable kernel
    - having blueprints of the kernel and a factory to build everything
      enables us to build the modules and middlewares without running the kernel
      which makes it easier to build the the same kernel with different configurations
    
The kernel package provides 3 main structures now:
    - The Blueprint contains the building parts of a kernel
    - The Factory is used to build a immutable Kernel based on a Blueprint and a provided configuration
    - The final Kernel can now be run and stopped to control the exection of an application


pkg/cli/cli.go
```diff
@@ -35,16 +35,21 @@ func Run(module kernel.ModuleFactory, otherModuleMaps ...map[string]kernel.Modul
...
 	for _, otherModuleMap := range otherModuleMaps {
 		for name, otherModule := range otherModuleMap {
-			k.Add(name, otherModule)
+			options = append(options, kernel.WithModuleFactory(name, otherModule))
 		}
 	}
+
+	k, err := kernel.BuildKernel(ctx, config, logger, options)
+	if err != nil {
+		defaultErrorHandler("can not initialize the kernel: %w", err)
```
ajscholl, line 51: We don't necessarily know that the default error handler doesn't return, so we should return afterwards to avoid running a nil kernel

pkg/kernel/builder.go
```diff
@@ -0,0 +1,32 @@
...
+	"github.com/justtrackio/gosoline/pkg/log"
+)
+
+func BuildFactory(ctx context.Context, config cfg.Config, logger log.Logger, options []Option) (*factory, error) {
+	blueprint := NewBlueprint(options...)
+
+	var err error
+	var factory *factory
+
+	if factory, err = NewFactory(ctx, config, logger, blueprint); err != nil {
```
ajscholl, line 16: Is `NewFactory` returning a non-nil factory if it returns an error? If not, we can directly return the result from it, right?

pkg/kernel/factory.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+
+	if err := factory.build(); err != nil {
+		return nil, fmt.Errorf("can not build kernel factory: %w", err)
+	}
+
+	return factory, nil
+}
+
+func (f *factory) GetKernel() (Kernel, error) {
```
ajscholl, line 40: This function never returns an error, why does the caller have to check for one?

pkg/kernel/factory.go
```diff
@@ -0,0 +1,173 @@
...
+	}
+
+	if !f.stages.hasModules() {
+		err = fmt.Errorf("no modules to run")
+		return
+	}
+
+	if f.stages.countForegroundModules() == 0 {
+		err = fmt.Errorf("no foreground modules to run")
+		return
```
ajscholl, line 84: To be honest, I don't really like the implicit nature of these returns - why not `return err`, `return fmt.Errorf(...)`, or `return nil`? This way it is a lot easier to see that at that point I am not returning any error for example...

pkg/kernel/factory_test.go
```diff
@@ -0,0 +1,83 @@
...
+	})
+	assert.EqualError(t, err, "can not build kernel factory: no foreground modules to run")
+}
+
+func TestFactoryModuleMultiFactoryError(t *testing.T) {
+	config, logger, _ := createMocks()
+	factoryErr := fmt.Errorf("error in module factory")
+	logger.On("Error", "error building additional modules by multiFactories: %w", factoryErr)
+
+	assert.NotPanics(t, func() {
```
ajscholl, line 46: Is there a difference between `assert.NotPanics` and just running the test? I always thought that a panicking test would also fail

pkg/kernel/kernel_test.go
```diff
@@ -516,79 +465,79 @@ func (m *realModule) Run(ctx context.Context) error {
...
-			k.Add("exist-slow", func(ctx context.Context, config cfg.Config, logger log.Logger) (kernel.Module, error) {
-				return &slowExitModule{
-					kernel: k,
-				}, nil
-			}, kernel.ModuleStage(s))
-		}
-		k.Run()
-	})
-}
+//func TestKernel_RunRealModule(t *testing.T) {
```
ajscholl, line 468: What is with this test?

pkg/cli/cli.go
j4k4: added the returns :+1:

pkg/kernel/factory.go
j4k4: changed

pkg/kernel/factory_test.go
j4k4: Not sure, I've just copied that from the old tests.. but yes, it should work without it.. Will remove them..

pkg/kernel/factory.go
j4k4: changed

pkg/kernel/kernel_test.go
j4k4: must have missed it.. added again..

ajscholl (approved):

PR #905, opened 2022-06-16
mdl: added NilIfEmpty function;

SamuelTheJackson:
Some apis returning empty values instead of null for optional values. With this function we can set the values on our end to nil if they are not set.


selm0 (approved):

SamuelTheJackson:
Will be added to https://github.com/justtrackio/gosoline/pull/901

PR #908, opened 2022-06-28
apiserver: record the size of the compressed and uncompressed request as log fields;

ajscholl:


docs/getting_started/how_to_build_an_api.md
```diff
@@ -65,7 +65,7 @@ TimeoutSettings configures IO timeouts.
 [structmd]:# (pkg/apiserver/compression.go CompressionSettings)
 **CompressionSettings**
 
-CompressionSettings allow the enabling of gzip support for requests and responses. By default compressed requests are accepted, and compressed responses are returned (if suitable).
+CompressionSettings control gzip support for requests and responses. By default, compressed requests are accepted, and compressed responses are returned (if accepted by the client).
```
applike-ss, line 68:
```suggestion
CompressionSettings control gzip support for requests and responses. By default compressed requests are accepted and compressed responses are returned (if accepted by the client).
```

pkg/apiserver/compression.go
```diff
@@ -1,15 +1,15 @@
...
+	"strconv"
+
 	ginGzip "github.com/gin-contrib/gzip"
 	"github.com/gin-gonic/gin"
-	"io"
-	"strconv"
 )
 
-// CompressionSettings allow the enabling of gzip support for requests and responses. By default compressed requests are accepted, and compressed responses are returned (if suitable).
+// CompressionSettings control gzip support for requests and responses. By default, compressed requests are accepted, and compressed responses are returned (if accepted by the client).
```
applike-ss, line 12: same as above

pkg/apiserver/compression_logs.go
```diff
@@ -0,0 +1,50 @@
...
+			result["request_compression"] = data.contentEncoding
+			result["request_uncompressed_size"] = *data.size
+		}
+	}
+
+	responseSizeFields, found := ginCtx.Get(responseSizeFields)
+	if found && responseSizeFields != nil {
+		if data, ok := responseSizeFields.(encodedSizeData); ok {
+			result["response_compression"] = data.contentEncoding
+			result["uncompressed_bytes"] = *data.size
```
applike-ss, line 45:
```suggestion
			result["response_uncompressed_size"] = *data.size
```

applike-ss (changes requested):

pkg/apiserver/compression_logs.go
ajscholl: Hmm, we already have a field `bytes` which records the raw size (i.e., compressed if compression was applied) of a response, should I also rename that field then? I would like to keep the similarly named.  `gosoline/pkg/apiserver/middleware_logger.go:40`

applike-ss: i see your point of keeping things named similarly. Since bytes is more speaking than size, i would actually prefer to have it say "bytes" everywhere. Could you adjust that?

applike-ss (approved):

PR #642, opened 2021-12-14
parquet: updated parquet-go dependency to v1.6.2 for not needing CGO_…

applike-ss:
…ENABLED=1 anymore;

PR #924, opened 2022-08-10
bumped go-parquet module version

nikplxjt:
Currently we use a very old version of `parquet-go` which has race conditions in its session handling and also does not use go modules yet (I can create a local development state that completely breaks local development with gosoline). I bumped the version and corrected the parquet reader and writer to be in line with the things that changed between versions.


pkg/parquet/reader.go
```diff
@@ -51,8 +51,6 @@ type s3Reader struct {
 }
 
 func NewReader(ctx context.Context, config cfg.Config, logger log.Logger, settings *ReaderSettings) (*s3Reader, error) {
-	s3Cfg := gosoS3.GetLegacyConfig(config, "default")
```
selm0, line 54: Can we remove the legacy config now? My Goland can't find any further usages

selm0 (approved):

pkg/parquet/reader.go
nikplxjt: good point 

done

nikplxjt:
closes https://github.com/justtrackio/gosoline/pull/642

PR #930, opened 2022-08-19
apiserver: use list of audiences;

selm0:
To support multiple calling projects, we want to support several google projects as a login source.


bogdanfinn:
+1

PR #943, opened 2022-09-19
apiserver: don't build a nice object with query parameters for failed requests

ajscholl:
If we get hit by a crawler, we often get unexpected query parameters. To avoid generating too many log fields in Kibana, we only add the query parameters as a raw object on successful requests to the log message.

### DIFF
--- a/pkg/apiserver/middleware_logger.go
+++ b/pkg/apiserver/middleware_logger.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"net/http"
 	"strings"
 	"time"
 
@@ -26,40 +27,47 @@ func LoggingMiddleware(logger log.Logger) gin.HandlerFunc {
 		pathRaw := getPathRaw(ginCtx)
 
 		referer := req.Referer()
+		userAgent := req.UserAgent()
+		status := ginCtx.Writer.Status()
 
-		query := req.URL.Query()
 		queryRaw := req.URL.RawQuery
-		queryParameters := make(map[string]string)
-
-		for k := range query {
-			queryParameters[k] = query.Get(k)
-		}
 
 		method := ginCtx.Request.Method
 		requestTimeNano := time.Since(start)
 		requestTimeSecond := float64(requestTimeNano) / float64(time.Second)
 
-		ctxLogger := chLogger.WithContext(ctx).WithFields(log.Fields{
-			"bytes":                    ginCtx.Writer.Size(),
-			"client_ip":                ginCtx.ClientIP(),
-			"host":                     req.Host,
-			"protocol":                 req.Proto,
-			"request_method":           method,
-			"request_path":             path,
-			"request_path_raw":         pathRaw,
-			"request_query":            queryRaw,
-			"request_query_parameters": queryParameters,
-			"request_referer":          referer,
-			"request_time":             requestTimeSecond,
-			"scheme":                   req.URL.Scheme,
-			"status":                   ginCtx.Writer.Status(),
-		})
+		fields := log.Fields{}
+		fields["bytes"] = ginCtx.Writer.Size()
+		fields["client_ip"] = ginCtx.ClientIP()
+		fields["host"] = req.Host
+		fields["protocol"] = req.Proto
+		fields["request_method"] = method
+		fields["request_path"] = path
+		fields["request_path_raw"] = pathRaw
+		fields["request_query"] = queryRaw
+		fields["request_referer"] = referer
+		fields["request_user_agent"] = userAgent
+		fields["request_time"] = requestTimeSecond
+		fields["scheme"] = req.URL.Scheme
+		fields["status"] = status
+
+		// only log query parameters in full for successful requests to avoid logging them from bad crawlers
+		if status != http.StatusUnauthorized && status != http.StatusForbidden && status != http.StatusNotFound {
+			queryParameters := make(map[string]string)
+			query := req.URL.Query()
+
+			for k := range query {
+				queryParameters[k] = query.Get(k)
+			}
+
+			fields["request_query_parameters"] = queryParameters
+		}
 
 		if requestId := req.Header.Get("X-Request-Id"); requestId != "" {
-			ctxLogger = ctxLogger.WithFields(log.Fields{
-				"request_id": requestId,
-			})
+			fields["request_id"] = requestId
 		}
+
+		ctxLogger := chLogger.WithContext(ctx).WithFields(fields)
 
 		if len(ginCtx.Errors) == 0 {
 			ctxLogger.Info("%s %s %s", method, path, req.Proto)


docs/getting_started/how_to_build_an_api.md
```diff
@@ -65,7 +65,7 @@ TimeoutSettings configures IO timeouts.
 [structmd]:# (pkg/apiserver/compression.go CompressionSettings)
 **CompressionSettings**
 
-CompressionSettings allow the enabling of gzip support for requests and responses. By default compressed requests are accepted, and compressed responses are returned (if suitable).
+CompressionSettings allow the enabling of gzip support for requests and responses. By default, compressed requests are accepted, and compressed responses are returned (if suitable).
```
applike-ss, line 68:
```suggestion
CompressionSettings allow to enable gzip support for requests and responses. By default compressed requests are accepted and compressed responses are returned (if suitable).
```

docs/getting_started/module.md
```diff
@@ -224,33 +224,6 @@ Having this order is useful. For example, the metric daemon can collect metrics
...
-**ModuleConfig**
-
-ModuleConfig stores attributes used in starting and stopping a module.
-
-| field       | type     | default     | description     |
-| :------------- | :----------: | :----------: | -----------: |
-| Essential | bool |  | Essential causes the kernel to stop after its first essential module finishes. |
-| Background | bool |  | Background modules run as long as there is at least one foreground module still running. |
-| Stage | int |  | Stage in which this module will be started. |
-
```
applike-ss, line 253: why is this being removed?

pkg/apiserver/compression.go
```diff
@@ -1,15 +1,16 @@
...
+	"strconv"
+
 	ginGzip "github.com/gin-contrib/gzip"
 	"github.com/gin-gonic/gin"
-	"io"
-	"strconv"
 )
 
-// CompressionSettings allow the enabling of gzip support for requests and responses. By default compressed requests are accepted, and compressed responses are returned (if suitable).
+// CompressionSettings allow the enabling of gzip support for requests and responses. By default, compressed requests are accepted, and compressed responses are returned (if suitable).
```
applike-ss, line 13:
```suggestion
// CompressionSettings allow to enable gzip support for requests and responses. By default compressed requests are accepted and compressed responses are returned (if suitable).
```

pkg/apiserver/compression.go
```diff
@@ -54,58 +55,78 @@ func configureCompression(router *gin.Engine, settings CompressionSettings) erro
 	}
 
 	router.Use(ginGzip.Gzip(level, opts...))
+	router.Use(recordResponseSize)
```
applike-ss, line 58: does it make sense to set this here? I would think that you want to record the request/response size no matter whether compression is anebled or not.

pkg/apiserver/compression.go
```diff
@@ -54,58 +55,78 @@ func configureCompression(router *gin.Engine, settings CompressionSettings) erro
...
+	}
+	reader, err := gzip.NewReader(body)
 
-func (r gzipBodyReader) Close() error {
-	err := r.reader.Close()
 	if err != nil {
-		return err
+		// the body is not a proper gzip encoded body, so don't do anything
+		// the client most likely set the wrong content encoding on the message
+		return
```
applike-ss, line 94: shouldn't we actually let it fail to adhere to standards? i would think that we are not supposed to "fix" something that is wrong from the client side

pkg/apiserver/compression_io.go
```diff
@@ -0,0 +1,61 @@
...
+	return readBytes, err
+}
+
+func (r *gzipBodyReader) Close() error {
+	err := r.reader.Close()
+	if err != nil {
+		return err
+	}
+
+	return r.body.Close()
```
applike-ss, line 60: this could be leaking if reader can't be closed, but i guess that is very unlikely?

pkg/apiserver/compression_io.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+func (r *countingBodyReader) Read(p []byte) (int, error) {
+	readBytes, err := r.ReadCloser.Read(p)
+	r.readCompressedBytes += readBytes
+
+	return readBytes, err
+}
+
+type gzipBodyReader struct {
```
applike-ss, line 41: i would have liked to see factory functions for these structs and them also being used where the structs are currently crafted manually.

Also i would've liked if this pr was split up into the functional bits it consists of.

pkg/apiserver/middleware_logger.go
```diff
@@ -26,35 +27,53 @@ func LoggingMiddleware(logger log.Logger) gin.HandlerFunc {
...
+			ctxLogger = ctxLogger.WithFields(*sizeFields)
+		}
+
+		if sizeFields := getResponseSizeFields(ctx); sizeFields != nil {
+			ctxLogger = ctxLogger.WithFields(*sizeFields)
+		}
+
 		if requestId := req.Header.Get("X-Request-Id"); requestId != "" {
 			ctxLogger = ctxLogger.WithFields(log.Fields{
 				"request_id": requestId,
```
applike-ss, line 79: would it make sense to do something like:
```go
ctxLoggerFields := log.Fields{
			"bytes":                    ginCtx.Writer.Size(),
			"client_ip":                ginCtx.ClientIP(),
			"host":                     req.Host,
			"protocol":                 req.Proto,
			"request_method":           method,
			"request_path":             path,
			"request_path_raw":         pathRaw,
			"request_query":            queryRaw,
			"request_query_parameters": queryParameters,
			"request_referer":          referer,
			"request_time":             requestTimeSecond,
			"scheme":                   req.URL.Scheme,
			"status":                   ginCtx.Writer.Status(),
			"bytes":              ginCtx.Writer.Size(),
			"client_ip":          ginCtx.ClientIP(),
			"host":               req.Host,
			"protocol":           req.Proto,
			"request_method":     method,
			"request_path":       path,
			"request_path_raw":   pathRaw,
			"request_query":      queryRaw,
			"request_referer":    referer,
			"request_user_agent": userAgent,
			"request_time":       requestTimeSecond,
			"scheme":             req.URL.Scheme,
			"status":             status,
		})

		// only log query parameters in full for successful requests to avoid logging them from bad crawlers
		if status != http.StatusUnauthorized && status != http.StatusForbidden && status != http.StatusNotFound {
			queryParameters := make(map[string]string)
			query := req.URL.Query()

			for k := range query {
				queryParameters[k] = query.Get(k)
			}

			ctxLoggerFields.Add(log.Fields{
				"request_query_parameters": queryParameters,
			}))
		}

		if sizeFields := getRequestSizeFields(ctx); sizeFields != nil {
			ctxLoggerFields.Add(*sizeFields)
		}

		if sizeFields := getResponseSizeFields(ctx); sizeFields != nil {
			ctxLoggerFields.Add(*sizeFields)
		}

		if requestId := req.Header.Get("X-Request-Id"); requestId != "" {
			ctxLogger = ctxLoggerFields.Add(log.Fields{
				"request_id": requestId,
.....
		}
		ctxLogger := chLogger.
					WithContext(ctx).
					WithFields(ctxLoggerFields)
```
?

applike-ss (changes requested):

docs/getting_started/module.md
ajscholl: Because this is internal state and this has no place in a getting started guide. I also made the ModuleConfig type now private as there is no use in having that public either (already did the same for ModuleState).

pkg/apiserver/compression.go
ajscholl: We are already recording the response size by default in the field `bytes`, so we only need to record the uncompressed size when we are compressing (and `bytes` will then carry the compressed size)

pkg/apiserver/compression.go
ajscholl: Hmm, the problem here is for me how to handle this properly. I can't return an error because the library doesn't expect an error return. So I have 3 options:
- ignore the error as we do (and did in the past). the request will either fail once we try to parse the body as json or similar or it will succeed (if the content-encoding header was just bogus). kinda in line with specifying any other content-encoding header we currently don't handle
- create a special reader which immediately returns the correct error
- add it to the errors of the gin request. I guess we will print this at the end, but still continue to handle the request

pkg/apiserver/compression_io.go
ajscholl: better safe than sorry, fixed

applike-ss (approved):

PR #966, opened 2022-11-25
Improve db pkg

glebnaz:
- add tx to db pkg
- add new client with cox
- add new function with ctx
- gofumpt project


pkg/db/client.go
```diff
@@ -138,40 +139,76 @@ func (c *ClientSqlx) GetResult(query string, args ...interface{}) (*Result, erro
...
+	return c.db.GetContext(ctx, dest, query, args...)
+}
+
+func (c *ClientSqlx) BeginTx(ctx context.Context, ops *sql.TxOptions) (*sql.Tx, error) {
+	c.logger.Debug("start tx")
+
+	return c.db.BeginTx(ctx, ops)
+}
+
+func (c *ClientSqlx) OnTx(ctx context.Context, ops *sql.TxOptions, do func(ctx context.Context, tx *sql.Tx) error) error {
```
ajscholl, line 186: Please revisit the error handling in this function:
- name the return value and use multierror.Append to return both errors should there be an error during rollback
- check your error handling in this function - returning no error in case of `do` returning an error looks extremely fishy
- the last condition looks quite redundant - both branches return `err`

pkg/db/mocks/ClientContext.go
```diff
@@ -0,0 +1,420 @@
+// Code generated by mockery v2.13.1. DO NOT EDIT.
```
ajscholl, line 1: Does the interface for this still exist?

pkg/db/mocks/SqlResult.go
```diff
@@ -50,3 +50,18 @@ func (_m *SqlResult) RowsAffected() (int64, error) {
...
 	return r0, r1
 }
+
+type mockConstructorTestingTNewSqlResult interface {
+	mock.TestingT
+	Cleanup(func())
+}
+
+// NewSqlResult creates a new instance of SqlResult. It also registers a testing interface on the mock and a cleanup function to assert the mocks expectations.
+func NewSqlResult(t mockConstructorTestingTNewSqlResult) *SqlResult {
```
ajscholl, line 60: Where is this coming from?

pkg/db/client.go
```diff
@@ -138,40 +140,77 @@ func (c *ClientSqlx) GetResult(query string, args ...interface{}) (*Result, erro
...
+		return err
+	}
+
+	defer func() {
+		if err != nil {
+			errRollback := tx.Rollback()
+			if errRollback != nil {
+				err = multierror.Append(err, errRollback)
+				c.logger.Error("error rollback tx: %w", errRollback)
+				return
```
ajscholl, line 199: You are not returning the error here - you are assigning a new value to a local variable and then throwing that away

pkg/db/client.go
```diff
@@ -138,40 +140,77 @@ func (c *ClientSqlx) GetResult(query string, args ...interface{}) (*Result, erro
...
+	if err != nil {
+		return err
+	}
+
+	defer func() {
+		if err != nil {
+			errRollback := tx.Rollback()
+			if errRollback != nil {
+				err = multierror.Append(err, errRollback)
+				c.logger.Error("error rollback tx: %w", errRollback)
```
ajscholl, line 198: Please don't log and return errors (on level error) - either do one or the other. If something calling your code can handle the error, we still have errors in the logs although everything would be fine.

pkg/db/client.go
```diff
@@ -138,40 +140,77 @@ func (c *ClientSqlx) GetResult(query string, args ...interface{}) (*Result, erro
...
+	}()
+
+	err = do(ctx, tx)
+	if err != nil {
+		return err
+	}
+
+	err = tx.Commit()
+	if err != nil {
+		return err
```
ajscholl, line 212: This check and return is redundant. But maybe it would be good to wrap the errors so we know from which part they are coming from

ajscholl (changes requested):

ajscholl (approved):

pkg/fixtures/purger_mysql.go
```diff
@@ -29,6 +30,7 @@ func newMysqlPurger(config cfg.Config, logger log.Logger, tableName string) (*my
 }
 
 func (p *mysqlPurger) purgeMysql() error {
+	ctx := context.Background()
```
j4k4, line 33: How much effort would it be to provide a "real context" here?

PR #984, opened 2022-12-20
cfg: introduce nodecode flag to disable string augmentation

j4k4:
string values can be prefixed with !nodecode to disable string augmentation..
E.g. setting an env variable like `FOO="!nodecode this should be {plain} bar"` will not try to replace `{plain}`..

### DIFF
--- a/pkg/cfg/config.go
+++ b/pkg/cfg/config.go
@@ -9,10 +9,15 @@ import (
 
 	"github.com/go-playground/validator/v10"
 	"github.com/hashicorp/go-multierror"
+	"github.com/justtrackio/gosoline/pkg/funk"
 	"github.com/justtrackio/gosoline/pkg/mapx"
 	"github.com/justtrackio/gosoline/pkg/refl"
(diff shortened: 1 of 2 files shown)


pkg/cfg/config.go
```diff
@@ -345,6 +351,19 @@ func (c *config) UnmarshalKey(key string, output interface{}, defaults ...Unmars
 }
 
 func (c *config) augmentString(str string) string {
+	groups := valFlagsRegexp.FindStringSubmatch(str)
+	flags := make([]string, 0)
+
+	if len(groups[2]) > 0 {
+		flags = strings.Split(groups[2], ",")
+		flags = funk.Map(flags, strings.ToLower)
+		str = groups[3]
```
applike-ss, line 360: this looks unsafe, because we didn't check that the slice does have 4 elements or more in total

applike-ss (changes requested):

ajscholl (approved):

PR #988, opened 2023-01-02
cache: add type parameter and made cache generic - breaking change;

ajscholl:
This change refactors the cache package and forces you to specify a type for the value you store in the cache (and frees you from manually casting the values returned from the cache to the correct type again). To improve the package further, two new methods were added which abstract over the check - request - set pattern. As a last change, you can now specify an option to also cache zero values (e.g., nil) for a specified time, allowing you to basically cache missing values as well.


pkg/cache/ccache_test.go
```diff
@@ -1,38 +1,56 @@
 package cache
```
j4k4, line 1: shouldn't this be package `cache_test`?

j4k4 (approved):

PR #990, opened 2023-01-05
cloud: added appIds group replacement;

applike-ss:


applike-ss:
Also available here: https://github.com/justtrackio/gosoline/tree/add-app-group-replacement-for-naming

PR #992, opened 2023-01-09
Improve fixture loading

selm0:
This PR adds extended fixture capabilities such as named lists, auto number generators and delayed fixture loading so env and app options are properly applied


pkg/fixtures/named.go
```diff
@@ -0,0 +1,150 @@
...
+	if identifiable, ok := value.(mdl.Identifiable); ok {
+		return mdl.EmptyIfNil(identifiable.GetId()), true
+	}
+
+	if model, ok := value.(mdlsub.Model); ok {
+		switch modelId := model.GetId().(type) {
+		case string:
+			return modelId, true
+		case *string:
+			return mdl.EmptyIfNil(modelId), true
```
bogdanfinn, line 109: could you not just check if its a pointer in general here (maybe via reflection) ? if yes return the value with an `EmptyIfNil()` before otherwise without?

or is that to slow?

pkg/test/suite/options_suite.go
```diff
@@ -116,9 +116,9 @@ func WithEnvSetup(setups ...func() error) Option {
 	}
 }
 
-func WithFixtures(fixtureSets []*fixtures.FixtureSet) Option {
+func WithFixtureBuilderFactory(factories ...fixtures.FixtureBuilderFactory) Option {
```
bogdanfinn, line 119: maybe `WithFixtureBuilderFactories()` ?

bogdanfinn (commented):
@selm0 very minor comments

pkg/fixtures/named.go
selm0: mdl.EmptyIfNil doesn't work with reflected values. I added some more generic code, please have another look

pkg/fixtures/loader.go
```diff
@@ -25,7 +25,7 @@ func NewFixtureLoader(ctx context.Context, config cfg.Config, logger log.Logger)
 	logger = logger.WithChannel("fixture_loader")
 
 	settings := &fixtureLoaderSettings{}
-	config.UnmarshalKey("fixtures", settings)
+	config.UnmarshalKey("fixtureSets", settings)
```
bogdanfinn, line 28: actually i personally do not like the renaming of the config key as the package itself is also named fixtures.
its personal preference i would say but it feels more consistent if the config key settings (root level) is the same as the package name.

j4k4 (approved):

PR #995, opened 2023-01-11
limit: added rate limiting package;

nikplxjt:
We do not have rate limiting software in gosoline yet. Currently we implement rate limiting only in our internal projects but as the demand grows to having it in multiple project and the code is pretty generic and can we applied in multiple situations, we want to have it in gosoline.

### DIFF
--- a/pkg/limit/fixed_window_ddb.go
+++ b/pkg/limit/fixed_window_ddb.go
@@ -1,0 +1,110 @@
+package limit
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/justtrackio/gosoline/pkg/cfg"
+	"github.com/justtrackio/gosoline/pkg/clock"
+	"github.com/justtrackio/gosoline/pkg/ddb"
+	"github.com/justtrackio/gosoline/pkg/log"
+)
+
+const (
+	attrVal = "val"
+	attrTtl = "ttl"
+)
+
+type window struct {
+	Key string `json:"key" ddb:"key=hash"`
+	Val int    `json:"val"`
+	Ttl int64  `json:"ttl" ddb:"ttl=enabled"`
+}
+
+type fixedWindowDdb struct {
+	logger log.Logger
+	clock  clock.Clock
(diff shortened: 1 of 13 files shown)


pkg/limit/fixed_window_ddb.go
```diff
@@ -0,0 +1,100 @@
...
+		ReturnAllNew()
+
+	resp, err := f.repo.UpdateItem(ctx, reset, item)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	if resp.ConditionalCheckFailed {
+		increment := f.repo.UpdateItemBuilder().
+			Add(attrVal, 1).
```
selm0, line 85: Do you really need this second request?

DDB has the SetIfNotExists capability, and you can call Add on a non-existent property and it will use the the default value (for numbers: 0)

selm0: So, your previous Update could look like this:

```
 f.repo.UpdateItemBuilder().
		WithCondition(ddb.Or(
			ddb.AttributeNotExists(attrTtl),
			ddb.Lt(attrTtl, now().Unix()))).
		Add(attrVal, 1).
		SetIfNotExists(attrTtl, now().Add(f.window).Unix()).
		ReturnAllNew()
```

You need to adapt the conditional check logic then; in the new case it would mean that the element already has expired and we're in a new window, so directly returning with the permission to execute is fine?

pkg/limit/fixed_window_ddb.go
```diff
@@ -0,0 +1,100 @@
...
+	attrVal = "val"
+	attrTtl = "ttl"
+)
+
+type window struct {
+	Key string `json:"key" ddb:"key=hash"`
+	Val int    `json:"val"`
+	Ttl int64  `json:"ttl" ddb:"ttl=enabled"`
+}
+type fixedWindowDdb struct {
```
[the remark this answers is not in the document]
nikplxjt: done

selm0 (approved):

PR #996, opened 2023-01-18
redis: added LPush, LRem, RPop and all redis 1.0.0 set operations

n4n0lix:
removed `withScores` parameter from `ZRandMember` as it doesn't do anything (go-redis bug, fixed in later version and then removed from signature)

added support for optional arguments in `ZADD` and `ZRANGE`
- `ZAddArgs`
- `ZAddArgsIncr`
- `ZRangeArgs`
- `ZRangeArgsWithScore`

added following redis operations, which all are supported since redis 1.0.0
- `LPush`
- `LRem`
- `RPop`
- `SDiff`
- `SDiffStore`
- `SInter`
- `SInterStore`
- `SMembers`
- `SMove`
- `SPop`
- `SRem`
- `SRandMember`
- `SUnion`
- `SUnionStore`

### DIFF
--- a/pkg/redis/client.go
+++ b/pkg/redis/client.go
@@ -19,6 +19,33 @@ type ErrCmder interface {
 	Err() error
 }
 
+type Z struct {
+	Score  float64
+	Member interface{}
+}
+
+type ZAddArgs struct {
+	Key     string
+	NX      bool
+	XX      bool
+	LT      bool
+	GT      bool
+	Ch      bool
+	Incr    bool
+	Members []Z
+}
+
+type ZRangeArgs struct {
+	Key     string
+	Start   interface{}
+	Stop    interface{}
+	ByScore bool
+	ByLex   bool
+	Rev     bool
+	Offset  int64
+	Count   int64
+}
+
 //go:generate mockery --name Pipeliner
 type Pipeliner interface {
 	baseRedis.Pipeliner
@@ -45,7 +72,10 @@ type Client interface {
 	BLPop(ctx context.Context, timeout time.Duration, keys ...string) ([]string, error)
 	LPop(ctx context.Context, key string) (string, error)
 	LLen(ctx context.Context, key string) (int64, error)
+	LPush(ctx context.Context, key string, values ...interface{}) (int64, error)
+	LRem(ctx context.Context, key string, count int64, value interface{}) (int64, error)
 	RPush(ctx context.Context, key string, values ...interface{}) (int64, error)
+	RPop(ctx context.Context, key string) (string, error)
 
 	HDel(ctx context.Context, key string, fields ...string) (int64, error)
 	HExists(ctx context.Context, key string, field string) (bool, error)
@@ -59,7 +89,18 @@ type Client interface {
 
 	SAdd(ctx context.Context, key string, values ...interface{}) (int64, error)
 	SCard(ctx context.Context, key string) (int64, error)
+	SDiff(ctx context.Context, keys ...string) ([]string, error)
+	SDiffStore(ctx context.Context, destination string, keys ...string) (int64, error)
+	SInter(ctx context.Context, keys ...string) ([]string, error)
+	SInterStore(ctx context.Context, destination string, keys ...string) (int64, error)
+	SMembers(ctx context.Context, key string) ([]string, error)
 	SIsMember(ctx context.Context, key string, value interface{}) (bool, error)
+	SMove(ctx context.Context, sourceKey string, destKey string, member interface{}) (bool, error)
+	SPop(ctx context.Context, key string) (string, error)
+	SRem(ctx context.Context, key string, values ...interface{}) (int64, error)
+	SRandMember(ctx context.Context, key string) (string, error)
+	SUnion(ctx context.Context, keys ...string) ([]string, error)
+	SUnionStore(ctx context.Context, destination string, keys ...string) (int64, error)
 
 	Decr(ctx context.Context, key string) (int64, error)
 	DecrBy(ctx context.Context, key string, amount int64) (int64, error)
@@ -71,16 +112,19 @@ type Client interface {
 	PFMerge(ctx context.Context, dest string, keys ...string) (string, error)
 
 	ZAdd(ctx context.Context, key string, score float64, member string) (int64, error)
+	ZAddArgs(ctx context.Context, args ZAddArgs) (int64, error)
+	ZAddArgsIncr(ctx context.Context, args ZAddArgs) (float64, error)
 	ZCard(ctx context.Context, key string) (int64, error)
 	ZCount(ctx context.Context, key string, min string, max string) (int64, error)
 	ZIncrBy(ctx context.Context, key string, increment float64, member string) (float64, error)
 	ZScore(ctx context.Context, key string, member string) (float64, error)
 	ZMScore(ctx context.Context, key string, members ...string) ([]float64, error)
 	ZRange(ctx context.Context, key string, start int64, stop int64) ([]string, error)
-	ZRandMember(ctx context.Context, key string, count int, withScores bool) ([]string, error)
+	ZRangeArgs(ctx context.Context, args ZRangeArgs) ([]string, error)
+	ZRangeArgsWithScore(ctx context.Context, args ZRangeArgs) ([]Z, error)
+	ZRandMember(ctx context.Context, key string, count int) ([]string, error)
 	ZRank(ctx context.Context, key string, member string) (int64, error)
 	ZRem(ctx context.Context, key string, members ...string) (int64, error)
-	ZRevRange(ctx context.Context, key string, start int64, stop int64) ([]string, error)
 	ZRevRank(ctx context.Context, key string, member string) (int64, error)
 
 	IsAlive(ctx context.Context) bool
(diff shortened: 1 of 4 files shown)


pkg/redis/mocks/Client.go
```diff
@@ -1,4 +1,4 @@
-// Code generated by mockery v2.13.1. DO NOT EDIT.
+// Code generated by mockery v2.15.0. DO NOT EDIT.
```
applike-ss, line 1: @j4k4 what do you think, should we update mockery? If so, should we simply go with the most recent version (v2.16.0)?

j4k4: not sure about the impact and how much work that would be

applike-ss: maybe @BastiJD can check that for us?

applike-ss: i did a `go install github.com/vektra/mockery/v2@v2.16.0` and recreated all mocks with:
```shell
BASE=$PWD
for folder in $(grep -H -r "go:generate mockery" * | cut -d':' -s -f1 | grep "\\.go" | xargs -n1 dirname | uniq)                                                                   ─╯
      do
        echo "processing ${folder}"
        cd ${BASE}/${folder}

        # gather all go generate statements for mockery and cut away the parts that are not valid shell code
        if ! ls | grep ".go" > /dev/null; then
          continue
        fi
        grep -H "go:generate mockery" *.go | cut -d'/' -f3 | cut -b 13- | sh
      done

```
and then did a go test and it works just fine. So seems there is no breaking changes in gosoline

n4n0lix: To not mix changes too much I will regenerate the mocks with v2.13.1 and create a ticket to update mockery. I will pick it up in the next few days.

pkg/redis/client.go
```diff
@@ -496,6 +642,28 @@ func (c *redisClient) ZRevRange(ctx context.Context, key string, start int64, st
...
+		result = make([]Z, len(zs))
+		for i := range zs {
+			result[i] = Z{
+				Score:  zs[i].Score,
+				Member: zs[i].Member,
+			}
+		}
+	} else {
+		result = make([]Z, 0)
+	}
```
applike-ss, line 662: Please follow this pattern (early return) when coding in gosoline:
```suggestion
	if err != nil {
		return make([]Z, 0), err
	}
	
	zs := cmd.(*baseRedis.ZSliceCmd).Val()
	result = make([]Z, len(zs))
	for i := range zs {
		result[i] = Z{
			Score:  zs[i].Score,
			Member: zs[i].Member,
		}
	}

```

pkg/redis/client.go
```diff
@@ -496,6 +642,28 @@ func (c *redisClient) ZRevRange(ctx context.Context, key string, start int64, st
 	return cmd.(*baseRedis.StringSliceCmd).Val(), err
 }
 
+func (c *redisClient) ZRevRangeWithScores(ctx context.Context, key string, start int64, stop int64) ([]Z, error) {
```
applike-ss, line 645: It seems that this is a deprecated function, so we should not add it. See: https://redis.io/commands/zrevrange/

Quote:
```
As of Redis version 6.2.0, this command is regarded as deprecated.

It can be replaced by [ZRANGE](https://redis.io/commands/zrange) with the REV argument when migrating or writing new code.
```

I think you might be able to achieve the same with `ZRangeArgsWithScores` from the base client.

applike-ss (changes requested):

pkg/redis/client.go
n4n0lix: Ended up scrapping the error handling. Worst case go-redis returns a nil-slice here, which will then result in a []Z slice of length 0 being returned.

pkg/redis/client.go
```diff
@@ -533,3 +768,38 @@ func (c *redisClient) execute(ctx context.Context, wrappedCmd func() ErrCmder) (
...
 	})
 }
+
+func (c *redisClient) toGosolineZs(zs []baseRedis.Z) []Z {
+	result := make([]Z, len(zs))
+	for i := range zs {
+		result[i] = Z{
+			Score:  zs[i].Score,
+			Member: zs[i].Member,
+		}
```
applike-ss, line 778: Wouldn't this work as well?
```suggestion
		result[i] = Z(zs[i])
```

pkg/redis/client.go
```diff
@@ -533,3 +768,38 @@ func (c *redisClient) execute(ctx context.Context, wrappedCmd func() ErrCmder) (
...
+	return result
+}
+
+func (c *redisClient) toGoRedisZs(zs []Z) []baseRedis.Z {
+	result := make([]baseRedis.Z, len(zs))
+	for i := range zs {
+		result[i] = baseRedis.Z{
+			Score:  zs[i].Score,
+			Member: zs[i].Member,
+		}
```
applike-ss, line 790:
```suggestion
		result[i] = baseRedis.Z(zs[i])
```

pkg/redis/client.go
```diff
@@ -533,3 +768,38 @@ func (c *redisClient) execute(ctx context.Context, wrappedCmd func() ErrCmder) (
...
+
+func (c *redisClient) toGoRedisZAddArgs(args ZAddArgs, zs []baseRedis.Z) baseRedis.ZAddArgs {
+	return baseRedis.ZAddArgs{
+		NX:      args.NX,
+		XX:      args.XX,
+		LT:      args.LT,
+		GT:      args.GT,
+		Ch:      args.Ch,
+		Members: zs,
+	}
```
applike-ss, line 804:
```suggestion
	return baseRedis.ZAddArgs(args)
```

pkg/redis/client_test.go
```diff
@@ -324,42 +476,100 @@ func (s *ClientWithMiniRedisTestSuite) TestZRange() {
...
+	s.NoError(err, "there should be no error on ZRangeArgsWithScore")
+	s.EqualValues(2, len(zs))
+	s.Equal("member1", zs[0].Member)
+	s.Equal(float64(50), zs[0].Score)
+	s.Equal("member2", zs[1].Member)
+	s.Equal(float64(90), zs[1].Score)
 }
 
-func (s *ClientWithMiniRedisTestSuite) TestZRem() {
+func (s *ClientWithMiniRedisTestSuite) TestZRevRangeWithScores() {
```
applike-ss, line 501:
```suggestion
func (s *ClientWithMiniRedisTestSuite) TestZRangeArgsWithScores() {
```

pkg/redis/client_test.go
```diff
@@ -324,42 +476,100 @@ func (s *ClientWithMiniRedisTestSuite) TestZRange() {
 	s.Equal("member2", members[1])
 }
 
-func (s *ClientWithMiniRedisTestSuite) TestZRank() {
+func (s *ClientWithMiniRedisTestSuite) TestZRangeWithScores() {
```
applike-ss, line 479:
```suggestion
func (s *ClientWithMiniRedisTestSuite) TestZRangeArgsWithScores() {
```

pkg/redis/client_test.go
```diff
@@ -324,42 +476,100 @@ func (s *ClientWithMiniRedisTestSuite) TestZRange() {
...
+	s.NoError(err, "there should be no error on ZRangeArgsWithScore")
+	s.EqualValues(2, len(zs))
+	s.Equal("member3", zs[0].Member)
+	s.Equal(float64(100), zs[0].Score)
+	s.Equal("member2", zs[1].Member)
+	s.Equal(float64(90), zs[1].Score)
 }
 
-func (s *ClientWithMiniRedisTestSuite) TestZRevRange() {
+func (s *ClientWithMiniRedisTestSuite) TestZRangeByScoreWithLimit() {
```
applike-ss, line 524:
```suggestion
func (s *ClientWithMiniRedisTestSuite) TestZRangeArgsByScoreWithLimit() {
```

applike-ss (changes requested):

pkg/redis/client.go
```diff
@@ -457,16 +651,57 @@ func (c *redisClient) ZMScore(ctx context.Context, key string, members ...string
...
+	zRangeArgs := baseRedis.ZRangeArgs{
+		Key:     args.Key,
+		Start:   args.Start,
+		Stop:    args.Stop,
+		ByScore: args.By == ByScore,
+		ByLex:   args.By == ByLex,
+		Rev:     args.Rev,
+		Offset:  args.LimitOffset,
+		Count:   args.LimitCount,
+	}
```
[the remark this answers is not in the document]
n4n0lix: I guess that is a matter of taste, as the `ByScore` and `ByLex` options are mutual exclusive which is not reflected by the go-redis interface which will silently default to `ByLex`.

I'm fine with either way, both approaches (mirroring interface 1:1 vs. improving on it) have their pros and cons.

pkg/redis/client.go
n4n0lix: Done

pkg/redis/client.go
n4n0lix: Done

pkg/redis/client.go
n4n0lix: Would love to do it that way, but `gosoline.redis.ZAddArgs` has the field`[]gosoline.Z Members` where `baseRedis.ZAddArgs` has the field`[]baseRedis.Z Members`

pkg/redis/client_test.go
n4n0lix: Done

pkg/redis/client_test.go
n4n0lix: Done

pkg/redis/client_test.go
n4n0lix: Done (as `TestZRangeArgsWithScoreRev`)

applike-ss (approved):

PR #1000, opened 2023-01-27
Fix blob runner test suite exit

nikplxjt:
With the current implementation of the application test suite we can run into problems with foreground modules:
If we have some module running in the service stage that coordinates its exit using the context provided by the kernel that our test depends on, the test will be started in the application stage. If the test application then exists, the kernel package will not teardown modules in the service or essential stage, which will lead to our service being run indefinitely.
This PR makes every module in the application stage (the app under test) an essential module, which will lead to all other modules being teared down correctly.

### DIFF
--- a/pkg/test/suite/testcase_application.go
+++ b/pkg/test/suite/testcase_application.go
@@ -7,7 +7,9 @@ import (
 
 	"github.com/justtrackio/gosoline/pkg/appctx"
 	"github.com/justtrackio/gosoline/pkg/application"
+	"github.com/justtrackio/gosoline/pkg/cfg"
 	"github.com/justtrackio/gosoline/pkg/kernel"
+	"github.com/justtrackio/gosoline/pkg/log"
 	"github.com/justtrackio/gosoline/pkg/test/env"
 	"github.com/stretchr/testify/assert"
 )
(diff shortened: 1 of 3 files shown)


pkg/test/suite/testcase_application.go
```diff
@@ -90,3 +96,24 @@ func runTestCaseApplication(t *testing.T, suite TestingSuite, suiteOptions *suit
...
+	return struct {
+		kernel.Module
+		kernel.EssentialModule
+	}{
+		mod,
+		kernel.EssentialModule{},
+	}
+}
+
+func makeAppModuleEssential(factory kernel.ModuleFactory) kernel.ModuleFactory {
```
selm0, line 110:
```suggestion
func newAppModuleEssential(factory kernel.ModuleFactory) kernel.ModuleFactory {
```

pkg/test/suite/testcase_application.go
```diff
@@ -90,3 +96,24 @@ func runTestCaseApplication(t *testing.T, suite TestingSuite, suiteOptions *suit
 	testcase(appUnderTest)
 	app.Stop("test done")
 }
+
+func buildEssentialModule(mod kernel.Module) kernel.Module {
```
selm0, line 100:
```suggestion
func newEssentialModule(mod kernel.Module) kernel.Module {
```

selm0 (approved):

PR #1008, opened 2023-02-09
stream: allow to set compression per message;

netsi:


pkg/stream/message.go
```diff
@@ -38,3 +38,9 @@ func (m *Message) UnmarshalFromBytes(data []byte) error {
 func (m *Message) UnmarshalFromString(data string) error {
 	return m.UnmarshalFromBytes([]byte(data))
 }
+
+func (m *Message) WithGzipCompression() *Message {
```
j4k4, line 42: I'm not so sure about this one. It misleads a little bit into thinking that calling this would compress the message..  
Instead it's just setting an attribute..

netsi: good point. I guess it either should be changed to allow anyone to set any message attribute or to also perform the compression. Will have to rethink this.

netsi:
closing this, because we don't really need it

PR #1026, opened 2023-03-14
fixing the issue for invalid dimension

Miraddo:
before publishing data points to metrics, in or log we get this error:

```
invalid metric dimension: 1 error occurred:
	* invalid dimension 'path' = '' for metric NAME, this will later be rejected
```

The issue is that we send '/' as the path, but due to the use of TrimRight in the getAbsolutePath() method, the path gets deleted. To resolve this problem, we can add an 'if' statement to check whether the absolutePath is empty, and if so, we can add '/' to it.

### DIFF
--- a/pkg/apiserver/definition.go
+++ b/pkg/apiserver/definition.go
@@ -24,7 +24,7 @@ func (d *Definition) getAbsolutePath() string {
 	groupPath := d.group.getAbsolutePath()
 
 	absolutePath := fmt.Sprintf("%s/%s", groupPath, d.relativePath)
-	absolutePath = strings.TrimRight(absolutePath, "/")
+	absolutePath = trimRightPath(absolutePath)
 
 	return removeDuplicates(absolutePath)
 }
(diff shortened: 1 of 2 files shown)


pkg/apiserver/definition.go
```diff
@@ -26,6 +26,10 @@ func (d *Definition) getAbsolutePath() string {
 	absolutePath := fmt.Sprintf("%s/%s", groupPath, d.relativePath)
 	absolutePath = strings.TrimRight(absolutePath, "/")
 
+	if absolutePath == "" {
```
selm0, line 29: We have similar logic in gosoline/pkg/apiserver/middleware_metric.go:38
Does it make sense to unify them and look for other occurrences in this package?

Miraddo: yes make sense to have a small function to do the Trim for the path if it's empty we can return `/` in a safe way it will be more clear and we don't need to duplicate!

selm0 (approved):

PR #1031, opened 2023-04-17
Add circuit breaker to http client

ajscholl:


pkg/http/circuit_breaker.go
```diff
@@ -0,0 +1,151 @@
...
+
+	// state updated with atomics
+	recentFailures atomic.Int32
+	nextRetryAt    atomic.Int64
+}
+
+type circuitBreakerName string
+
+type CircuitIsOpenError struct {
+}
```
j4k4, line 38: should these type declarations should go up top?
not between the client stuff?

j4k4 (approved):

PR #1036, opened 2023-06-05
added more mdl methods, fixed keylock, small improvements for local development

ajscholl:
- added some methods to mdl to work with empty values
- fixed KeyLock race condition
- run change history migrations per model in parallel for moderate speedup of local development with remote database
- improved error message for missing external container


erdac:
+1

PR #1041, opened 2023-06-21
Db migrations goose

j4k4:


glebnaz:
I don't have any comments on the review related to the code stylus. 

A small question about the configuration. As I see we now have two providers, golang_migrate and goose. I'm worried that this may lead to misconfiguration on the side of our services. Is it really worth keeping golang_migrate as a possible option? 

I guess this is to keep backward compatibility. If I am right, how long do we want to keep golang_migrate as a possible option?

glebnaz:
> There is no timeline to drop support so far.. But I would also be fine with making goose the default migration tool right away..

Good idea =) Could you please do?

PR #1047, opened 2023-08-08
ci: add sbom generation

ss89:


ss89:
@j4k4 mind a review?

applike-ss:
We've internally discussed this and came to the conclusion that it makes more sense to create sboms for go apps (and libraries which are only distributed as binary), so we are closing this one.

PR #1059, opened 2023-11-02
kernel: add health checks to modules

j4k4:
## Health checks
This release will add a health check mechanism to the kernel. There is a `HealthCheckedModule` interface now, which can be implemented by kernel modules to indicate their healthiness. On application run, these health checks are used. The application is switching into the running state, which can be checked by calling `app.Running()`, only after all modules are reporting a healthy state. Without implementing this interface, a module will report healthy by default.

The interface is defined as:
```go
type HealthCheckedModule interface {
	IsHealthy(ctx context.Context) (bool, error)
}
```

Checking the healthiness state on run can be configured via:
```yaml
# configuration for the health check on application run
kernel:
  health_check:
    # if the app isn't getting healthy after calling run, the application will stop again
    timeout: 10s
    # the interval for checking the healthyness on run
    wait_interval: 1s
```

In addition to the health check on run, the healthiness state of the modules is checked on calling the api health check routes now, too.

## Breaking changes
There are two small breaking changes included in this release:

#### config name changed for the kernel kill timeout:
```
kernel:
    killTimeout: 10s
```
changed to:
```
kernel:
    kill_timeout: 10s
```

#### the fixture builder factory is allowed to return an error now
```
type FixtureBuilderFactory func(ctx context.Context) FixtureBuilder
```
changed to:
```
type FixtureBuilderFactory func(ctx context.Context) (FixtureBuilder, error)
```


pkg/cli/cli.go
```diff
@@ -10,7 +10,7 @@ import (
 )
 
 type kernelSettings struct {
-	KillTimeout time.Duration `cfg:"killTimeout" default:"10s"`
+	KillTimeout time.Duration `cfg:"kill_timeout" default:"10s"`
```
selm0, line 13: Do we still need this as a separate setting here? It's already defined somewhere else, not sure where though :thinking:

pkg/kernel/factory.go
```diff
@@ -30,21 +33,25 @@ func NewFactory(ctx context.Context, config cfg.Config, logger log.Logger, bp *b
...
 		return nil, fmt.Errorf("can not build kernel factory: %w", err)
 	}
 
 	return factory, nil
 }
 
-func (f *factory) GetKernel() Kernel {
-	k := New(f.ctx, f.config, f.logger, f.middlewares, f.stages)
+func (f *factory) GetKernel() (Kernel, error) {
+	f.kernel.init(f.middlewares, f.stages)
```
selm0, line 48: Why do you need to store the kernel instance on the factory? And should it be secured as a proper singleton?

pkg/kernel/health_check.go
```diff
@@ -0,0 +1,70 @@
...
+		if m.Err != nil {
+			err = multierror.Append(err, fmt.Errorf("error during health check in module %s: %w", m.Name, m.Err))
+		}
+	}
+
+	return err
+}
+
+type (
+	HealthChecker        func() HealthCheckResult
```
selm0, line 62: Nit: I think interface names are suggested to end on *er, I'm not sure if there is a convention for function names

pkg/kernel/stage.go
```diff
@@ -76,61 +87,81 @@ func (s *stage) run(k *kernel) {
...
+		sleepTicker.Stop()
+		result = s.healthcheck()
 
-func (s stages) hasModules() bool {
-	// no need to iterate in order as we are only checking
-	for _, stage := range s {
-		if len(stage.modules.modules) > 0 {
-			return true
+		if result.IsHealthy() {
+			return result.Err()
```
selm0, line 129: Why not return nil to make it more transparent there is no error?

pkg/kernel/stage.go
```diff
@@ -76,61 +87,81 @@ func (s *stage) run(k *kernel) {
...
 		}
-	}
 
-	return false
-}
+		if result.Err() != nil {
+			s.logger.Warn("errors during health checks in stage %d: %s", s.index, result.Err())
```
selm0, line 133: Do we log the errors on the health check side? I assume it's more informative with the context around there

pkg/kernel/stages.go
```diff
@@ -0,0 +1,45 @@
...
+				count++
+			}
+		}
+	}
+
+	return count
+}
+
+func (s stages) getIndices() []int {
+	keys := make([]int, len(s))
```
selm0, line 34: keys := maps.Keys(s)
slices.Sort(keys)

return keys

selm0: There are also other helpers in maps, slices and funk; but not all of them provide better readability in your cases here

pkg/metric/daemon.go
```diff
@@ -105,9 +105,14 @@ func (d *Daemon) GetStage() int {
 	return common.StageEssential
 }
 
+func (d *Daemon) IsHealthy(ctx context.Context) (bool, error) {
```
selm0, line 108: Do you want to define a default health check? Then you can just embed that and don't need to define it every time

pkg/test/env/environment.go
```diff
@@ -204,8 +204,13 @@ func (e *Environment) LoadFixtureBuilderFactories(factories ...fixtures.FixtureB
 		return nil
 	}
 
+	var err error
+	var fixtureBuilder fixtures.FixtureBuilder
```
selm0, line 208: Nit: fixtureBUilder doesn't have to be scoped outside the for loop

selm0 (changes requested):

pkg/cli/cli.go
j4k4: you're right, this is not needed anymore :+1:

pkg/kernel/stage.go
j4k4: mh yes, I already just saw that the kernel shuts down if it receives an error on this initial health check..  
So I guess it makes more sense to return nil on healthiness and log a potential error as warning..

pkg/test/env/environment.go
j4k4: :+1:

pkg/kernel/stages.go
j4k4: I've just moved this code and don't want to change it now.. :)

pkg/kernel/stage.go
j4k4: I've added logging to the api health checks and also added the context there: `pkg/apiserver/health_check.go:89`  
Is that answering / solving your question?

pkg/kernel/health_check.go
j4k4: Ya, for me this *er thing adds a certain functionality.. this could be a function or interface..  
I'm open for a better name.. :)

pkg/kernel/stage.go
selm0: Yes

selm0 (approved):

pkg/kernel/stage.go
```diff
@@ -76,61 +87,81 @@ func (s *stage) run(k *kernel) {
...
+	return result
 }
 
-func (s *stage) len() int {
-	return s.modules.len()
-}
+func (s *stage) waitUntilHealthy() error {
+	var result HealthCheckResult
+
+	timeoutTicker := clock.NewRealTimer(s.healthCheckSettings.Timeout)
```
ajscholl, line 121: Should you stop these tickets when leaving the function (aka add a defer calling stop on both)? Otherwise we will leak them

j4k4: added the defer statements :+1:

PR #1062, opened 2023-11-06
guard: implement audit logger;

selm0:


pkg/guard/audit_logger.go
```diff
@@ -0,0 +1,56 @@
...
+
+	if len(deciders) == 0 {
+		logger.Info("no policy allowed access")
+		return
+	}
+
+	rejecter := deciders[len(deciders)-1]
+
+	logger.Info("policy %s denied the access", rejecter.GetID())
+}
```
wthsths, line 37: what if I want to identify the policies that grant access? it appears that the final policy is the one that refuses the request. i would like to view the ids of the policies that permit access.

wthsths (approved):

PR #1066, opened 2023-11-14
guard: unify guard tables;

wthsths:
Increases guard table transparency by unifying guard tables. The original library uses JSON policies, the changes allows us to switch to JSON representation of the policy instead of the fully-denormalized way

### DIFF
--- a/pkg/guard/manager_sql.go
+++ b/pkg/guard/manager_sql.go
@@ -9,16 +9,12 @@ import (
 	"github.com/Masterminds/squirrel"
 	"github.com/justtrackio/gosoline/pkg/cfg"
 	"github.com/justtrackio/gosoline/pkg/db"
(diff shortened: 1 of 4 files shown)


pkg/guard/manager_sql.go
```diff
@@ -15,10 +15,7 @@ import (
 )
 
 const (
-	tablePolicies  = "guard_policies"
-	tableSubjects  = "guard_subjects"
-	tableResources = "guard_resources"
-	tableActions   = "guard_actions"
+	tableName = "guard_policies_new"
```
selm0, line 18: I suggest to stick with `guard_policies` and provide a migration script which transfers the data

selm0 (approved):

selm0 (approved):

PR #1069, opened 2023-12-11
kafka: deduplicate batches;

netsi:
This change will allow us to remove any duplicates from the batches.


netsi:
blocking this until further notice, will add some logging on our fork to see if we actually have duplicates

j4k4 (commented):
Will wait until further investigation :)

netsi:
> Will wait until further investigation :)

I had a look at what might be the reason of the duplicates and I concluded that the duplicates seem to appear when the partitions are rebalanced.  The code responsible seem to have a comment that informs about the possibility of re-processig of messages https://github.com/segmentio/kafka-go/blob/ebca72eaee918d303c532feb7ff29afdcd8c2efa/reader.go#L114 

Sadly from what I can see this process of rebalancing is transaprent to us so we can't change the logic to fiish processing the partial batch

PR #1070, opened 2023-12-11
http: add disable_cookies config flag;

netsi:
Adds a configuration flag `disable_cookies` to the http client.

after this change you should be able to configure a http client without cookies. It shouldn't have any breaking changes, the default option is with cookies enabled.

usage example:

```
http_client:
  default:
    request_timeout: 1s
    retry_count: 3
    follow_redirects: false
    disable_cookies: true
```

### DIFF
--- a/pkg/http/client.go
+++ b/pkg/http/client.go
@@ -71,6 +71,7 @@ type client struct {
 }
 
 type Settings struct {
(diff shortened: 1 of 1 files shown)


netsi:
@j4k4 can you please help me merge this one? I don't have permission to do so

PR #1074, opened 2023-12-20
log: Split logger fields into local and global fields and made them mutable

ajscholl:
This commit introduces a distinction between local and global logger fields in a context (there used to be only global fields in the past). Global fields are forwarded to downstream services when using the stream framework while local fields are kept local to the current application. This can be useful to establish fields shared for the duration of a single request between different services without affecting downstream services.
Additionally, this commit introduces the notation of mutating these fields. If a context already contains fields, a callee can thus change fields in a context, allowing the caller to also log these fields. Support was added for the stream and apiserver frameworks already - when handling a message or a request, we are now able to log the returned error in case of a failure with the fields attached to the context by the consumer callback or api handler.

This commit is a breaking change. The following changes have been performed:

- `log.NewLoggerContext` has been removed from the public API
- `log.AppendLoggerContextField` is now called `log.AppendGlobalLoggerContextField`
- `log.ContextLoggerFieldsResolver` is now called `log.MergedContextLoggerFieldsResolver`

### DIFF
--- a/docs/docs/reference/package-log.mdx
+++ b/docs/docs/reference/package-log.mdx
@@ -75,7 +75,7 @@ logger.Error("Message")
 #### Usage
 
 ```go
-if err := logger.Option(log.WithContextFieldsResolver(log.ContextLoggerFieldsResolver)); err != nil {
+if err := logger.Option(log.WithContextFieldsResolver(log.ContextFieldsResolver)); err != nil {
 	panic(err)
 }
 ```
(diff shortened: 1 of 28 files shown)


docs/docs/reference/package-log.mdx
```diff
@@ -75,7 +75,7 @@ logger.Error("Message")
 #### Usage
 
 ```go
-if err := logger.Option(log.WithContextFieldsResolver(log.ContextLoggerFieldsResolver)); err != nil {
+if err := logger.Option(log.WithContextFieldsResolver(log.MergedContextLoggerFieldsResolver)); err != nil {
```
j4k4, line 78: `log.MergedContextLoggerFieldsResolver`  
Does this have to be renamed?  
At least I can't find it in the code.

j4k4 (changes requested):

j4k4 (approved):

PR #1076, opened 2024-01-05
lambda: changed logger initialization to get handlers from config

j4k4:
This change enables the option to set up the handlers of the logger via configuration.

### DIFF
--- a/pkg/lambda/lambda.go
+++ b/pkg/lambda/lambda.go
@@ -2,7 +2,6 @@ package lambda
 
 import (
 	"context"
-	"os"
 
 	awsLambda "github.com/aws/aws-lambda-go/lambda"
 	"github.com/justtrackio/gosoline/pkg/appctx"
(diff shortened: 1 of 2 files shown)


pkg/lambda/lambda.go
```diff
@@ -15,59 +14,65 @@ import (
...
-		logger.Error("can not apply post processor on config: %w", err)
+	if cfgPostProcessors, err = cfg.ApplyPostProcessors(config); err != nil {
+		defaultErrorHandler("can not apply post processor on config: %w", err)
 
-		os.Exit(1)
-	} else {
-		for name, priority := range cfgPostProcessors {
-			logger.Debug("applied priority %d config post processor '%s'", priority, name)
-		}
+		return
```
applike-ss, line 45: default error handler does os.Exit(1), so you don't need to return, do you?

pkg/lambda/lambda.go
```diff
@@ -15,59 +14,65 @@ import (
...
 
 	// create handler function and give lambda control
-	lambdaHandler, err := handlerFactory(ctx, config, logger)
-	if err != nil {
-		logger.Error("failed to create lambda handler: %w", err)
+	if lambdaHandler, err = handlerFactory(ctx, config, logger); err != nil {
+		defaultErrorHandler("failed to create lambda handler: %w", err)
 
-		os.Exit(1)
+		return
```
applike-ss, line 75: same as above, return might not be needed when defaultErrorHandler does os.Exit(1) already

pkg/lambda/lambda.go
j4k4: but if someone overwrites the default error handler with something which doesn't do an os.Exit, I still want to make sure, that it will not continue with the lambda start.

pkg/lambda/lambda.go
j4k4: same as above :)

pkg/lambda/lambda.go
```diff
@@ -15,59 +14,65 @@ import (
...
-
 	// configure and create config
 	mergedConfigOptions := append([]cfg.Option{
 		cfg.WithEnvKeyReplacer(cfg.DefaultEnvKeyReplacer),
 		cfg.WithSanitizers(cfg.TimeSanitizer),
 		cfg.WithErrorHandlers(func(msg string, args ...interface{}) {
-			logger.Error(msg, args...)
-			os.Exit(1)
+			defaultErrorHandler(msg, args...)
 		}),
```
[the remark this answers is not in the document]
j4k4: it would :)

applike-ss (approved):

PR #1077, opened 2024-01-10
kafka: make the kafka group-id configurable;

netsi:


pkg/kafka/consumer/settings.go
```diff
@@ -13,10 +13,12 @@ type Settings struct {
...
 
-	Topic string `cfg:"topic" validate:"required"`
-	// FQTopic is the fully-qualified topic name (with prefixes).
-	FQTopic      string
-	GroupID      string
+	Topic   string `cfg:"topic" validate:"required"`
+	GroupID string `cfg:"group_id"`
+	// FQTopic is the fully-qualified topic name (with prefix).
+	FQTopic string
+	// FQTopic is the fully-qualified group id (with prefix).
```
j4k4, line 20: In the comment it should be FQGroupID?

pkg/kafka/consumer/reader.go
```diff
@@ -44,7 +44,7 @@ func NewReader(
 
 		// Topic.
 		Topic:                 conf.FQTopic,
-		GroupID:               conf.GroupID,
+		GroupID:               conf.FQGroupID,
```
j4k4, line 47: I would like to rename the conf variable to settings, as we usually name it settings.

pkg/kafka/consumer/reader_test.go
```diff
@@ -14,9 +14,9 @@ import (
 var (
 	readerDialer = &kafka.Dialer{ClientID: "my-client"}
 	readerConf   = (&consumer.Settings{
```
j4k4, line 16: same here  
I know it's nitpicking but just for the consistency.. :)

j4k4 (changes requested):

pkg/kafka/consumer/reader_test.go
netsi: it makes sense

netsi:
@j4k4 I noticed that recently there was a change on the naming strategies, you think we need to adapt it also here?

netsi:
> > @j4k4 I noticed that recently there was a change on the naming strategies, you think we need to adapt it also here?
> 
> Yes, we tend to not use these naming strategies anymore which had to be done in code. Since around v0.8.0 we provide patterns via config. With this it's getting a lot easier to change the naming e.g. via env variables.. You can get an idea of this in `gosoline/pkg/cloud/aws/sqs/naming.go`. Would make sense to adjust it accordingly.. 👍

Adapted the naming strategy part to use the configs instead of code.

the config will be as such

```
kafka:
  naming:
    topic_pattern: "{env}-{topicId}"
    group_pattern: "{env}-{app}-{groupId}"
```

The default values were chosen in such a way to maintain compatibility, additionally the groupId naming still supports the legacy naming strategy where if the groupId is not configured it will simply return the `app` name, maybe I could also specify here a pattern instead

j4k4:
Looks fine for me for now.  
If you want, you could squash it and I will merge it afterwards.

j4k4 (approved):

netsi:
> Looks fine for me for now. If you want, you could squash it and I will merge it afterwards.

thank you, I squashed the commits, so you can merge when you'll get time

PR #1088, opened 2024-02-23
test: component repository and image tags are configurable now

j4k4:
## Configuration
Test components are docker container like wiremock or localstack which support you in your integration tests to mock
your applications infrastructure.

#### image configuration
The used image for a component can be configured on a global and on a per component level. The following configuration
snipped shows a config for two wiremock components. **mockA** gets the image configuration from the global level which is
configured via the key `test.defaults.images.wiremock` and **mockB** has local configuration. With this, **mockA** gets
tag `3.4.0` and **mockB** gets `3.3.0`.

```yaml
test:
  defaults:
    images:
      wiremock:
        repository: wiremock/wiremock
        tag: 3.4.0
  components:
    wiremock:
      mockA:
        expire_after: 5m
      mockB:
        image:
          repository: wiremock/wiremock
          tag: 3.3.0
        expire_after: 5m
```

### DIFF
--- a/pkg/test/env/config.default.yml
+++ b/pkg/test/env/config.default.yml
@@ -1,0 +1,21 @@
+test:
+  defaults:
+    images:
+      ddb:
+        repository: "amazon/dynamodb-local"
+        tag: "2.2.1"
+      localstack:
+        repository: "localstack/localstack"
+        tag: "3.1"
+      mysql:
+        repository: "mysql/mysql-server"
+        tag: "8.0"
+      redis:
+        repository: "redis"
+        tag: "7-alpine"
+      s3:
+        repository: "minio/minio"
+        tag: "RELEASE.2024-02-17T01-15-57Z"
+      wiremock:
+        repository: "wiremock/wiremock"
+        tag: "3.4.1"
--- a/pkg/test/env/environment.go
+++ b/pkg/test/env/environment.go
@@ -2,6 +2,7 @@ package env
 
 import (
 	"context"
+	_ "embed"
 	"fmt"
 	"testing"
 	"time"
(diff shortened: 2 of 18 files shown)


pkg/test/env/config.default.yml
```diff
@@ -0,0 +1,21 @@
...
+        tag: "8.0"
+      redis:
+        repository: "redis"
+        tag: "7-alpine"
+      s3:
+        repository: "minio/minio"
+        tag: "RELEASE.2023-03-09T23-16-13Z"
+      wiremock:
+        repository: "wiremock/wiremock"
+        tag: "3.3.1-alpine"
```
applike-ss, line 21: please change to non-alpine to ensure arm64 compatibility

pkg/test/env/environment.go
```diff
@@ -212,7 +233,7 @@ func (e *Environment) LoadFixtureBuilderFactories(factories ...fixtures.FixtureB
 			return fmt.Errorf("can not build fixture builder: %w", err)
 		}
 
-		if err = e.fixtureLoader.Load(e.ctx, fixtureBuilder.Fixtures()); err != nil {
+		if err := e.fixtureLoader.Load(e.ctx, fixtureBuilder.Fixtures()); err != nil {
```
applike-ss, line 236: why re-create this var? in line 229 you already have an err var defined.

applike-ss (changes requested):

pkg/test/env/environment.go
j4k4: sometimes the linter complains about these things due to: https://go-critic.com/overview.html#sloppyreassign  
you get something like:
```
sloppyReassign: re-assignment to `err` can be replaced with `err := env.runner.RunContainers(skeletons)` (gocritic)
```

pkg/test/env/config.default.yml
j4k4: changed to `3.4.1`

pkg/test/env/environment.go
applike-ss: strange, my goland often complains for these when i re-assign a var that exists already. But fine with me.

applike-ss (approved):

PR #1109, opened 2024-05-24
tracing: add support for  OTEL

netsi:
### Feature

* Add OTEL support to the `tracking` pkg.
* Instrument `http` and `grpcserver` packages

### Breaking Changes

xray `tracing` config structure has changed before:

```plaintext
tracing:
  provider: xray
  addr_type: local
  addr_value: ""
```

after:

```plaintext
tracing:
  provider: xray
  xray:
    addr_type: local
    addr_value: ""
```

### Usage examples

```plaintext
tracing:
  provider: otel
  otel:
    exporter: otel_http
    http:
      endpoint: localhost:4318
      url_path: /v1/traces
      compression: true
      insecure: true
      timeout: 10s
      retry:
        enabled: false
```


pkg/tracing/tracer_otel.go
```diff
@@ -0,0 +1,177 @@
...
+func (t *otelTracer) HttpClient(baseClient *http.Client) *http.Client {
+	return &http.Client{
+		Transport:     otelhttp.NewTransport(baseClient.Transport),
+		CheckRedirect: baseClient.CheckRedirect,
+		Jar:           baseClient.Jar,
+		Timeout:       baseClient.Timeout,
+	}
+}
+
+// GrpcUnaryServerInterceptor we still need to use the UnaryServerInterceptor because to maintain
```
ajscholl, line 146: This sentence reads broken

pkg/grpcserver/server.go
```diff
@@ -57,29 +59,42 @@ func New(name string, definer ServiceDefiner, middlewares ...MiddlewareFactory)
...
+		for _, m := range middlewares {
+			interceptors = append(interceptors,
+				grpc.UnaryServerInterceptor(m(logger)))
+		}
+
+		if tracer, err = tracing.ProvideTracer(ctx, config, logger); err != nil {
+			return nil, fmt.Errorf("can not create tracer: %w", err)
+		}
+
+		interceptors = append(interceptors, tracer.GrpcUnaryServerInterceptor())
```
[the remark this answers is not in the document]
netsi: moved it to the NewWithInterfaces

pkg/grpcserver/server.go
```diff
@@ -126,7 +141,9 @@ func (g *Server) Run(ctx context.Context) error {
 
 	err := g.server.Serve(g.listener)
 	if err != nil && !errors.Is(err, http.ErrServerClosed) {
-		g.logger.Error("grpc_server closed unexpected: %w", err.Error())
+		g.logger.WithFields(log.Fields{
```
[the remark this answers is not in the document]
netsi: We don't want to have logs with high cardinality, and want just the general message of what happened in the log message and have the details in a separate field in case we want them

pkg/tracing/tracer.go
```diff
@@ -8,39 +8,38 @@ import (
...
 }
 
 var tracerContainer = struct {
 	sync.Mutex
 	instance Tracer
 }{}
 
-func ProvideTracer(config cfg.Config, logger log.Logger) (Tracer, error) {
+func ProvideTracer(ctx context.Context, config cfg.Config, logger log.Logger) (Tracer, error) {
 	tracerContainer.Lock()
```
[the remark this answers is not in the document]
netsi: Makes sense, changed it now to use the appctx

pkg/tracing/tracer.go
```diff
@@ -8,39 +8,38 @@ import (
 
 	"github.com/justtrackio/gosoline/pkg/cfg"
 	"github.com/justtrackio/gosoline/pkg/log"
+	"google.golang.org/grpc"
 )
 
 //go:generate mockery --name Tracer
 type Tracer interface {
 	HttpHandler(h http.Handler) http.Handler
+	HttpClient(baseClient *http.Client) *http.Client
```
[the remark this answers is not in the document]
netsi: I split the interface, into the auto-instrumentation and manual instrumentation parts. Maybe you can check the naming and see if you agree

pkg/tracing/tracer_otel.go
netsi: Fixed it, basically I want to say here why we still use the deprecated function instead of the stas.Handler.

pkg/tracing/instrumentor_aws.go
```diff
@@ -0,0 +1,52 @@
...
+	return &awsInstrumentor{
+		AppId: appId,
+	}
+}
+
+func (t *awsInstrumentor) HttpHandler(h http.Handler) http.Handler {
+	name := fmt.Sprintf("%s-%s-%s-%s-%s", t.Project, t.Environment, t.Family, t.Group, t.Application)
+	handlerFunc := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		ctx := r.Context()
+		seg := xray.GetSegment(r.Context())
```
ajscholl, line 35:
```suggestion
		seg := xray.GetSegment(ctx)
```

pkg/tracing/instrumentor.go
```diff
@@ -0,0 +1,14 @@
+package tracing
+
+import (
+	"net/http"
+
+	"google.golang.org/grpc"
+)
+
+//go:generate mockery --name Instrumentor
+type Instrumentor interface {
```
ajscholl, line 10: Can we have some comments here what this is and when this is used? Or is this some basically internal interface? Looks like some kind of wrapper or middleware to me

pkg/tracing/instrumentor_otel.go
```diff
@@ -0,0 +1,68 @@
...
+	return &otelInstrumentor{
+		AppId: appId,
+	}
+}
+
+func (t *otelInstrumentor) HttpHandler(h http.Handler) http.Handler {
+	name := fmt.Sprintf("%v-%v-%v-%v", t.Project, t.Environment, t.Family, t.Application)
+	handlerFunc := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		ctx := r.Context()
+		span := trace.SpanFromContext(r.Context())
```
ajscholl, line 38:
```suggestion
		span := trace.SpanFromContext(ctx)
```

pkg/tracing/tracer.go
```diff
@@ -3,66 +3,68 @@ package tracing
...
+	})
+}
+
+func ProvideInstrumentor(ctx context.Context, config cfg.Config, logger log.Logger) (Instrumentor, error) {
+	return appctx.Provide(ctx, appCtxKey("instrumentor"), func() (Instrumentor, error) {
+		return NewInstrumentor(ctx, config, logger)
+	})
+}
+
+func NewTracer(ctx context.Context, config cfg.Config, logger log.Logger) (Tracer, error) {
```
ajscholl, line 38: Why is NewTracer and NewInstrumentor public when we have Provide* methods? Do we ever need more than one?

pkg/kafka/consumer/consumer_offset_manager_test.go
```diff
@@ -252,7 +252,7 @@ func TestOffsetManager_FetchMessageErrors(t *testing.T) {
 		readerErr = errors.New("reader: failed")
 	)
 
-	reader.On("FetchMessage", ctx).Return(
+	reader.On("FetchMessage", mock.AnythingOfType("context.backgroundCtx")).Return(
```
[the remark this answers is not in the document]
ajscholl: Should we revert these changes we still have in this MR then?

pkg/tracing/instrumentor_otel.go
netsi: :+1:

pkg/kafka/consumer/consumer_offset_manager_test.go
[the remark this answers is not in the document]
netsi: went through all the changes related to this and reverted them

ajscholl (approved):

pkg/tracing/tracer_noop.go
```diff
@@ -9,11 +9,13 @@ import (
...
+	AddTracerProvider(ProviderNoop, func(context.Context, cfg.Config, log.Logger) (Tracer, error) {
 		return NewNoopTracer(), nil
 	})
 }
 
+var _ Tracer = &noopTracer{}
+
 type noopTracer struct{}
 
 func NewNoopTracer() Tracer {
```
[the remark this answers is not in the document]
netsi: correct, missed that

pkg/tracing/tracer_aws.go
```diff
@@ -15,14 +14,23 @@ import (
...
 }
 
 const (
 	dnsSrv                        = "srv"
 	xrayDefaultMaxSubsegmentCount = 20
 )
 
+var _ Tracer = &awsTracer{}
+
+type XrayTracerSettings struct {
```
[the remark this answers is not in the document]
netsi: had no idea we had a module, but yeah this change instroduces a breaking change.
the config from
```
tracing:
  provider: xray
  addr_type: local
  addr_value: ""
```

becomes 
```
tracing:
  provider: xray
  xray:
    addr_type: local
    addr_value: ""
```

pkg/tracing/otel_exporters.go
```diff
@@ -0,0 +1,67 @@
...
+	MaxElapsedTime  time.Duration `cfg:"max_elapsed_time"`
+}
+
+type OtelExporterFactory func(ctx context.Context, config cfg.Config, logger log.Logger) (*otlptrace.Exporter, error)
+
+func AddTraceExporter(name string, exporter OtelExporterFactory) {
+	TraceExporters[name] = exporter
+}
+
+var TraceExporters = map[string]OtelExporterFactory{
```
ajscholl, line 35: I think this shouldn't be exported. Also, I think this should be called `otelTraceExporters` as it seems OTEL specific.

pkg/tracing/otel_exporters.go
```diff
@@ -0,0 +1,67 @@
...
+	Endpoint    string        `cfg:"endpoint" default:"localhost:4318"`
+	UrlPath     string        `cfg:"url_path" default:"/v1/traces"`
+	Compression bool          `cfg:"compression" default:"true"`
+	Insecure    bool          `cfg:"insecure" default:"false"`
+	Timeout     time.Duration `cfg:"timeout" default:"10s"`
+	Retry       RetryConfig   `cfg:"retry"`
+}
+
+type RetryConfig struct {
+	Enabled         bool          `cfg:"enabled" default:"false"`
```
ajscholl, line 23: Why are retries disabled by default? Sounds like having it enabled would be more robust

pkg/tracing/otel_exporters.go
```diff
@@ -0,0 +1,67 @@
...
+	InitialInterval time.Duration `cfg:"initial_interval"`
```
ajscholl, line 24: I would suggest providing defaults for these settings

pkg/tracing/otel_trace_provider.go
```diff
@@ -0,0 +1,76 @@
...
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/propagation"
+	"go.opentelemetry.io/otel/sdk/resource"
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+	semconv "go.opentelemetry.io/otel/semconv/v1.21.0"
+	"go.opentelemetry.io/otel/trace"
+)
+
+type OtelSettings struct {
+	Exporter      string  `cfg:"exporter"`
```
ajscholl, line 19: Does it make sense to provide the name of the default exporter as default as well?

pkg/tracing/otel_trace_provider.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+func newOtelTraceProvider(ctx context.Context, config cfg.Config, logger log.Logger) (trace.TracerProvider, error) {
+	appId := cfg.AppId{}
+	appId.PadFromConfig(config)
+
+	settings := &OtelSettings{}
+	config.UnmarshalKey("tracing.otel", settings)
+
+	otelExporterFactory := TraceExporters[settings.Exporter]
```
ajscholl, line 48: You should check that the exporter is actually a valid name and throw a proper error instead of crashing once you call the factory in the next code line

pkg/tracing/tracer_otel.go
```diff
@@ -0,0 +1,93 @@
...
+
+func (t *otelTracer) spanFromTrace(ctx context.Context, trc *Trace, name string) (context.Context, Span) {
+	var tFlags trace.TraceFlags
+	if trc.GetSampled() {
+		tFlags = trace.FlagsSampled
+	}
+
+	tID, err := trace.TraceIDFromHex(trc.GetTraceId())
+	if err != nil {
+		tID = trace.TraceID{}
```
ajscholl, line 78: So if we get an error, we get a parially filled object we have to zero out? Is this a behavior we should have a comment about?

pkg/tracing/otel_exporters.go
netsi: you are right, it is OTEL specific and yeah it doesn't have to be exported because if someone wants to add their implementation they can use the `AddTraceExporter` func (which I'll also rename to AddOtelTraceExporter)

pkg/tracing/tracer_otel.go
netsi: added a comment and also a warning log if that happens, under normal circumstances this should never happen

pkg/tracing/otel_trace_provider.go
netsi: added a default value

pkg/tracing/tracer_otel.go
```diff
@@ -73,12 +77,19 @@ func (t *otelTracer) spanFromTrace(ctx context.Context, trc *Trace, name string)
 		tFlags = trace.FlagsSampled
 	}
 
+	// The Trace ID is expected to be compliant with the W3C trace-context specification. If it is not
+	// an empty traceID will be used for the new span.
 	tID, err := trace.TraceIDFromHex(trc.GetTraceId())
 	if err != nil {
+		t.logger.Warn("could not parse trace id %w", err)
```
ajscholl, line 84: `%w` doesn't work when logging warnings, you want to use `%s` in this case (using `%w` results in something like `%!w(*errors.errorString=&{your error message}) ` because the arguments are passed to `fmt.Sprintf` and not `fmt.Errorf` for non-error levels)

ajscholl (approved):

PR #1112, opened 2024-06-04
cloud/aws: added servicediscovery client

ajscholl:


pkg/cloud/aws/metadata/provider.go
```diff
@@ -0,0 +1,175 @@
...
+	PathSystem                                               = "system"
+)
+
+// A Provider gives you convenient access to the metadata of your EC2 instance.
+// See also https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-retrieval.html.
+type Provider interface {
+	GetMetadata(ctx context.Context, path string) (string, error)
+}
+
+type NotAvailableError struct{}
```
bt-justtrack, line 75: Why have it as an empty struct and not do:

```go
var ErrNotAvailable = errors.New("metadata not available")
```

pkg/cloud/aws/metadata/provider.go
```diff
@@ -0,0 +1,175 @@
...
+
+	if res.StatusCode != netHttp.StatusOK {
+		return "", fmt.Errorf("unexpectd status %d", res.StatusCode)
+	}
+
+	return string(res.Body), nil
+}
+
+func (m *provider) getToken(ctx context.Context) (string, error) {
+	m.lck.Lock()
```
bt-justtrack, line 151: Nit: sync.RWMutex with early return would be more elegant

pkg/cloud/aws/metadata/provider.go
```diff
@@ -0,0 +1,175 @@
+package metadata
```
bt-justtrack, line 1: If i understand it correctly this is ec2 specific metadata ?
If so i would add another package layer so its /pkg/cloud/aws/metadata/ec2 (might be that we add such a provider aswell for elastic kubernetes service

pkg/cloud/aws/metadata/provider.go
ajscholl: I disagree. For the simple case that we already have a token, locking just a normal mutex is enough. There are no concurrent modifications happening until our token expires.

However, if our token is expired / was never set before, we now have really ugly logic: We now need to RLock, then check if the token is acceptable, if not, we have to RUnlock, Lock, check the token again (because we can't upgrade our lock and someone else could've fetched it after we RUnlock'ed) and then do the fetch (unless we want multiple fetches at the same time, which I would prefer to avoid).

All these problems disappear by using a normal mutex - because we have no place where we just read the token, we always have the chance to write it as well (because there is just a single place we need the lock for).

bt-justtrack: Fair argument, conversation resolved :+1:

bt-justtrack (approved):

PR #1113, opened 2024-06-06
conc: disabled task runner by default;

selm0:
When having the task runner enabled, it adds a foreground module to the app. So every application, which adds additional ones, will have troubles shutting down for themselves, as there is always a second foreground module for which the kernel is waiting.

This change disables the task runner by default, so you have to be aware about the implications.

### DIFF
--- a/pkg/conc/task_runner/task_runner.go
+++ b/pkg/conc/task_runner/task_runner.go
@@ -18,6 +18,8 @@ import (
 //
 // Try to avoid using this module if possible. It is only meant as a last resort.
 //
+// Enable it by setting task_runner.enabled = true
+//
 //go:generate mockery --name TaskRunner
 type TaskRunner interface {
 	kernel.Module
@@ -25,10 +27,11 @@ type TaskRunner interface {
 }
 
 type Settings struct {
-	Enabled bool `cfg:"enabled" default:"true"`
+	Enabled bool `cfg:"enabled" default:"false"`
 }
 
 type taskRunner struct {
+	kernel.BackgroundModule
 	lck          sync.Mutex
 	done         bool
 	pendingTasks chan kernel.Module


ajscholl:
Wouldn't the correct solution be to make it a background module? So it doesn't keep the app running?

selm0:
> Wouldn't the correct solution be to make it a background module? So it doesn't keep the app running?

I'm not sure about the use cases; will there always be a foreground module?
What if that foreground module  is spawning tasks, but processing takes longer than spawning -> you end up with the spawner exiting and the task runner shutting down, right?

ajscholl:
> > Wouldn't the correct solution be to make it a background module? So it doesn't keep the app running?
> 
> I'm not sure about the use cases; will there always be a foreground module? What if that foreground module is spawning tasks, but processing takes longer than spawning -> you end up with the spawner exiting and the task runner shutting down, right?

Correct. Also using the task runner in this case is incorrect, the correct abstraction would be to create a coffin and wait for that. The task runner is intended for running background tasks where you can't really put them anywhere else, but are part of some other module (lets say you implement access to some service - in this case, calls to your service wrapper shouldn't return before you received a response).

We can make it opt-in instead of opt-out as well, but it definitely needs to be a background module. Otherwise you can never use it in an application which exits after it is done with some amount of work.

selm0:
> > > Wouldn't the correct solution be to make it a background module? So it doesn't keep the app running?
> > 
> > 
> > I'm not sure about the use cases; will there always be a foreground module? What if that foreground module is spawning tasks, but processing takes longer than spawning -> you end up with the spawner exiting and the task runner shutting down, right?
> 
> Correct. Also using the task runner in this case is incorrect, the correct abstraction would be to create a coffin and wait for that. The task runner is intended for running background tasks where you can't really put them anywhere else, but are part of some other module (lets say you implement access to some service - in this case, calls to your service wrapper shouldn't return before you received a response).
> 
> We can make it opt-in instead of opt-out as well, but it definitely needs to be a background module. Otherwise you can never use it in an application which exits after it is done with some amount of work.

Okay, I'll add the background module as well

selm0:
@ajscholl updated

ajscholl (approved):

PR #1116, opened 2024-06-18
small db improvement

j4k4:
This PR adds several improvements to the db package:

## Sqler and ExecMultiInTx
We added a convenient function to execute multiple sql statements in one transaction:
```golang
sqls := []db.Sqler{
	squirrel.Update("`schemas`").Set("is_default", false),
	squirrel.Insert("`schemas`").Options("IGNORE").Columns("name", "is_default").Values(name, true),
	squirrel.Update("`schemas`").Set("is_default", true).Where(squirrel.Eq{"name": name}),
}

if _, err = client.ExecMultiInTx(ctx, sqls...); err != nil {
	return fmt.Errorf("can not execute queries: %w", err)
}
```	

## MySQL logging
We've added the gosoline logger to the MySQL driver so that critical errors get not printed to `os.Stderr` but instead getting logged as an error.

## Settings
We've added a couple of settings for the db connection. The following yaml shows how to set them. The example shows the default values:
```yaml
db:
  default:
    charset: utf8mb4
    collation: utf8mb4_general_ci
    connection_max_idletime: 120s
    timeouts:
      readTimeout: 0
      writeTimeout: 0
      timeout: 0 # 0 means that the os default will get used
```

### DIFF
--- a/pkg/db/client.go
+++ b/pkg/db/client.go
@@ -24,6 +24,28 @@ type SqlResult interface {
 	RowsAffected() (int64, error)
 }
 
+type Sqler interface {
+	ToSql() (string, []interface{}, error)
+}
+
+func SqlFmt(format string, a ...any) Sqler {
+	return sqlFmt{
+		format: format,
+		a:      a,
+	}
+}
+
+type sqlFmt struct {
+	format string
+	a      []any
+}
+
+func (s sqlFmt) ToSql() (qry string, args []interface{}, err error) {
+	qry = fmt.Sprintf(s.format, s.a...)
+
+	return
+}
+
 type (
 	ResultRow map[string]string
 	Result    []ResultRow
@@ -34,6 +56,7 @@ type Client interface {
 	GetSingleScalarValue(ctx context.Context, query string, args ...interface{}) (int, error)
 	GetResult(ctx context.Context, query string, args ...interface{}) (*Result, error)
 	Exec(ctx context.Context, query string, args ...interface{}) (sql.Result, error)
+	ExecMultiInTx(ctx context.Context, sqlers ...Sqler) (results []sql.Result, err error)
 	Prepare(ctx context.Context, query string) (*sql.Stmt, error)
 	Query(ctx context.Context, query string, args ...interface{}) (*sql.Rows, error)
 	Queryx(ctx context.Context, query string, args ...interface{}) (*sqlx.Rows, error)
@@ -94,7 +117,11 @@ func (c *ClientSqlx) GetResult(ctx context.Context, query string, args ...interf
 		return nil, err
 	}
 
-	cols, _ := rows.Columns()
+	cols, err := rows.Columns()
+	if err != nil {
+		return nil, fmt.Errorf("can not get columns: %w", err)
+	}
+
 	types := make(map[string]string)
 
 	for rows.Next() {
@@ -130,6 +157,7 @@ func (c *ClientSqlx) GetResult(ctx context.Context, query string, args ...interf
 				m[colName] = strconv.FormatFloat((*val).(float64), 'f', -1, 64)
 			default:
 				errStr := fmt.Sprintf("could not convert mysql result into string map: %v -> %v is %v", colName, *val, reflect.TypeOf(*val))
+
 				return nil, errors.New(errStr)
 			}
 		}
@@ -144,6 +172,56 @@ func (c *ClientSqlx) Exec(ctx context.Context, query string, args ...interface{}
 	c.logger.Debug("> %s %q", query, args)
 
 	return c.db.ExecContext(ctx, query, args...)
+}
+
+func (c *ClientSqlx) ExecMultiInTx(ctx context.Context, sqlers ...Sqler) (results []sql.Result, err error) {
+	var tx *sql.Tx
+	var res sql.Result
+	var queries []string
+	var argss [][]interface{}
+
+	for i, sqler := range sqlers {
+		var buildErr error
+		var qry string
+		var args []interface{}
+
+		if qry, args, buildErr = sqler.ToSql(); buildErr != nil {
+			return nil, fmt.Errorf("can not build sql #%d: %w", i, err)
+		}
+
+		queries = append(queries, qry)
+		argss = append(argss, args)
+	}
+
+	if tx, err = c.BeginTx(ctx, &sql.TxOptions{}); err != nil {
+		return nil, fmt.Errorf("can not begin transaction: %w", err)
+	}
+
+	defer func() {
+		if err == nil {
+			return
+		}
+
+		if errRollback := tx.Rollback(); errRollback != nil {
+			err = multierror.Append(err, fmt.Errorf("can not rollback tx: %w", errRollback)).ErrorOrNil()
+
+			return
+		}
+	}()
+
+	for i, qry := range queries {
+		if res, err = c.Exec(ctx, qry, argss[i]...); err != nil {
+			return nil, fmt.Errorf("can not exec qry %s: %w", qry, err)
+		}
+
+		results = append(results, res)
+	}
+
+	if err = tx.Commit(); err != nil {
+		return nil, fmt.Errorf("can not commit transaction: %w", err)
+	}
+
+	return
 }
 
 func (c *ClientSqlx) Prepare(ctx context.Context, query string) (*sql.Stmt, error) {
@@ -195,7 +273,8 @@ func (c *ClientSqlx) WithTx(ctx context.Context, ops *sql.TxOptions, do func(ctx
 		if err != nil {
 			errRollback := tx.Rollback()
 			if errRollback != nil {
-				err = multierror.Append(err, fmt.Errorf("can not roolback tx: %w", errRollback))
+				err = multierror.Append(err, fmt.Errorf("can not rollback tx: %w", errRollback))
+
 				return
 			}
 			c.logger.WithContext(ctx).Debug("rollback successfully done")
--- a/pkg/db/driver_factory.go
+++ b/pkg/db/driver_factory.go
@@ -5,21 +5,31 @@ import (
 	"fmt"
 
 	"github.com/golang-migrate/migrate/v4/database"
+	"github.com/justtrackio/gosoline/pkg/log"
 )
 
-type DriverFactory interface {
+type DriverFactory func(logger log.Logger) (Driver, error)
+
+type Driver interface {
 	GetDSN(settings Settings) string
 	GetMigrationDriver(db *sql.DB, database string, migrationsTable string) (database.Driver, error)
 }
 
 var connectionFactories = map[string]DriverFactory{}
 
-func GetDriverFactory(driverName string) (DriverFactory, error) {
-	factory, exists := connectionFactories[driverName]
+func GetDriver(logger log.Logger, driverName string) (Driver, error) {
+	var ok bool
+	var err error
+	var factory DriverFactory
+	var driver Driver
 
-	if !exists {
+	if factory, ok = connectionFactories[driverName]; !ok {
 		return nil, fmt.Errorf("no driver factory defined for %s", driverName)
 	}
 
-	return factory, nil
+	if driver, err = factory(logger); err != nil {
+		return nil, fmt.Errorf("failed to get driver for %s: %w", driverName, err)
+	}
+
+	return driver, nil
 }
--- a/pkg/db/driver_mysql.go
+++ b/pkg/db/driver_mysql.go
@@ -3,47 +3,70 @@ package db
 import (
 	"database/sql"
 	"fmt"
-	"net/url"
+	"net"
 	"strconv"
 
-	_ "github.com/go-sql-driver/mysql"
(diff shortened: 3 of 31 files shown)


pkg/db/client.go
```diff
@@ -146,6 +169,57 @@ func (c *ClientSqlx) Exec(ctx context.Context, query string, args ...interface{}
 	return c.db.ExecContext(ctx, query, args...)
 }
 
+func (c *ClientSqlx) ExecMultiInTx(ctx context.Context, sqllers ...Sqller) (results []sql.Result, err error) {
```
applike-ss, line 172: should we allow to pass an optional `*sql.TxOptions`?

pkg/db/client.go
```diff
@@ -24,6 +24,28 @@ type SqlResult interface {
 	RowsAffected() (int64, error)
 }
 
+type Sqller interface {
```
applike-ss, line 27: Should this be a `Sqler`?

pkg/db/client.go
```diff
@@ -146,6 +169,57 @@ func (c *ClientSqlx) Exec(ctx context.Context, query string, args ...interface{}
...
+		return
+	}
+
+	defer func() {
+		if err == nil {
+			return
+		}
+
+		if errRollback := tx.Rollback(); errRollback != nil {
+			err = multierror.Append(err, fmt.Errorf("can not roolback tx: %w", errRollback))
```
applike-ss, line 201: Do we need to do it like this instead? `err = multierror.Append(err, fmt.Errorf("can not roolback tx: %w", errRollback)).ErrorOrNil()`

pkg/db/client.go
```diff
@@ -146,6 +169,57 @@ func (c *ClientSqlx) Exec(ctx context.Context, query string, args ...interface{}
...
+			return
```
applike-ss, line 202: i'm not entirely sure how the coding style might have evolved over the last months/years, but should there be am empty line before this return? Same for line 209 and 217.

pkg/db/driver_mysql.go
```diff
@@ -24,25 +23,21 @@ func NewMysqlDriverFactory() DriverFactory {
...
-	return uri[2:]
+	cfg := mysql.NewConfig()
+	cfg.User = settings.Uri.User
+	cfg.Passwd = settings.Uri.Password
+	cfg.Net = "tcp"
+	cfg.Addr = net.JoinHostPort(settings.Uri.Host, strconv.Itoa(settings.Uri.Port))
+	cfg.DBName = settings.Uri.Database
+	cfg.MultiStatements = true
+	cfg.ParseTime = true
+	cfg.Collation = "utf8mb4"
```
applike-ss, line 34: Should we have a `mysql.NewConfigWithSettings` func to move the logic to a location where it belongs?

pkg/db/driver_mysql.go
```diff
@@ -24,25 +23,21 @@ func NewMysqlDriverFactory() DriverFactory {
...
+
+	return cfg.FormatDSN()
 }
 
 func (m *mysqlDriverFactory) GetMigrationDriver(db *sql.DB, database string, migrationsTable string) (database.Driver, error) {
-	return mysql.WithInstance(db, &mysql.Config{
+	return migrateMysql.WithInstance(db, &migrateMysql.Config{
```
applike-ss, line 40: interesting, i didn't realize we are still using the golang migrate library here when i reviewed the goose db changes. Should this change as well?

applike-ss: In fact this whole GetMigrationDriver func does only seem to be used here: [pkg/db/migrations_golang_migrate.go](https://github.com/justtrackio/gosoline/blob/c6717629ecc235748824b1f4735e18ff368605b4/pkg/db/migrations_golang_migrate.go#L31)

Do we want to deprecate it or re/move it immediately?

applike-ss (changes requested):

pkg/db/client.go
j4k4: ya, I guess in 99% of the cases it is not needed so I would like to keep the signature simple for now

pkg/db/client.go
j4k4: renamed

pkg/db/client.go
j4k4: applied

pkg/db/client.go
j4k4: changed

pkg/db/driver_mysql.go
j4k4: ya, could be that this could be improved.. but is a little bit out of scope for me now..

pkg/db/client.go
```diff
@@ -24,6 +24,28 @@ type SqlResult interface {
 	RowsAffected() (int64, error)
 }
 
+type Sqler interface {
+	ToSql() (string, []interface{}, error)
+}
+
+func SqllerFmt(format string, a ...any) Sqler {
```
applike-ss, line 31: should this be `Sqler` as well?

pkg/db/client.go
```diff
@@ -24,6 +24,28 @@ type SqlResult interface {
...
+	return sqllerFmt{
+		format: format,
+		a:      a,
+	}
+}
+
+type sqllerFmt struct {
```
applike-ss, line 38: should this be `sqlerFmt ` or maybe rather `sqlFmt`?

pkg/db/client.go
```diff
@@ -146,6 +169,54 @@ func (c *ClientSqlx) Exec(ctx context.Context, query string, args ...interface{}
 	return c.db.ExecContext(ctx, query, args...)
 }
 
+func (c *ClientSqlx) ExecMultiInTx(ctx context.Context, sqllers ...Sqler) (results []sql.Result, err error) {
```
applike-ss, line 172: In the interface you have called in sqlers, so i assume it should be the same on implementations.

```suggestion
func (c *ClientSqlx) ExecMultiInTx(ctx context.Context, sqlers ...Sqler) (results []sql.Result, err error) {
```

pkg/db/driver_mysql.go
```diff
@@ -24,25 +23,21 @@ func NewMysqlDriverFactory() DriverFactory {
...
-
-	return uri[2:]
+	cfg := mysql.NewConfig()
+	cfg.User = settings.Uri.User
+	cfg.Passwd = settings.Uri.Password
+	cfg.Net = "tcp"
+	cfg.Addr = net.JoinHostPort(settings.Uri.Host, strconv.Itoa(settings.Uri.Port))
+	cfg.DBName = settings.Uri.Database
+	cfg.MultiStatements = true
+	cfg.ParseTime = true
```
applike-ss, line 33: do we want to get this from settings as well (like before) via `settings.ParseTime`?

pkg/db/driver_mysql.go
```diff
@@ -24,25 +23,21 @@ func NewMysqlDriverFactory() DriverFactory {
...
+	cfg.Collation = "utf8mb4_general_ci"
```
applike-ss, line 34: should we make this part of the settings too, so the user can configure and overwrite it?

applike-ss: Also we have been setting a charset before, now we are setting a collation. According to the release notes of go-sql-driver/mysql 1.8.0 we are staying compatible when not specifying a Collation or Charset at all.

pkg/db/driver_mysql.go
```diff
@@ -2,13 +2,12 @@ package db
 
 import (
 	"database/sql"
-	"fmt"
-	"net/url"
+	"net"
 	"strconv"
 
-	_ "github.com/go-sql-driver/mysql"
+	"github.com/go-sql-driver/mysql"
```
applike-ss, line 8: Can you also update it to 1.8.1? It fixes a race condition with cancelled contexts: https://github.com/go-sql-driver/mysql/releases/tag/v1.8.1

applike-ss (changes requested):

pkg/db/client.go
j4k4: going with `SqlFmt`..

pkg/db/client.go
j4k4: going with `sqlFmt`..

pkg/db/client.go
j4k4: :+1:

pkg/db/driver_mysql.go
j4k4: :+1:

pkg/db/client.go
```diff
@@ -24,6 +24,28 @@ type SqlResult interface {
...
+type sqlFmt struct {
+	format string
+	a      []any
+}
+
+func (s sqlFmt) ToSql() (qry string, args []interface{}, err error) {
+	qry = fmt.Sprintf(s.format, s.a...)
+
+	return
+}
```
applike-ss, line 47: depending on how nit-picky you like it, you could change all `interface{}` => `any`. Makes it nicer readable, but is optional.

From the funcs name it makes sense to parse the format with the args in `ToSql`. However there is something odd here: You always return an empty args slice as the second return value. Additionally parsing it like this poses to be a security risk when it comes to sql injections imho.

pkg/db/driver_factory.go
```diff
@@ -5,21 +5,30 @@ import (
...
+	var driver Driver
 
-	if !exists {
+	if factory, ok = connectionFactories[driverName]; !ok {
 		return nil, fmt.Errorf("no driver factory defined for %s", driverName)
 	}
 
-	return factory, nil
+	if driver, err = factory(logger); err != nil {
+	}
```
applike-ss, line 31: do we want to return an error here?

pkg/db/driver_mysql.go
applike-ss: pls still update the mysql driver

pkg/db/driver_mysql.go
```diff
@@ -3,47 +3,59 @@ package db
...
-	connectionFactories[DriverMysql] = NewMysqlDriverFactory()
+	connectionFactories[DriverMysql] = NewMysqlDriver
 }
 
-func NewMysqlDriverFactory() DriverFactory {
-	return &mysqlDriverFactory{}
-}
+func NewMysqlDriver(logger log.Logger) (Driver, error) {
+	log := &mysqlLogger{logger: logger}
+	mysql.SetLogger(log)
```
applike-ss, line 23: In driver_cratedb.go you are calling this func still with Factory suffix. Should we have this unified across all drivers?

applike-ss, line 23:
```suggestion
func NewMysqlDriverFactory(logger log.Logger) (Driver, error) {
	sqlLogger := &mysqlLogger{logger: logger}
	
	err := mysql.SetLogger(sqlLogger)
	if err != nil {
	  return fmt.Errorf("failed to set mysql logger: %w", err)
	}

	return &mysqlDriver{}, nil
```

applike-ss (changes requested):

pkg/db/client.go
j4k4: This one is for a mix of string formatting and sql placeholder replacement..  
useful if one wants to replace things like the table name which is no possible for sql placeholders..

pkg/db/driver_factory.go
j4k4: :+1:

pkg/db/driver_mysql.go
j4k4: dropped the factory suffix

j4k4: :+1:

pkg/db/driver_mysql.go
j4k4: :+1:

pkg/db/client.go
```diff
@@ -146,6 +169,54 @@ func (c *ClientSqlx) Exec(ctx context.Context, query string, args ...interface{}
...
+		return nil, fmt.Errorf("can not begin transaction: %w", err)
+	}
+
+	defer func() {
+		if err == nil {
+			return
+		}
+
+		if errRollback := tx.Rollback(); errRollback != nil {
+			err = multierror.Append(err, fmt.Errorf("can not roolback tx: %w", errRollback)).ErrorOrNil()
```
bt-justtrack, line 200: roolback is typo

pkg/db/client.go
```diff
@@ -146,6 +169,54 @@ func (c *ClientSqlx) Exec(ctx context.Context, query string, args ...interface{}
...
+	var tx *sql.Tx
+	var res sql.Result
+	var buildErr error
+	var qry string
+	var queries []string
+	var args []interface{}
+	var argss [][]interface{}
+
+	for i, sqler := range sqlers {
+		if qry, args, buildErr = sqler.ToSql(); buildErr != nil {
```
bt-justtrack, line 182: tbh i don't see the point in predefining qry, args and buildErr as vars as they are only used within the the scope of this for loop and not outside aswell.

bt-justtrack (approved):

PR #1125, opened 2024-07-17
db: Implement generic wrapper type for json columns

bt-justtrack:
This new type wraps the underlying types stored in a jsonb column in databases, eliminating the need to have dedicated marshal / unmarshal boilerplate code on converting to / from the db model.

Example usage
```go
type DBModel struct {
    SomeModel db.JSON[SomeModel, db.NonNullable]
    SomeNullableModel db.JSON[*SomeModel, db.Nullable]
}

type SomeModel struct {
    Foo string
    Bar string
}

func NewDBModel(someModel SomeModel) DBModel {
    return DBModel{
        SomeModel: db.NewJSON(someModel, db.NonNullableBehaviour),
        SomeNullableModel: db.NewJSON((*SomeModel)(nil), db.NullableBehaviour),
    }
}

func (dbModel DBModel) ToSomeModel() SomeModel {
    return dbModel.SomeModel.Get()
}
```


pkg/db/jsonType.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+)
+
+func NewJSONType[T any](val T) JSONType[T] {
+	return JSONType[T]{
+		val: val,
+	}
+}
+
+func (t *JSONType[T]) AsJSONNull() JSONType[T] {
```
ajscholl, line 31: Do we need this to be a pointer receiver? Right now we could do
```go
jt.AsJSONNull()
```
as well as
```go
jt2 := jt1.AsJSONNull()
```
However, in this second case I could expect `jt1` to stay as it is as I am creating a new object with my change. Which would be wrong as `jt1` as well as `jt2` have the null flag set.
Finally, I think this won't work, right?
```go
jt := NewJSONType[t](getT()).AsJSONNull()
```

Finally, I guess a comment would be nice what this method changes (it turns a database null into the string `null` which is a JSON null, right?)

pkg/db/jsonType.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+func (t *JSONType[T]) Scan(src any) error {
+	if src == nil {
+		return nil
+	}
+
+	switch src := src.(type) {
+	case []byte:
+		return json.Unmarshal(src, &t.val)
```
ajscholl, line 56: Do I get this correct that we accept `NULL` from SQL as well as the string `null`, but the empty string is invalid? However, when I unmarshal the JSON `null` and my object is not a pointer, it will stay as it was initialized, right?

pkg/db/jsonType_test.go
```diff
@@ -0,0 +1,98 @@
...
+	"testing"
+
+	"github.com/justtrackio/gosoline/pkg/db"
+	"github.com/stretchr/testify/require"
+)
+
+func TestJSONType(t *testing.T) {
+	t.Parallel()
+
+	v := struct{}{}
```
ajscholl, line 14: I would maybe store something in the struct instead of using an empty struct. For example, doing nothing in `Scan` would be a correct implementation for this test

pkg/db/jsonType.go
```diff
@@ -0,0 +1,60 @@
...
+)
+
+var (
+	_                        driver.Valuer = JSONType[struct{}]{}
+	_                        sql.Scanner   = &JSONType[struct{}]{}
+	ErrJSONInvalidDriverType               = errors.New("invalid incoming driver type for json type")
+)
+
+type (
+	JSONType[T any] struct {
```
ajscholl, line 19: I am unsure about the name of the type - it is a JSON value after all, not a type. And as a value, it should behave like any other value, right? So if I want to use it in a nullable column, I use a pointer to it, not have a flag on the actual value (which btw doesn't survive getting stored in the database - `Scan` never sets it)

pkg/db/jsonType.go
```diff
@@ -0,0 +1,60 @@
...
+	"encoding/json"
+	"errors"
+
+	"github.com/justtrackio/gosoline/pkg/mdl"
+)
+
+var (
+	_                        driver.Valuer = JSONType[struct{}]{}
+	_                        sql.Scanner   = &JSONType[struct{}]{}
+	ErrJSONInvalidDriverType               = errors.New("invalid incoming driver type for json type")
```
ajscholl, line 15: What is a driver type? Sounds like the type of a driver to connect to SQL. And why is it invalid? What would've been a valid type?

ajscholl (changes requested):

pkg/db/jsonType.go
bt-justtrack: Yes that was my idea that both of the first 2 code snippets are valid.
If that is unintuitive I guess I would remove the pointer receiver and have it just as example 2.

Regarding snippet 3:
You cant do that but you would need to wrap the NewJSONType Call in a function like mdl.Box.

pkg/db/jsonType.go
bt-justtrack: For a SQL NULL we will produce the zero value for T, meaning if you want to preserve the semantic of that SQL NULL your T needs to be a pointer to your desired type.
The string null will also produce the zero value for T, however we only write this when configured on the writing part of JSONType.

The empty string as input to Value will produce `[]byte("\"\"")` as driver.Value. If you parse that again this is valid (even though afaik it is not valid json according to the json specs).

The empty string (as byte slice so `[]byte("")`) as input to Scan is indeed invalid since any empty file or file with only whitespace is not valid json.

ajscholl (approved):

PR #1129, opened 2024-07-23
fixtures: refactor fixture loading

applike-ss:


pkg/fixtures/writer_mysql_plain.go
```diff
@@ -50,50 +49,100 @@ func NewMysqlPlainFixtureWriterWithInterfaces(logger log.Logger, client db.Clien
...
+		Replace(m.metadata.TableName).
+		PlaceholderFormat(squirrel.Question).
+		Columns(cols...)
+	offset := 0
+	var stmt string
+	var args []any
+
+	for i, values := range fixtures {
+		// if we exceed the max amount of parameters for a prepared statement, make it a new statement instead
+		if ((i+1)*len(cols))-offset > 65536 {
```
j4k4, line 80: magic number? move it into a const?

pkg/httpserver/auth/anonymous.go
```diff
@@ -0,0 +1,28 @@
...
+
+func NewAnonymousAuthenticatorWithInterfaces() Authenticator {
+	return &anonymousAuthenticator{}
+}
+
+func (a *anonymousAuthenticator) IsValid(ginCtx *gin.Context) (bool, error) {
+	user := &Subject{
+		Name:            Anonymous,
+		Anonymous:       true,
+		AuthenticatedBy: ByApiKey,
```
j4k4, line 21: AuthenticatedBy = ByAnonymous?

pkg/fixtures/writer_mysql_plain.go
applike-ss: done that. It's the max amount of prepared statement arguments mysql allows you to use.

pkg/httpserver/auth/anonymous.go
applike-ss: adjusted, thanks!

test/db/migration-golang-migrate/mysql_test.go
```diff
@@ -34,32 +38,38 @@ func (s *MysqlTestSuite) SetupSuite() []suite.Option {
...
-			Writer: fixtures.MysqlPlainFixtureWriterFactory(&fixtures.MysqlPlainMetaData{
-				TableName: "mysql_plain_writer_test",
-				Columns:   []string{"id", "name"},
-			}),
-			Fixtures: []interface{}{
-				fixtures.MysqlPlainFixtureValues{2, "testName2"},
-				fixtures.MysqlPlainFixtureValues{2, "testName3"},
-			},
-		},
+func provideFixtureData(ctx context.Context, config cfg.Config, logger log.Logger) (fixtures.NamedFixtures[fixtures.MysqlPlainFixtureValues], fixtures.FixtureWriter, bool, bool, error) {
```
j4k4, line 56: can't this one directly return the FixtureSet?

applike-ss: fixed

test/fixtures/dynamodb/dynamodb_test.go
```diff
@@ -337,29 +317,66 @@ var ddbSettings = &ddb.Settings{
...
 
-func (s *DynamoDbSuite) dynamoDbDisabledPurgeFixtures() []*fixtures.FixtureSet {
-	return []*fixtures.FixtureSet{
-		{
-			Enabled: true,
-			Writer:  fixtures.DynamoDbFixtureWriterFactory(ddbSettings),
-			Fixtures: []interface{}{
-				&Person{Name: "Ash", Age: 10},
-			},
+func (s *DynamoDbSuite) dynamoDbFixtures1(ctx context.Context, config cfg.Config, logger log.Logger) (fixtures.NamedFixtures[*Person], fixtures.FixtureWriter, bool, error) {
```
j4k4, line 320: these could also just return the FixtureSet already?
Could it even look something like this?
```
func (s *DynamoDbSuite) dynamoDbFixtures(purge bool, data fixtures.NamedFixtures[*Person]) ([]fixtures.FixtureSet, error) {
	writer, err := fixtures.NewDynamoDbFixtureWriter(s.Env().Context(), s.Env().Config(), s.Env().Logger(), ddbSettings)
	if err != nil {
		return nil, fmt.Errorf("failed to create ddb fixture writer: %w", err)
	}
	
	return []fixtures.FixtureSet{fixtures.NewSimpleFixtureSet(data, writer, true, purge)}, nil
}
```

applike-ss: Pls check my implementation. Do you agree with it?

applike-ss:
Release notes draft:

**Gosoline Fixture loader refactoring changes:**

In this release I have refactored the fixture loading process to make it more flexible for custom implementations.

The `FixtureBuilderFactory` type was renamed to `FixtureSetsFactory` for clarity of what it is. Also its Signature changed, as can be seen below.

The `WithFixtureBuilderFactory` application option was renamed to `WithFixtureSetFactory` and gets a `FixtureSetsFactory` passed (`func(ctx context.Context, config cfg.Config, logger log.Logger) ([]FixtureSet, error)`).

The `FixtureSet` struct was renamed to `simpleFixtureSet` and should be used with its factory func `NewSimpleFixtureSet`.
`FixtureSet` is now an interface with only a `Write(ctx context.Context) error` func.
This makes the fixture loading more customizable, if you want to have your own implementation that is specific to your use-case (not covered by gosoline).

The `FixtureLoader` doesn't use a writer factory anymore and makes the user provide a writer (when using a `simpleFixtureSet`), also its interface doesn't get passed a `[]*fixtures.FixtureSet` anymore but `[]fixtures.FixtureSet` (slice of pointers to structs => slice of interfaces).

The `FixtureBuilder` interface was removed without replacement, as the `FixtureSetsFactory` is supposed to deliver the fixture sets directly.

The `FixtureWriterFactory` interface was removed without replacement, as the `simpleFixtureSet` contains a writer, when used. If a different `FixtureSet` implementation is used, it is supposed to take care of writing (or abstract it away).

The `FixtureWriter` now gets passed a `[]any` as fixtures instead of a `*FixtureSet`, as the implementations always only accessed the fixtures key of the original FixtureSet anyways.

The `SimpleFixtureBuilderFactory` was removed in favor of the `NewSimpleFixtureSet` factory func.

The `WithFixtureBuilderFactories` test suite option was renamed to `WithFixtureSetFactories` and accepts any amount of `FixtureSetsFactory` args.

The logic for checking whether a fixture set is enabled, should purge or write has been moved to the `SimpleFixtureSet`. Custom implementations may want to adopt the changes accordingly to stay compatible with prior application behavior.

The `NamedFixture` struct is now mostly Generic compatible, making it more type-safe and nicer to deal with due to proper typing.

The `Writer` factory funcs were replaced and their signatures changed:
* `BlobFixtureWriterFactory(settings *BlobFixturesSettings) FixtureWriterFactory` => `NewBlobFixtureWriter(ctx context.Context, config cfg.Config, logger log.Logger, settings *BlobFixturesSettings) (FixtureWriter, error)`
* `ConfigurableKvStoreFixtureWriterFactory[T any](name string) FixtureWriterFactory` => `NewConfigurableKvStoreFixtureWriter[T any](ctx context.Context, config cfg.Config, logger log.Logger, name string) (FixtureWriter, error)`
* `DynamoDbFixtureWriterFactory(settings *ddb.Settings, options ...DdbWriterOption) FixtureWriterFactory` => `NewDynamoDbFixtureWriter(ctx context.Context, config cfg.Config, logger log.Logger, settings *ddb.Settings, options ...DdbWriterOption) (FixtureWriter, error)`
* `DynamoDbKvStoreFixtureWriterFactory[T any](modelId *mdl.ModelId) FixtureWriterFactory` => `NewDynamoDbKvStoreFixtureWriter[T any](ctx context.Context, config cfg.Config, logger log.Logger, modelId *mdl.ModelId) (FixtureWriter, error)`
* `MysqlOrmFixtureWriterFactory(metadata *db_repo.Metadata) FixtureWriterFactory` => `NewMysqlOrmFixtureWriter(ctx context.Context, config cfg.Config, logger log.Logger, metadata *db_repo.Metadata) (FixtureWriter, error)`
* `MysqlPlainFixtureWriterFactory(metadata *MysqlPlainMetaData) FixtureWriterFactory` => `NewMysqlPlainFixtureWriter(ctx context.Context, config cfg.Config, logger log.Logger, metadata *MysqlPlainMetaData) (FixtureWriter, error)`
* `MysqlSqlxFixtureWriterFactory(metadata *MysqlSqlxMetaData) FixtureWriterFactory` => `NewMysqlSqlxFixtureWriter(ctx context.Context, config cfg.Config, logger log.Logger, metadata *MysqlSqlxMetaData) (FixtureWriter, error)`
* `RedisFixtureWriterFactory(name *string, operation *string) FixtureWriterFactory` => `NewRedisFixtureWriter(ctx context.Context, config cfg.Config, logger log.Logger, name string, operation string) (FixtureWriter, error)`
* `RedisKvStoreFixtureWriterFactory[T any](modelId *mdl.ModelId) FixtureWriterFactory` => `NewRedisKvStoreFixtureWriter[T any](ctx context.Context, config cfg.Config, logger log.Logger, modelId *mdl.ModelId) (FixtureWriter, error)`


The `MysqlPlainFixtureWriter` was adjusted to be able to load larger fixture amounts for the same table/model.

<h2>Please check out the examples to see how to migrate from original to new usage.</h2>

**Other notable changes:**
* Fixtures are now loaded **after** the change history (if the app option is given), to ensure you can also load a predefined history.
* The mysql container for integration tests now specifies max_connection=1000 (original default was 151 as per mysql documentation). This should allow for bigger integration tests and parallel integration tests running more smoothly.

* `db.SqlFmt` func Signature changed
  * includes a formatValues `[]any` parameter (to behave like fmt.* funcs) and `a` was renamed to `args` for better understanding of what the parameter is used for.
  * The `ToSql` receiver func first performs a fmt.Format on the passed format with the given `formatValues` (instead of `a` originally) and passes `args` back for later usage with e.g. sqlx for letting the db escape the values in a prepared statement to prevent potential sql injection

* There is now a `NewAnonymousAuthenticator` func in the `httpserver/auth` package to allow handlers to to database operations even when the user is not authenticated (yet)

* Redis clients are now being provided via the `appctx.Provide` func instead of via a `sync.Mutex` locking with local variable cache mechanism.

* `Mack` and `Suzy` were renamed to `Bob` and `Alice` as their parents did a mistake when filling out the birth papers.

j4k4 (approved):

PR #1136, opened 2024-07-25
db: enable custom parameters for connections

j4k4:
This PR adds two improvements to the `db` and `fixtures` package.

## DB connection parameters
This PR adds the possibility to add custom system variables to the db connection.  
In the case of a mysql connection, these parameters will translate into session variables.  
The following configuration would result in executing `SET foobar = 42;` on every new connection.

```yaml
db:
  default:
    driver: mysql
    parameters:
      foobar: 42
```

## SQLX fixtures
We've added a fixture writer to make it easier to load "db structs" as fixtures.  
You can find a full example under `gosoline/examples/fixtures/sqlx`.

```golang
type User struct {
	Id       int    `db:"id"`
	Name     string `db:"name"`
	IsActive bool   `db:"is_active"`
}

var fixtureSets = []*fixtures.FixtureSet{
	{
		Enabled: true,
		Purge:   true,
		Writer:  fixtures.MysqlSqlxFixtureWriterFactory(&fixtures.MysqlSqlxMetaData{TableName: "users"}),
		Fixtures: []interface{}{
			User{
				Id:       1,
				Name:     "Mack",
				IsActive: true,
			},
			User{
				Id:       2,
				Name:     "Suzy",
				IsActive: false,
			},
		},
	},
}
```


pkg/funk/slice.go
```diff
@@ -81,6 +81,16 @@ func Contains[T any](in []T, elem T) bool {
 	return ContainsFunc(in, equalTo)
 }
 
+func ContainsAll[T any](left []T, right []T) bool {
```
bt-justtrack, line 84: tbh I find it unintuitive that we check that left is a subset of right and not the otherway round.
Typically in the stdlib its the otherway round, you first specify the set and then the individual elements.
At least we would need to add a comment documenting if we deviate from that.

bt-justtrack (changes requested):

pkg/funk/slice.go
j4k4: adjusted

bt-justtrack (approved):

pkg/refl/tags.go
```diff
@@ -0,0 +1,35 @@
+package refl
+
+import (
+	"strings"
+)
+
+// GetTags reads the values of tags with the name tagName.
+// Tag values get extracted until the first colon.
```
bt-justtrack, line 8: until the first comma

bt-justtrack (approved):

PR #1142, opened 2024-08-20
metric: added calculator and handlers for requests / messages per runner

j4k4:
This PR introduces a more powerful version of the "messages per runner" metric from the stream package.\
The MPR metric was produced by a kernel module and used to perform auto scaling for SQS consumers.\
The whole functionality was implemented and limited to the stream package.\

The feature of doing metric calculations via a kernel module was moved to the `metric/calculator` package which implements a more generic approach to this. To achieve this, the calculator calls a set of provided handlers periodically. The handlers are returning the calculated metrics which should be published:
```golang
type Handler interface {
	GetMetrics(ctx context.Context) (metric.Data, error)
}
```
By default there are two registered handlers. The `stream_messages_per_runner` which is the replacement for the old MPR metric and the `httpserver_requests_per_runner` handler from the httpserver package which calculates the count of requests served by one instance of the server. 

Once enabled via `metric.calculator.enabled: true` the default settings will look similar to:
```yaml
metric:
    calculator:
        cloudwatch:
            client: "default"
        dynamodb:
            naming:
                pattern: '{env}-{modelId}'
        ecs:
            cluster: '{env}'
            service: '{app_group}-{app_name}'
        enabled: true
        handlers:
            httpserver_requests_per_runner:
                enabled: true
                max_increase_percent: 200
                max_increase_period: 5m0s
                period: 1m0s
                target_value: 100
            stream_messages_per_runner:
                enabled: true
                max_increase_percent: 200
                max_increase_period: 5m0s
                period: 1m0s
                target_value: 100
        leader_election: metric_calculator
        period: 1m0s
```


pkg/httpserver/middleware_metric.go
```diff
@@ -49,48 +49,50 @@ func metricMiddleware(name string, ginCtx *gin.Context, writer metric.Writer) {
...
+		"Method":     method,
+		"Path":       path,
+		"ServerName": name,
+	}, metric.Dimensions{
+		"ServerName": name,
+	}))
+}
+
+func createMetricsWithDimensions(metrics metric.Data, dimensions ...metric.Dimensions) metric.Data {
+	return funk.Flatten(funk.Map(metrics, func(metricDatum *metric.Datum) metric.Data {
```
ajscholl, line 84: Maybe we should introduce a FlatMap function at some point

ajscholl (approved):

bt-justtrack (approved):

PR #1144, opened 2024-08-29
currency: add health check to module;

selm0:
currency: add health check to module;
change-history: drop columns and recreate triggers;
all: update exp/slices to slices;

### DIFF
--- a/pkg/currency/module.go
+++ b/pkg/currency/module.go
@@ -3,6 +3,7 @@ package currency
 import (
 	"context"
 	"fmt"
+	"sync/atomic"
 	"time"
 
 	"github.com/justtrackio/gosoline/pkg/cfg"
@@ -11,11 +12,15 @@ import (
 )
 
 type Module struct {
-	kernel.BackgroundModule
+	kernel.EssentialBackgroundModule
 	kernel.ServiceStage
-	updaterService UpdaterService
 	logger         log.Logger
+	updaterService UpdaterService
+	healthy        *atomic.Bool
 }
+
+// ensure interface compatibility
+var _ kernel.HealthCheckedModule = Module{}
 
 func NewCurrencyModule() kernel.ModuleFactory {
 	return func(ctx context.Context, config cfg.Config, logger log.Logger) (kernel.Module, error) {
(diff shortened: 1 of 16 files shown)


pkg/currency/module.go
```diff
@@ -11,10 +12,12 @@ import (
 )
 
 type Module struct {
-	kernel.BackgroundModule
+	kernel.HealthCheckedModule
```
ajscholl, line 15: I don't think you want to embed the interface here. If you want to ensure the module implements the interface, use `var _ kernel.HealthCheckedModule = &Module{}`

pkg/currency/module.go
```diff
@@ -11,10 +12,12 @@ import (
...
+	kernel.EssentialBackgroundModule
 	kernel.ServiceStage
-	updaterService UpdaterService
 	logger         log.Logger
+	updaterService UpdaterService
+	healthy        *atomic.Bool
```
ajscholl, line 20: As the module is only used with a pointer to it, we can directly embed the bool and don't need a pointer indirection

pkg/currency/module.go
```diff
@@ -27,16 +30,29 @@ func NewCurrencyModule() kernel.ModuleFactory {
 		module := &Module{
 			logger:         logger,
 			updaterService: updater,
+			healthy:        &atomic.Bool{},
```
ajscholl, line 33: Which would make this line no longer needed

pkg/currency/module.go
```diff
@@ -27,16 +30,29 @@ func NewCurrencyModule() kernel.ModuleFactory {
...
 
+func (module *Module) IsHealthy(ctx context.Context) (bool, error) {
+	return module.healthy.Load(), nil
+}
+
 func (module *Module) Run(ctx context.Context) error {
-	ticker := time.NewTicker(time.Duration(1) * time.Hour)
+	defer func() {
+		module.healthy.Store(false)
+	}()
```
ajscholl, line 47: I think this can be simplified (as defer just needs an expression it can execute later):

```suggestion
	defer module.healthy.Store(false)
```

pkg/currency/module.go
```diff
@@ -27,16 +30,29 @@ func NewCurrencyModule() kernel.ModuleFactory {
...
+
+	// complete one refresh, then the module is healthy
```
ajscholl, line 49: Hmm, but refresh and importExchangeRates don't ensure the operation is actually successful, so if they throw an error, the module becomes healthy anyway

pkg/currency/module.go
```diff
@@ -27,16 +30,29 @@ func NewCurrencyModule() kernel.ModuleFactory {
...
 	module.refresh(ctx)
 	module.importExchangeRates(ctx)
+
+	module.healthy.Store(true)
+
+	ticker := time.NewTicker(time.Duration(1) * time.Hour)
```
ajscholl, line 55:
```suggestion
	ticker := time.NewTicker(time.Hour)
```

pkg/db-repo/change_history_manager.go
```diff
@@ -155,24 +155,19 @@ func (c *ChangeHistoryManager) executeMigration(originalTable, historyTable *tab
...
-	if updated {
+	} else if updated, statement := c.updateHistoryTable(historyTable); updated {
+		// if the table exists, we would "update" it with the same columns - skip it
 		statements = append(statements, statement)
-		statements = append(statements, c.dropHistoryTriggers(originalTable, historyTable)...)
-		statements = append(statements, c.createHistoryTriggers(originalTable, historyTable)...)
 		c.logger.Info("updating change history setup")
-		return c.execute(statements)
+	} else {
+		c.logger.Info("change history setup was already up to date")
```
ajscholl, line 164: Are you missing a `return nil` here? Or do we want to always drop and recreate the triggers? Wouldn't that run the risk that a task starting up drops the triggers, some other task does something, and we miss some update?

ajscholl (changes requested):

pkg/db-repo/change_history_manager.go
selm0: Indeed we want to always recreate the triggers.
The migration is run during the initial start up of the app, it won't be healthy without being complete. So it shouldn't be a problem.

pkg/db-repo/change_history_manager.go
```diff
@@ -13,15 +13,16 @@ import (
 	"github.com/justtrackio/gosoline/pkg/funk"
 	"github.com/justtrackio/gosoline/pkg/log"
 	"github.com/justtrackio/gosoline/pkg/mdl"
+	"golang.org/x/exp/slices"
```
ajscholl, line 16: I think the slices package is out of experimental and you can just import "slices" (might have a slightly different API though)

pkg/db-repo/change_history_manager.go
```diff
@@ -299,19 +287,97 @@ func (c *ChangeHistoryManager) rowUpdatedCondition(originalTable *tableMetadata)
...
+	}
+
+	// remove untracked columns - without this, the triggers will not work
+	dropColumns, err := c.buildDropColumns(originalTable.tableName, historyTable.tableName)
+	if err != nil {
+		return nil, fmt.Errorf("unable to fetch columns to be dropped: %w", err)
+	}
+
+	historyColumnNames := historyTable.columnNames()
+	dropColumns = funk.Filter(dropColumns, func(column string) bool {
```
ajscholl, line 312: Hmm, this is basically implementing `funk.Difference` by hand (and then just taking one of the two results). Maybe you want to go with that instead?

test/db-repo/change_history/schema/change_history_test.go
```diff
@@ -182,6 +182,7 @@ func (s *ChangeHistoryTestSuite) TestChangeHistoryMigration_Migrate_UpdateTable(
 	historyManager, err := db_repo.NewChangeHistoryManager(ctx, envConfig, envLogger)
 	s.NoError(err)
 
+	// this adds and drops on column each
```
ajscholl, line 185: The sentence seems unfinished

PR #1149, opened 2024-09-12
fixtures: remove pointer receiver from NamedFixtures and allow purging tables which don't exist (yet) in ddb

ajscholl:
When purging a table in ddb, we delete and re-create it. However, this currently fails if the table doesn't exist yet. As deleting a non-existing table achieves our goal (of the table no longer existing), we don't need to treat this as an error.

Additionally, removing the pointer receiver from `fixtures.NamedFixtures` allows for easier usage of the type without impacting performance too much (if at all).

### DIFF
--- a/pkg/fixtures/purger_dynamodb.go
+++ b/pkg/fixtures/purger_dynamodb.go
@@ -2,10 +2,12 @@ package fixtures
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/service/dynamodb"
+	"github.com/aws/aws-sdk-go-v2/service/dynamodb/types"
 	"github.com/justtrackio/gosoline/pkg/cfg"
 	gosoDynamodb "github.com/justtrackio/gosoline/pkg/cloud/aws/dynamodb"
 	"github.com/justtrackio/gosoline/pkg/ddb"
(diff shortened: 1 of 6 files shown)


pkg/fixtures/purger_dynamodb.go
```diff
@@ -40,6 +42,15 @@ func NewDynamodbPurger(ctx context.Context, config cfg.Config, logger log.Logger
 func (p *dynamodbPurger) Purge(ctx context.Context) error {
 	p.logger.Info("purging table %s", p.tableName)
 	_, err := p.client.DeleteTable(ctx, &dynamodb.DeleteTableInput{TableName: aws.String(p.tableName)})
+
+	var errResourceNotFoundException *types.ResourceNotFoundException
+	if errors.As(err, &errResourceNotFoundException) {
+		p.logger.Info("purging table %s skipped as it didn't exist", p.tableName)
```
applike-ss, line 48:
```suggestion
		p.logger.Info("purging table %s skipped as it does not exist", p.tableName)
```

pkg/fixtures/simple_fixture_set.go
```diff
@@ -47,3 +49,16 @@ func (c *simpleFixtureSet[T]) Write(ctx context.Context) error {
...
+
+	if c.Fixtures.Len() > 0 {
+		example = c.Fixtures[0].Value
+		if kvExample, ok := example.(*KvStoreFixture); ok {
+			example = kvExample.Value
+		}
+	}
+
+	return fmt.Sprintf("SimpleFixtureSet(len=%d, type=%T)", c.Fixtures.Len(), example)
+}
```
applike-ss, line 64:
```suggestion
func (c *simpleFixtureSet[T]) String() string {
	var model any = mdl.Empty[T]()

	if c.Fixtures.Len() > 0 {
		model = c.Fixtures[0].Value
		if kvModel, ok := model.(*KvStoreFixture); ok {
			model = kvModel.Value
		}
	}

	return fmt.Sprintf("%T(len=%d, type=%T)", c, c.Fixtures.Len(), model)
}
```

test/guard/fixtures_test.go
```diff
@@ -13,15 +13,53 @@ var metadata = &fixtures.MysqlPlainMetaData{
...
+			"a97b104f-4d93-4c15-a97a-4e3173e75cde",
+			`{
+				"description": "global - read and write access",
+				"effect": "allow",
+				"conditions": {},
+				"subjects": ["r:1"],
+				"resources": ["mrn:<.+>"],
+				"actions": ["<.+>"]
+			}`,
+		},
```
applike-ss, line 27: You dropped the id field, was this intentional?

applike-ss (changes requested):

test/guard/fixtures_test.go
ajscholl: no

applike-ss (approved):

PR #1160, opened 2024-10-10
fix: Implement restartability on producer daemon without reset;

bt-justtrack:
This PR fixes using the testing Suite.WithSharedEnvironment in scenarios where there is a producer daemon and more than one test case.

### DIFF
--- a/pkg/stream/output_channel.go
+++ b/pkg/stream/output_channel.go
@@ -10,6 +10,7 @@ type OutputChannel interface {
 	Read() ([]WritableMessage, bool)
 	Write(msg []WritableMessage)
 	Close()
+	IsClosed() bool
 }
 
 type outputChannel struct {
@@ -58,4 +59,11 @@ func (c *outputChannel) Close() {
 	} else {
 		c.logger.Warn("duplicate close to output channel: channel is already closed")
 	}
+}
+
+func (c *outputChannel) IsClosed() bool {
+	c.lck.RLock()
+	defer c.lck.RUnlock()
+
+	return c.closed
 }
--- a/pkg/stream/producer_daemon.go
+++ b/pkg/stream/producer_daemon.go
@@ -8,6 +8,7 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/justtrackio/gosoline/pkg/appctx"
 	"github.com/justtrackio/gosoline/pkg/cfg"
 	"github.com/justtrackio/gosoline/pkg/clock"
 	"github.com/justtrackio/gosoline/pkg/coffin"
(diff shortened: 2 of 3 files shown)


pkg/stream/output_channel.go
```diff
@@ -59,3 +60,7 @@ func (c *outputChannel) Close() {
 		c.logger.Warn("duplicate close to output channel: channel is already closed")
 	}
 }
+
+func (c outputChannel) IsClosed() bool {
```
ajscholl, line 64: As the linter says, this must be a pointer receiver

pkg/stream/output_channel.go
```diff
@@ -59,3 +60,7 @@ func (c *outputChannel) Close() {
...
+	return c.closed
```
ajscholl, line 65: you are not allowed to read closed without taking the lock for reading (otherwise you have a race and the race detector might scream)

pkg/stream/producer_daemon.go
```diff
@@ -31,23 +31,23 @@ var (
 )
 
 type ProducerDaemonSettings struct {
-	Enabled bool `cfg:"enabled" default:"false"`
+	Enabled bool `cfg:"enabled"                default:"false"`
```
ajscholl, line 34: Why are we inserting random spaces here? I don't think it makes sense to align them when nothing else is aligned because of the comments

pkg/stream/producer_daemon.go
```diff
@@ -94,13 +94,6 @@ type producerDaemon struct {
 	stopped    int32
 }
 
-func ResetProducerDaemons() {
```
ajscholl, line 97: If we are removing these old locks, should we maybe switch the complete architecture to provide producer daemons from the context instead of using these global locks?

pkg/stream/producer_daemon.go
```diff
@@ -175,11 +176,15 @@ func (d *producerDaemon) GetStage() int {
 }
 
 func (d *producerDaemon) Run(kernelCtx context.Context) error {
+	atomic.StoreInt32(&d.stopped, 0)
 	defer func() { atomic.StoreInt32(&d.stopped, 1) }()
 
 	// ensure we don't have a race with the code in Write checking if the ticker is nil
 	d.lck.Lock()
 	d.ticker = d.clock.NewTicker(d.settings.Interval)
+	if d.outCh.IsClosed() {
```
ajscholl, line 185: not sure if this belongs here, but ideally we can get rid of it anyways because it isn't sound - we can get a call to write before our module starts, which would cause us to drop a message (because the output channel is still closed)

ajscholl (changes requested):

pkg/stream/producer_daemon.go
bt-justtrack: I guess that was auto aligned that way because of a missing new line after each property before the next comment, so it was recognized as one block of properties. Fixed that by adding new lines.

pkg/stream/producer_daemon.go
bt-justtrack: Switched to appctx

pkg/stream/producer_daemon.go
```diff
@@ -30,96 +31,104 @@ var (
 	producerDaemons    = map[string]*producerDaemon{}
```
ajscholl, line 31: Remove this

pkg/stream/producer_daemon.go
```diff
@@ -30,96 +31,104 @@ var (
...
+	})
+	if err != nil {
+		return nil, fmt.Errorf("retrieving daemon from appctx: %w", err)
 	}
 
-	var err error
-	producerDaemons[name], err = NewProducerDaemon(ctx, config, logger, name)
-	if err != nil {
-		return nil, err
+	if producer.IsStopped() {
```
ajscholl, line 123: I would not make IsStopped and Restart public - user code shouldn't need to query them or reset them, if they have to, gosoline itself needs to be improved to cover this use-case

pkg/stream/producer_daemon.go
```diff
@@ -30,96 +31,104 @@ var (
...
+		producer.Restart()
 	}
 
-	return producerDaemons[name], nil
+	return producer, err
```
ajscholl, line 127: err is known to be nil here

pkg/stream/producer_daemon.go
```diff
@@ -170,6 +187,23 @@ func NewProducerDaemonWithInterfaces(logger log.Logger, metric metric.Writer, ag
...
+
+func (d *producerDaemon) Restart() {
+	d.lck.Lock()
+	defer d.lck.Unlock()
+
+	if d.outCh.IsClosed() {
+		d.outCh = NewOutputChannel(d.logger, d.settings.BufferSize)
+	}
+
+	if atomic.LoadInt32(&d.stopped) != 0 {
```
ajscholl, line 202: You don't need to check for != 0 here - writing a 0 to a variable holding 0 is fine.

While it isn't a problem here, this pattern additionally just feels like a race condition as you have a race between the check and the actual use later on

ajscholl (approved):

PR #1161, opened 2024-10-10
feat: Implement atomic mutates on the cache;

bt-justtrack:
This allows mutating the cache in a more in memory kvstore style where other workers cannot interfer with a mutation between reading and setting the mutated data.


pkg/cache/ccache.go
```diff
@@ -15,6 +15,8 @@ type Cache[T any] interface {
 	Get(key string) (T, bool)
 	Set(key string, value T)
 	SetX(key string, value T, ttl time.Duration)
+	Update(key string, update func(value T) T, provider func() T) T
```
ajscholl, line 18: I am unsure if this is the best API design - what do you think about this?

```suggestion
	Map(key string, mapper func(value *T) T) T
```

The idea is to pass nil in case there is no value yet and otherwise a reference to the value. for one, it reduces the number of functions by one, and second it clears up whether provide should already provide the mapped value or (in this case) provides the base value we then update.

Additionally, we should maybe add some comments what these functions do - my first thought for Update was that it sets a value if the value already was cached, and does nothing otherwise. Not sure if Map is that much better, but hopefully the signature helps here to clear that up a but

ajscholl (changes requested):

ajscholl (changes requested):

ajscholl (approved):

PR #1170, opened 2024-10-29
tracing, http: use the trace ID from the load balancer if available and forward traces via HTTP if configured

ajscholl:


pkg/http/client.go
```diff
@@ -207,6 +211,10 @@ type DialerSettings struct {
 	FallbackDelay time.Duration `cfg:"fallback_delay" default:"0s"`
 }
 
+type TracingSettings struct {
+	ForwardTraceId bool `cfg:"forward_trace_id" default:"false"`
```
j4k4, line 215: Do we wanna have `true` as default here?\
For me this looks more like a opt-out than a opt-in.

j4k4 (approved):

PR #1172, opened 2024-11-08
feat: added multi metric writer support

applike-ss:


pkg/httpserver/middleware_metric.go
```diff
@@ -10,9 +10,11 @@ import (
 )
 
 const (
-	MetricHttpRequestCount        = "HttpRequestCount"
-	MetricHttpRequestResponseTime = "HttpRequestResponseTime"
-	MetricHttpStatus              = "HttpStatus"
+	detailed                       = "Detailed"
+	MetricHttpRequestCount         = "HttpRequestCount"
+	MetricHttpRequestCountDetailed = "HttpRequestCountDetailed"
```
j4k4, line 15: Should that be `HttpRequestCountPerRoute`?

pkg/metric/writer_cw.go
```diff
@@ -31,101 +30,59 @@ const (
...
 	clock       clock.Clock
 	client      gosoCloudwatch.Client
 	cwNamespace string
 }
 
-func NewCwWriter(ctx context.Context, config cfg.Config, logger log.Logger) (Writer, error) {
-	testClock := clock.NewRealClock()
-	settings := getMetricSettings(config)
-	cwNamespace := GetCloudWatchNamespace(config)
+func ProvideCloudwatchWriter(ctx context.Context, config cfg.Config, logger log.Logger) (Writer, error) {
```
j4k4, line 45: Usually, we have the 3 `Provide*`, `New* ` and `New*WithInterfaces`?\
Do we want to split it here as well again?

pkg/metric/writer_multi.go
```diff
@@ -0,0 +1,36 @@
+package metric
+
+type multiWriter struct {
+	writers []Writer
+}
+
+func NewMultiWriter(writers []Writer) Writer {
```
j4k4, line 7: Should this be named then `NewMultiWriterWithInterfaces`?

pkg/metric/writer_factory.go
```diff
@@ -9,20 +9,34 @@ import (
...
 
-func ProvideMetricWriterByType(ctx context.Context, config cfg.Config, logger log.Logger, typ string) (Writer, error) {
-	switch typ {
-	case WriterTypeCw:
-		return NewCwWriter(ctx, config, logger)
-	case WriterTypeES:
-		return NewEsWriter(config, logger)
-	case WriterTypeProm:
-		return NewPromWriter(ctx, config, logger)
+func ProvideMetricWriter(ctx context.Context, config cfg.Config, logger log.Logger, types []string) (Writer, error) {
```
j4k4, line 17: this is not a real `Provide*` in our usual sense, isn't it?\
This is more like the normal `NewMultiWriter`?

pkg/httpserver/middleware_metric.go
```diff
@@ -10,9 +10,11 @@ import (
 )
 
 const (
-	MetricHttpRequestCount        = "HttpRequestCount"
-	MetricHttpRequestResponseTime = "HttpRequestResponseTime"
-	MetricHttpStatus              = "HttpStatus"
+	detailed                       = "Detailed"
```
j4k4, line 13: Do we need to change this to `PerRoute` as well?

applike-ss: yap, adjusted that.

j4k4 (approved):

applike-ss:
Release draft:

<h1>Kernel changes</h1>
Module factories can now return a nil value, to add a module to the current stage e.g. when they are disabled via configuration.

<h1>Metric changes</h1>
For properly supporting prometheus, several adjustments were made that also change the behavior of the other metric writers.

This was required, as prometheus does not allow the same metric name with different amount of labels.

Changes to metrics:

**Kinsumer**
The `ShardTaskRatio` metric without dimensions was renamed to `ShardTaskRatioMax`. The `ShardTaskRatio` metric with the `StreamName` dimenions was not touched.

The metrics that contained the dimenions `StreamName` or `StreamName` and `ShardId` were split. The latter have `PerShard` appended to their metric name.
This results in the following changes for metrics containing both dimenions:
| Old name | New name |
| --- | --- |
| SleepDuration | SleepDurationPerShard |
| FailedRecords | FailedRecordsPerShard |
| MillisecondsBehind | MillisecondsBehindPerShard |
| ProcessDuration | ProcessDurationPerShard |
| ReadCount | ReadCountPerShard |
| ReadRecords | ReadRecordsPerShard |
| WaitDuration | WaitDurationPerShard |

**Http server**
The metrics containing the dimentions `Method`, `Path` and `ServerName` to now have a `PerRoute` suffix. The metrics containing only the `ServerName` dimensions were not touched.

This results in the following changes for metrics containing the three dimenions:
| Old name | New name |
| --- | --- |
| HttpRequestCount | HttpRequestCountPerRoute |
| HttpRequestResponseTime | HttpRequestResponseTimePerRoute |
| HttpStatus2XX | HttpStatus2XXPerRoute |
| HttpStatus3XX | HttpStatus3XXPerRoute |
| HttpStatus4XX | HttpStatus4XXPerRoute |
| HttpStatus5XX | HttpStatus5XXPerRoute |

A log message writing after the channel was closed was limited to writing it only once to not spam logs.

The metric writing for prometheus was rewritten, as it wasn't working as expected. This also has the side effect that the metrics become avaiable after the metric daemon writer persist interval (`metric.interval` config key).
The prometheus metric server now only acts as a server for serving the metrics and does not have any metric storage logic anymore.

PR #1177, opened 2024-11-18
stream: delay context cancels for retry messages to ensure we can write them to the retry queue

ajscholl:
If we get a context cancel while processing a message from something like kinesis, writing to the retry queue would also happen on a canceled context, failing immediately and loosing our message. We need some small delay for the cancel to prevent this.

### DIFF
--- a/pkg/stream/consumer_base.go
+++ b/pkg/stream/consumer_base.go
@@ -12,6 +12,7 @@ import (
 	"github.com/justtrackio/gosoline/pkg/cfg"
 	"github.com/justtrackio/gosoline/pkg/clock"
 	"github.com/justtrackio/gosoline/pkg/coffin"
+	"github.com/justtrackio/gosoline/pkg/exec"
 	"github.com/justtrackio/gosoline/pkg/funk"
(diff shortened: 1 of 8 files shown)


pkg/stream/consumer_base.go
```diff
@@ -58,14 +59,15 @@ type baseConsumer struct {
...
-	retryHandler RetryHandler
+	clock          clock.Clock
+	uuidGen        uuid.Uuid
+	logger         log.Logger
+	metricWriter   metric.Writer
+	tracer         tracing.Tracer
+	encoder        MessageEncoder
+	retryInput     Input
+	retryHandler   RetryHandler
+	retryGraceTime time.Duration
```
j4k4, line 70: Is there a reason which I don't see why we have this `retryGraceTime` as a dedicated field?\
don't we have this one on the consumer settings on line 79 already?\

ajscholl: Good catch. I introduced a new field because I first took it from the kernel directly, but now it is in the settings I already have. Removed it again and used it directly from the settings.

j4k4 (approved):

PR #1178, opened 2024-11-19
httpserver: don't cancel requests for write operations externally;

selm0:
Currently, there is a risk for write operations in the httpserver to fail, as we don't have any transactions. To mitigate this, we should make the context used for the write operation independent of the request context.
Contains a minor adjustment for mysql logging, where everything is an error - but the root cause may be a context cancellation.


pkg/httpserver/crud/errors.go
```diff
@@ -1,5 +1,75 @@
...
+	"github.com/justtrackio/gosoline/pkg/log"
+	"github.com/justtrackio/gosoline/pkg/validation"
+)
 
 var ErrModelNotChanged = fmt.Errorf("nothing has changed on model")
+
+// handleErrorOnWrite handles errors for read operations.
+// Covers many default errors and responses like
+//   - context.Canceled, context.DeadlineExceed -> HTTP 299
+//   - validation.Error -> HTTP 400
```
ajscholl, line 21: 404 is missing in this list

pkg/httpserver/crud/errors.go
```diff
@@ -1,5 +1,75 @@
...
+}
+
+// handleErrorOnWrite handles errors for write operations.
+// Covers many default errors and responses like
+//   - context.Canceled, context.DeadlineExceed -> HTTP 500
+//   - db_repo.RecordNotFoundError -> HTTP 404
+//   - ErrModelNotChanged -> HTTP 304
+//   - db.IsDuplicateEntryError -> HTTP 409
+//   - validation.Error -> HTTP 400
+func handleErrorOnWrite(logger log.Logger, err error) (*httpserver.Response, error) {
```
ajscholl, line 46: Does it make sense to implement these methods on the base handler and then pass a context instead of a logger? Well, one would need to handle the cases where fields get set on the logger...

pkg/httpserver/crud/create.go
```diff
@@ -2,25 +2,28 @@ package crud
...
-	logger      log.Logger
-	transformer CreateHandler
+	logger       log.Logger
+	transformer  CreateHandler
+	releaseDelay time.Duration
 }
 
-func NewCreateHandler(logger log.Logger, transformer CreateHandler) gin.HandlerFunc {
+func NewCreateHandler(config cfg.Config, logger log.Logger, transformer CreateHandler) gin.HandlerFunc {
+	releaseDelay := config.GetDuration("crud.release_delay", 10*time.Minute)
```
ajscholl, line 21: Please create a `settings.go` where this is a field on a struct, allowing us to
- have a single place for the default
- have a place to document what this actually does
- you could even implement it such that there is a delay per CRUD method with some general default if you wanted to (but maybe that is too much?)

pkg/httpserver/crud/create_test.go
```diff
@@ -0,0 +1,77 @@
...
+		Name: mdl.Box("foobar"),
+	}
+
+	config := configMocks.NewConfig(t)
+	config.EXPECT().GetDuration("crud.release_delay", 10*time.Minute).Return(10 * time.Minute)
+
+	logger := logMocks.NewLoggerMock(logMocks.WithMockAll, logMocks.WithTestingT(t))
+	transformer := NewTransformer()
+
+	transformer.Repo.On("Create", mock.AnythingOfType("*exec.stoppableContext"), model).Run(func(args mock.Arguments) {
```
ajscholl, line 30: When moving / modifying tests, you could think about replacing `.On("Foo")` with `.EXPECT().Foo()`

pkg/httpserver/crud/errors.go
```diff
@@ -1,5 +1,81 @@
...
+	"github.com/justtrackio/gosoline/pkg/validation"
+)
+
+const HttpStatusClientClosedRequest = 499
 
 var ErrModelNotChanged = fmt.Errorf("nothing has changed on model")
+
+// handleErrorOnWrite handles errors for read operations.
+// Covers many default errors and responses like
+//   - context.Canceled, context.DeadlineExceed -> HTTP 299
```
ajscholl, line 23:
```suggestion
//   - context.Canceled, context.DeadlineExceed -> HTTP 499
```

pkg/httpserver/crud/errors.go
```diff
@@ -1,5 +1,81 @@
...
+//   - context.Canceled, context.DeadlineExceed -> HTTP 500
+//   - db_repo.RecordNotFoundError | db_repo.NoQueryResultsError -> HTTP 404
+//   - ErrModelNotChanged -> HTTP 304
+//   - db.IsDuplicateEntryError -> HTTP 409
+//   - validation.Error -> HTTP 400
+func handleErrorOnWrite(ctx context.Context, logger log.Logger, err error) (*httpserver.Response, error) {
+	logger = logger.WithContext(ctx)
+
+	if exec.IsRequestCanceled(err) {
+		logger.Error("failed to update model(s): %s", err.Error())
```
ajscholl, line 54: Use `%w` for errors when logging as error (we can't do that for Info or Warn):

```suggestion
		logger.Error("failed to update model(s): %w", err)
```

pkg/httpserver/crud/errors.go
```diff
@@ -1,5 +1,81 @@
...
+
+const HttpStatusClientClosedRequest = 499
 
 var ErrModelNotChanged = fmt.Errorf("nothing has changed on model")
+
+// handleErrorOnWrite handles errors for read operations.
+// Covers many default errors and responses like
+//   - context.Canceled, context.DeadlineExceed -> HTTP 299
+//   - db_repo.RecordNotFoundError | db_repo.NoQueryResultsError -> HTTP 404
+//   - validation.Error -> HTTP 400
```
ajscholl, line 25: I don't see the code for this anywhere

pkg/httpserver/crud/create.go
selm0: Added settings and documentation

PR #1181, opened 2024-12-02
test/suite: accept maps of tests and run them as sub-tests

ajscholl:
If you have complex logic generating test cases, it can be helpful to be able to generate a map of test cases to run. Normally, you would then use `t.Run(name, func() {...})` to execute each test case, but this isn't an option when using the test suite to run http or stream test cases. Therefore, we now allow you to return a `map[string]T` where `T` is either a `ToHttpserverTestCaseList` interface or directly a `[]*HttpserverTestCase` slice.

Additionally, the `HttpserverTestCaseListProvider` type was added to allow you to return a function creating the test cases. The function runs after fixtures (and the app itself) have been loaded compared to the test case generator (which runs before the app has been loaded).

---

This expands the test case map logic from the http server test cases to the stream test cases. `ToStreamTestCase` and `StreamTestCaseProvider` have been added with similar roles to their http server counterparts.

---

This changes the way we check method signatures of test methods slightly. For one, we are now checking the complete signature    (previously, we wouldn't check for example the receiver argument).    On the other hand, we now check that at least one matcher actually    matches a method starting with "Test" and raise an error otherwise.    This ensures that a test is not silently skipped because there is a    small error in the method signature.
    
Therefore, this is technically a breaking change.

### DIFF
--- a/pkg/test/suite/testcase_httpserver_extended.go
+++ b/pkg/test/suite/testcase_httpserver_extended.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"path"
 	"reflect"
+	"strings"
 	"testing"
 
 	"github.com/go-resty/resty/v2"
@@ -16,11 +17,10 @@ import (
 )
 
 func init() {
-	testCaseDefinitions["httpserverExtended"] = testCaseDefinition{
-		matcher: isTestCaseHttpserverExtended,
-		builder: buildTestCaseHttpserverExtended,
-	}
+	registerTestCaseDefinition("httpserverExtended", isTestCaseHttpserverExtended, buildTestCaseHttpserverExtended)
 }
+
+const expectedTestCaseHttpserverExtendedSignature = "func (s TestingSuite) TestFunc() T"
 
 // A HttpserverTestCase should be the return value from your test functions. The following signatures are supported:
 //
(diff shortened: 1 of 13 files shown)


pkg/test/suite/testcase_httpserver_extended.go
```diff
@@ -120,98 +157,234 @@ func (c HttpserverTestCase) request(client *resty.Client) (*resty.Response, erro
...
+		runner, err := runHttpServerExtendedTests(suite, func() []*HttpserverTestCase {
+			if testCasesProvider == nil {
+				return nil
+			}
 
-				remainingRedirects--
+			return testCasesProvider.ToTestCaseList()
+		})
+		if err != nil {
+			return nil, err
```
bt-justtrack, line 272: nit: wrapping

pkg/test/suite/testcase_httpserver_extended.go
```diff
@@ -20,6 +23,16 @@ func init() {
 	}
 }
```
bt-justtrack, line 25: A comment explaining roughly how the parts in here are sticked together would be nice, as it is not the most self explanatory.

bt-justtrack (approved):

PR #1182, opened 2024-12-06
currency: add locking to currency updater

### LINKED ISSUE #1169: Add locking to the currency module
When an application is exhibiting traffic spikes and thus a substantial amount of instances get scheduled, the starting currency modules can run into the following situation: 
- The currecy last update timestamp is expired and thus a starting task will start updating the currency table
- This does not happen instantly, as such a substantial amount of tasks will try to update the table
- This can exceed what the table backing the currency data is provisioned for / exceed aws account limits
- This leads to the currency modules not getting healthy in time, leading to the kernel exiting
- That causes new tasks to spawn, leading to a spiraling effect

As such we want to introduce a lock on the currency data leading to only one module running the update.

erdac:
closes #1169

### DIFF
--- a/pkg/currency/module.go
+++ b/pkg/currency/module.go
@@ -6,9 +6,17 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/hashicorp/go-multierror"
 	"github.com/justtrackio/gosoline/pkg/cfg"
+	"github.com/justtrackio/gosoline/pkg/conc"
+	"github.com/justtrackio/gosoline/pkg/conc/ddb"
 	"github.com/justtrackio/gosoline/pkg/kernel"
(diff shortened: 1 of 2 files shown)


pkg/currency/module.go
```diff
@@ -29,10 +38,22 @@ func NewCurrencyModule() kernel.ModuleFactory {
 			return nil, fmt.Errorf("can not create updater: %w", err)
 		}
 
+		appId := cfg.GetAppIdFromConfig(config)
+
+		lockProvider, err := ddb.NewDdbLockProvider(ctx, config, logger, conc.DistributedLockSettings{
```
beeemT, line 43: Imo we should document that this module needs a ddb table now

beeemT (approved):

bt-justtrack (approved):

PR #1183, opened 2024-12-11
db: close the sql.Rows object once we are done with it

ajscholl:
We sometimes have code which would iterate an sql.Rows object, returning an error if we failed to read the next row. As long as we don't get an error, this works, as calling `rows.Next()` will eventually close the rows object once the last row was read. However, in case of an error we have to make sure to explicitly call `rows.Close()`, otherwise we leak connections for some drivers.

### DIFF
--- a/examples/fixtures/sqlx/main.go
+++ b/examples/fixtures/sqlx/main.go
@@ -6,6 +6,7 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/hashicorp/go-multierror"
 	"github.com/jmoiron/sqlx"
 	"github.com/justtrackio/gosoline/pkg/application"
 	"github.com/justtrackio/gosoline/pkg/cfg"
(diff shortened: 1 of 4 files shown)


examples/fixtures/sqlx/main.go
```diff
@@ -61,14 +62,20 @@ func main() {
...
 					return err
 				}
 
+				defer func() {
+					if closeErr := rows.Close(); closeErr != nil {
+						err = multierror.Append(err, fmt.Errorf("closing rows failed: %w", closeErr))
+					}
+				}()
+
 				for rows.Next() {
```
bt-justtrack, line 79: Would be cool if we could simultaneously tackle checking rows.Err() after for loops iterating rows.Next()

bt-justtrack (changes requested):

bt-justtrack (approved):

PR #1188, opened 2024-12-19
httpserver: don't treat request cancels as errors;

selm0:
**HTTP Server**
The standard http server treats external context cancellations and disconnects as errors; even though the client may have closed or canceled the request on purpose.
As a follow up to https://github.com/justtrackio/gosoline/pull/1178 (targeting the crud handlers), this PR is adjusting the standard behaviour of any other handler similarly.
The most common use cases:
 - access forbidden transformed into HTTP 403
 - context cancellation / disconnects are transformed into HTTP 499
 - any other error being transformed into a HTTP 500

The logger middleware is also adjusted appropriately, logging connection and request cancelations as level info. Bind errors and render errors are continously logged as warnings; by default any other error is logged as an error.

**Exec*
- fixing bugs which causes the code to panic on nil errors

**Dispatcher**
The dispatcher is hiding some errors which may happen during execution of the callbacks. They're now propagated back to the callee, making them obvious and necessary to deal with.
In some cases, you may have to wrap adding callbacks into the appctx to prevent them from being added multiple times.

_This is a breaking change, as the dispatcher will now stop execution if any error is encountered_

### DIFF
--- a/pkg/httpserver/middleware_logger_test.go
+++ b/pkg/httpserver/middleware_logger_test.go
@@ -1,0 +1,212 @@
+package httpserver_test
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"net/http"
+	"net/http/httptest"
+	"net/url"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/gin-gonic/gin"
+	"github.com/justtrackio/gosoline/pkg/clock"
+	clockMocks "github.com/justtrackio/gosoline/pkg/clock/mocks"
+	"github.com/justtrackio/gosoline/pkg/httpserver"
+	"github.com/justtrackio/gosoline/pkg/log"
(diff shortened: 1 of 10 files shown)


pkg/httpserver/middleware_logger_test.go
```diff
@@ -0,0 +1,202 @@
...
+
+	s.logger.EXPECT().Info("%s %s %s", "GET", "path", "HTTP/1.1")
+
+	s.handler(ginCtx)
+}
+
+func (s *loggingMiddlewareTestSuite) TestRequestCanceledError() {
+	ginCtx := buildRequest()
+
+	_ = ginCtx.Error(context.Canceled) //nolint:errcheck // just returns the error passed as an argument
```
ajscholl, line 56: or you could just assert this actually happened and don't need nolints here

ajscholl (approved):

PR #1189, opened 2025-01-02
db-repo: use db client on creating orm instances

j4k4:
This release enables the use of different db client names on creating orm instances.

### DIFF
--- a/pkg/db-repo/orm.go
+++ b/pkg/db-repo/orm.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/iancoleman/strcase"
 	"github.com/jinzhu/gorm"
+	"github.com/jmoiron/sqlx"
 	"github.com/justtrackio/gosoline/pkg/cfg"
 	"github.com/justtrackio/gosoline/pkg/clock"
 	"github.com/justtrackio/gosoline/pkg/db"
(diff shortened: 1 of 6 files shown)


pkg/db-repo/orm.go
```diff
@@ -23,14 +24,21 @@ type OrmSettings struct {
...
 }
 
-func NewOrm(ctx context.Context, config cfg.Config, logger log.Logger) (*gorm.DB, error) {
-	connection, err := db.ProvideConnection(ctx, config, logger, "default")
-	if err != nil {
+func NewOrm(ctx context.Context, config cfg.Config, logger log.Logger, dbClientName string) (*gorm.DB, error) {
+	var err error
+	var connection *sqlx.DB
+
+	if dbClientName == "" {
```
selm0, line 31: I believe, for other entities we don't have this fallback. As it's more explicit, I suggest to pass in "default" at the callee site instead of an empty string.

pkg/db-repo/repository.go
```diff
@@ -37,7 +37,8 @@ var (
 
 type Settings struct {
 	cfg.AppId
-	Metadata Metadata
+	Metadata   Metadata
+	ClientName string
```
selm0, line 41: default value of "default"?

selm0 (changes requested):

selm0 (approved):

PR #1192, opened 2025-01-08
add kinsumer autoscale module

### LINKED ISSUE #1190: Autoscale kinsumer

erdac:
closes #1190


pkg/stream/kinsumer_autoscale_module.go
```diff
@@ -0,0 +1,191 @@
...
+func KinsumerAutoscaleModuleFactory(kinsumerInputName string) func(ctx context.Context, config cfg.Config, logger log.Logger) (map[string]kernel.ModuleFactory, error) {
+	return func(ctx context.Context, config cfg.Config, logger log.Logger) (map[string]kernel.ModuleFactory, error) {
+		modules := map[string]kernel.ModuleFactory{}
+		settings := readKinsumerAutoscaleSettings(config)
+
+		if !settings.Enabled {
+			return modules, nil
+		}
+
+		modules[kinsumerAutoscaleModuleName] = newKinsumerAutoscaleModule(kinsumerInputName, settings)
```
bt-justtrack, line 45: for what is the kinsumerAutoscaleModuleName used ? if we have more than one kinsumer in an app they would collide right ? For that we would need to make sure that this does not happen.

pkg/stream/kinsumer_autoscale_orchestrator_ecs.go
```diff
@@ -0,0 +1,62 @@
...
+		return 0, fmt.Errorf("failed to describe ecs service: %w", err)
+	}
+
+	currentDesiredCount := servicesOutput.Services[0].DesiredCount
+
+	return currentDesiredCount, nil
+}
+
+func (k kinsumerAutoscaleOrchestratorECS) UpdateTaskCount(ctx context.Context, taskCount int32) error {
+	updateServiceInput := &ecs.UpdateServiceInput{
```
bt-justtrack, line 50: do kinsumers now need any additional permissions for doing the describe and update of itself in ecs ?

erdac: yes there will be a change for the ecs module to give them these permissions

pkg/stream/kinsumer_autoscale_module.go
erdac: Adding the module twice for each kinsumer input doesn't really make sense.
If there are two different streams you would have to decide, which one to scale the app on by providing the right kinsumer input name in the `WithKinsumerAutoscaleModule` option.

bt-justtrack (approved):

bt-justtrack (approved):

PR #1196, opened 2025-02-02
resource lifecycles

j4k4:
This PR works towards unifying creation and purging of external resources.
Currently, all resources like SQS queues or DynamoDb tables are created "in place" during instantiation of their corresponding services.
This makes it harder to execute their life cycle actions in a controlled manner.
With this PR all life cycle actions will get executed in the same centralized place.


test/db-repo/change_history/entity/change_history_test.go
```diff
@@ -299,5 +299,6 @@ func (s *ChangeHistoryTestSuite) assertHistory(entityId uint, expected []definit
 		return *h
 	})
 
+	s.Equal(len(expected), len(items))
```
ajscholl, line 302:
```suggestion
	s.Len(items, len(expected))
```

pkg/reslife/container.go
```diff
@@ -0,0 +1,31 @@
...
+type (
+	containerCtxKey struct{}
+)
+
+func AddLifeCycleer(ctx context.Context, factory LifeCycleerFactory) error {
+	var err error
+	var container *[]LifeCycleerFactory
+
+	if container, err = provideContainer(ctx); err != nil {
+		return fmt.Errorf("could not add life cycle purger: %w", err)
```
ajscholl, line 19: This error looks wrong

pkg/reslife/container.go
```diff
@@ -0,0 +1,31 @@
...
+	}
+
+	*container = append(*container, factory)
```
ajscholl, line 22: Is it okay that this is not thread safe?

pkg/reslife/lifecycle.go
```diff
@@ -0,0 +1,256 @@
...
+	})
+}
+
+func (m *LifeCycleManager) refreshResources(ctx context.Context) error {
+	var err error
+	var container *[]LifeCycleerFactory
+	var resource LifeCycleer
+
+	if container, err = provideContainer(ctx); err != nil {
+		return fmt.Errorf("could not get lifecyleers: %w", err)
```
ajscholl, line 88: is this error message correct? we are getting factories, not lifecyclers directly

pkg/reslife/lifecycle.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+
+	for _, factory := range *container {
+		if resource, err = m.refresher(factory); err != nil {
+			return fmt.Errorf("could not build lifecycleer: %w", err)
+		}
+
+		m.resources = append(m.resources, resource)
```
ajscholl, line 96: So refresh extends our resources on every iteration? That feels... wrong. But not sure if it is actually wrong

pkg/reslife/lifecycle.go
```diff
@@ -0,0 +1,256 @@
...
+
+		return &LifeCycleManager{
+			logger:   logger,
+			clock:    clock.Provider,
+			settings: settings,
+			refresher: func(fac LifeCycleerFactory) (LifeCycleer, error) {
+				return fac(ctx, config, logger)
+			},
+			resources:  nil,
+			created:    []string{},
```
ajscholl, line 75: These three look like `funk.Set[string]` to me

pkg/reslife/lifecycle.go
```diff
@@ -0,0 +1,256 @@
...
+		took := m.clock.Since(now)
+		m.logger.Info("created resource %s in %s", res.GetId(), took)
+	}
+
+	m.logger.Info("executed the create lifecycle")
+
+	return nil
+}
+
+func (m *LifeCycleManager) Init(ctx context.Context) error {
```
ajscholl, line 142: Why is init special?
- it doesn't check if something was already initialized
- there is not time measured
- there is no log that this happened

pkg/reslife/lifecycle.go
```diff
@@ -0,0 +1,256 @@
...
+	for _, res := range m.resources {
+		if registerer, ok = res.(Registerer); !ok {
+			continue
+		}
+
+		if funk.Contains(m.registered, res.GetId()) {
+			continue
+		}
+
+		if key, data, err = registerer.Register(ctx); err != nil {
```
ajscholl, line 198: Do we need to measure the time here again?

pkg/cloud/aws/kinesis/kinsumer.go
```diff
@@ -150,33 +132,35 @@ func NewKinsumer(ctx context.Context, config cfg.Config, logger log.Logger, sett
...
 	}
 
 	return NewKinsumerWithInterfaces(logger, *settings, fullStreamName, kinesisClient, metadataRepository, metricWriter, clock.Provider, shardReaderFactory), nil
 }
 
-func NewKinsumerWithInterfaces(logger log.Logger, settings Settings, stream Stream, kinesisClient Client, metadataRepository MetadataRepository, metricWriter metric.Writer, clock clock.Clock, shardReaderFactory func(logger log.Logger, shardId ShardId) ShardReader) Kinsumer {
+func NewKinsumerWithInterfaces(
+	logger log.Logger,
+	settings Settings,
+	stream Stream,
```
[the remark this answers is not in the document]
j4k4: :+1:

pkg/cloud/aws/sns/topic.go
```diff
@@ -44,36 +41,26 @@ type TopicSettings struct {
...
-	}
-
-	if err = appctx.MetadataAppend(ctx, MetadataKeyTopics, metadata); err != nil {
-		return nil, fmt.Errorf("can not access the appctx metadata: %w", err)
+	topicArn := aws.String("")
+	if err = reslife.AddLifeCycleer(ctx, NewLifecycleManager(settings, topicArn)); err != nil {
+		return nil, fmt.Errorf("can not add lifecycle manager: %w", err)
 	}
 
 	return NewTopicWithInterfaces(logger, client, topicArn), nil
```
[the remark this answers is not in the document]
j4k4: :+1:

pkg/reslife/lifecycle.go
j4k4: It's not beautiful but as we don't know which of those life cycles get called first, this is the safe way to do it.. maybe we can make it nicer in the next iteration..

pkg/reslife/container.go
j4k4: :+1:

pkg/reslife/container.go
j4k4: locked

test/db-repo/change_history/entity/change_history_test.go
j4k4: :+1:

pkg/db/lifecycle.go
```diff
@@ -61,19 +62,24 @@ func (m *lifecycleManager) GetId() string {
...
+
 	rows, err := m.db.QueryContext(ctx, "SELECT TABLE_NAME FROM information_schema.TABLES WHERE TABLE_SCHEMA = ?;", m.settings.Uri.Database)
 	if err != nil {
 		return fmt.Errorf("failed to check tables of database: %w", err)
 	}
 
 	for rows.Next() {
 		if err = rows.Scan(&table); err != nil {
-			return fmt.Errorf("could not scan row: %w", err)
+			break
```
ajscholl, line 82: It is a bit counter-intuitive that we retrieve the error later again by calling rows.Err()... do we need a comment for this?

j4k4: can add a comment..

ajscholl (approved):

pkg/reslife/lifecycle.go
```diff
@@ -0,0 +1,257 @@
...
+	var resource LifeCycleer
+
+	if cont, err = provideContainer(ctx); err != nil {
+		err = fmt.Errorf("could not get lifecyleer factories: %w", err)
+
+		return
+	}
+
+	i := len(m.resources)
+	for ; i < len(cont.factories); i++ {
```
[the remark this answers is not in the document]
j4k4: Would be safer, I guess.. :+1:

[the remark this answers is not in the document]
bt-justtrack: Actually my thought was to lock the lock of the container and not the function, so we dont have any changes to the containers factories while we iterate over them

bt-justtrack (approved):

PR #1200, opened 2025-02-14
metric: refactoring prometheus metrics and metric daemon

applike-ss:


applike-ss:
Release Note draft:

In this release there is several metric related changes and other smaller fixes.

# Metrics 
## The metric channel
The metric writer channel is now private, as it is only supposed to be used with the metric writer and the metric daemon.

## Metric settings
The metric writers have moved their config key from `metric.writer` (singular) to `metric.writers` (plural) to properly show that you can actually configure multiple writers.

## Metric writer renames (**BREAKING**)
The metric writer were renamed for better readability:
| old writer name | new writer name |
| -- | -- |
| `cw` | `cloudwatch` |
| `es` | `elasticsearch` |
| `prom` | `prometheus` |

## Ability to add new writers
You can now add your own metric writer implementation by providing a factory func to `metric.RegisterWriterFactory` and configuring its usage in the `metric.writers` config key.
If you intent to create your own metric writer, please make it read the configuration unter `metric.writer_settings.<writer-name>`, e.g. `metric.writer_settings.my_metric_writer`

## Metric daemon behavior changes
The metric daemon now supports both (the original behavior) to aggregate the metrics for the writer implementation (ticking as per the configured interval in `metric.interval`, defaulting to 60s) and to send them raw (instant).
If you add your own implementation, then make sure to set `metric.writer_settings.<writer-name>.aggregate` to true if you want your metrics pre-aggregated.
The default is not to pre-aggregate metrics, unless the cloudwatch implementation is used, as you are paying for the api calls here and you could run into rate limits as well easily if the metrics were not aggregated.
The daemon writer module is now an essential background module making the application stop should it ever fail.

## CloudWatch changes
The Cloudwatch naming config was moved from `metric.cloudwatch.naming.pattern` to `metric.writer_settings.cloudwatch.naming.pattern`

## Prometheus changes
The prometheus metrics server is now an essential background module, so applications that stop on their own (without an explicit termination signal being passed to them) are now exiting as expected and don't stay running until they get terminated or killed.
The config key for prometheus metrics server is now `metric.writer_settings.prometheus` instead of `prometheus`.
This is to unify the configuration of metric writers.
The prometheus metrics server is now only enabled when the writer (`prometheus`) also happens to exist in the `metric.writers` config value AND the api is enabled via the settings (default=true) as well.
Fixed an issue where the summaries wrote wrong data due to pre-aggregated metrics logic being used. This could result in the app trying to write the same metric multiple times and the values being summed up, while the count was only incremented once.

## Removed writers
The multi metric writer (intermediate non-replacable default) and noop writer were removed and logic was moved into the metric daemon writer.

# Other changes

* Some github workflow related actions were updated.
* The golangci-lint config was fixed, so that the most revent github action wouldn't exit with an error.
* Dockertest dependency was updated.

pkg/metric/writer_prometheus.go
```diff
@@ -143,133 +146,155 @@ func (w *prometheusWriter) buildHelp(data *Datum) string {
...
+
+			return
 		}
 
-		metricI = metric
+		metric = metricR
+	} else {
+		err = w.addMetric()
+		if err != nil {
+			w.logger.Error(errFailedToAddMetricToPrometheusCounter, err)
```
j4k4, line 292: do we have to adjust the error message as this is not a counter?

pkg/metric/writer_prometheus.go
```diff
@@ -143,133 +146,155 @@ func (w *prometheusWriter) buildHelp(data *Datum) string {
...
```
j4k4, line 266: do we have to adjust the error message as this is not a counter?

pkg/metric/writer_prometheus.go
```diff
@@ -143,133 +146,155 @@ func (w *prometheusWriter) buildHelp(data *Datum) string {
...
```
j4k4, line 240: do we have to adjust the error message as this is not a counter?

pkg/metric/writer_prometheus.go
```diff
@@ -143,133 +146,155 @@ func (w *prometheusWriter) buildHelp(data *Datum) string {
...
-	id := w.DatumId(datum)
-
-	metricI, ok := w.promMetrics.Load(id)
-	if !ok {
-		var err error
-		metric := w.createHistogram(datum)
+	metric := w.createHistogram(datum)
 
-		err = w.addMetric(id, metric)
+	if err := w.registry.Register(metric); err != nil {
```
j4k4, line 280: this code block looks like copy and paste.. would it be possible to abstract it in a better way?

j4k4 (changes requested):

j4k4 (approved):

PR #1202, opened 2025-02-18
metric: limit warning logs and change metric queue to unbounded queue

### LINKED ISSUE #1198: Handle full metric channel
The metric daemon can write metrics if invalid metrics are published. This can cause the metric channel to get full, resulting in the metric daemon to deadlock.

- [x] Writing to the metric channel must be able to fail or at least not block forever. If the channel is full, we have to log a warning or error (hmm, this writes a metric again... so we need some way to prevent an infinite loop here) and drop the metric, or use an unbounded channel (not an actual go thing)
- [x] The metric daemon shouldn't be able to deadlock itself
- [x] If metrics are invalid, maybe we have to disable these logs for some time (like a minute?) to reduce the number of logs produced by this

ajscholl:
The metric daemon could run into two different deadlocks:
- Writing a metric to a full queue while the daemon was shutting down: In this case, the daemon was waiting for the lock on the metric channel struct, but the metric channel struct was busy writing to the channel and waiting for the daemon to read something from it.
- Writing a metric after an invalid metric: With proper timing, one could fill up the metric channel after the metric daemon took out an invalid message. Now the metric daemon would try to write a warning, which itself would write another metric. But the channel was already full and the metric daemon was now stuck waiting for itself reading a message from it.

This is now solved by changing the metric channel to only feature a single flag (are metrics available) and managing the batch of metrics that are available ourselves. That way, we can just ignore writes to the full flag channel and append the metrics to write anyway. When reading metrics, we can now take the whole batch instead of having to read single metrics from the channel and append them one by one.

Closes #1198

### DIFF
--- a/pkg/metric/daemon.go
+++ b/pkg/metric/daemon.go
@@ -39,7 +39,6 @@ type BatchedMetricDatum struct {
 
 type Daemon struct {
 	kernel.EssentialBackgroundModule
-	sync.Mutex
 	logger   log.Logger
 	settings *Settings
 
@@ -50,6 +49,9 @@ type Daemon struct {
 
 	batch          map[string]*BatchedMetricDatum
 	dataPointCount int
+
+	warningThrottlesLck sync.Mutex
+	warningThrottles    map[string]bool
 }
 
 func metricWriterAggrKey(typ string) string {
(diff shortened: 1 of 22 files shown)


pkg/metric/daemon.go
```diff
@@ -205,3 +208,27 @@ func (d *Daemon) buildMetricData() Data {
 
 	return data
 }
+
+// we don't want to log warnings every time they occur - it is enough to log them once they occur, at least for a minute
+func (d *Daemon) checkThrottleWarning(warning string) bool {
```
bt-justtrack, line 213: nit: Calling this shouldEmitThrottleWarning would make reading its usage even nicer

bt-justtrack (approved):

pkg/metric/daemon.go
ajscholl: True - changed it

PR #1205, opened 2025-02-21
Feat crud nested resources

bt-justtrack:


pkg/httpserver/crud/errors.go
```diff
@@ -7,7 +7,7 @@ import (
 	"net/http"
 
 	"github.com/justtrackio/gosoline/pkg/db"
-	"github.com/justtrackio/gosoline/pkg/db-repo"
+	db_repo "github.com/justtrackio/gosoline/pkg/db-repo"
```
selm0, line 10: Do we have a common rule here? I tend to remove these unnecessary aliases.

bt-justtrack: While technically unnecessary this one should be there imo. The import path might be db-repo but the package is named db_repo, thus the alias added by the formatter. It is changing it to what the package is actually called. Thus i would say it is not unnecessary but the root cause is that we named the import path for the packge with a dash instead of the underscore.

ajscholl: I normally import that as `dbRepo` as my IDE doesn't like names in snake case (and `db-repo` vs `db_repo` confuses some tools without an explicit name on the import)

pkg/httpserver/middleware_request_context.go
```diff
@@ -0,0 +1,23 @@
...
+import (
+	"context"
+
+	"github.com/gin-gonic/gin"
+)
+
+type ContextKey string
+
+const (
+	RequestContextKey ContextKey = "http-request"
```
ajscholl, line 12: Please make this a private const and a private type - there is no need to export this

pkg/httpserver/middleware_request_context.go
```diff
@@ -0,0 +1,23 @@
...
+)
+
+func ContextWithRequestMiddleware(c *gin.Context) {
+	ctx := context.WithValue(c.Request.Context(), RequestContextKey, c)
+	c.Request = c.Request.WithContext(ctx)
+	c.Next()
+}
+
+func GetRequestFromContext(ctx context.Context) *gin.Context {
```
ajscholl, line 21: I believe this should be fallible if you for some reason pass in a context without the request.

pkg/httpserver/middleware_request_context.go
```diff
@@ -0,0 +1,23 @@
...
+	"github.com/gin-gonic/gin"
+)
+
+type ContextKey string
+
+const (
+	RequestContextKey ContextKey = "http-request"
+)
+
+func ContextWithRequestMiddleware(c *gin.Context) {
```
ajscholl, line 15: Why is this public?

PR #1208, opened 2025-02-28
db: Allow passing options to the db.Client factories which replace either the executor entirely or provide additional error checkers

j4k4:


ajscholl (approved):

pkg/db/client.go
```diff
@@ -57,36 +57,49 @@ type (
...
+		clientOption    []func(*ClientSqlx)
+	}
 
-func ProvideClient(ctx context.Context, config cfg.Config, logger log.Logger, name string) (Client, error) {
+	ClientSqlxOption func(*clientSqlxOptions)
+
+	clientCtxKey string
+)
+
+// Provide client provides a client from context using the config for name.
```
ajscholl, line 98: I think you wanted to write at least

```suggestion
// ProvideClient provides a client from context using the config for name.
```

However, the part "using the config for name" sounds really strange

pkg/db/client.go
```diff
@@ -57,36 +57,49 @@ type (
...
+		WithTx(ctx context.Context, ops *sql.TxOptions, do func(ctx context.Context, tx *sqlx.Tx) error) error
+		Close() error
+	}
 
-type ClientSqlx struct {
-	logger   log.Logger
-	db       *sqlx.DB
-	executor exec.Executor
-}
+	ClientSqlx struct {
```
ajscholl, line 82: Why is this type public?

pkg/db/client.go
```diff
@@ -57,36 +57,49 @@ type (
...
+	}
 
-type clientCtxKey string
+	clientSqlxOptions struct {
+		executorOptions []exec.BackoffExecutorOption
+		clientOption    []func(*ClientSqlx)
+	}
 
-func ProvideClient(ctx context.Context, config cfg.Config, logger log.Logger, name string) (Client, error) {
+	ClientSqlxOption func(*clientSqlxOptions)
```
ajscholl, line 93: This is a client option, right? The sqlx part sounds like an implementation detail

pkg/db/client.go
```diff
@@ -95,36 +108,49 @@ func ProvideClient(ctx context.Context, config cfg.Config, logger log.Logger, na
...
 	}
 
 	if settings.Retry.Enabled {
 		path := fmt.Sprintf("db.%s.retry", name)
-		executor = NewExecutor(config, logger, name, path)
+		executor = NewExecutorWithOptions(config, logger, name, path, []exec.Notify{}, options.executorOptions...)
 	}
 
-	return NewClientWithInterfaces(logger, connection, executor), nil
+	client := NewClientWithInterfaces(logger, connection, executor).(*ClientSqlx)
```
ajscholl, line 147: That feels... ugly

pkg/db/client.go
```diff
@@ -465,3 +491,17 @@ func (c *ClientSqlx) WithTx(ctx context.Context, ops *sql.TxOptions, do func(ctx
 func (c *ClientSqlx) Close() error {
 	return c.db.Close()
 }
+
+func ClientSqlxWithExecutor(executor exec.Executor) ClientSqlxOption {
```
ajscholl, line 495: Why isn't this just called WithExecutor?

pkg/db/client.go
```diff
@@ -465,3 +491,17 @@ func (c *ClientSqlx) WithTx(ctx context.Context, ops *sql.TxOptions, do func(ctx
...
+	return func(c *clientSqlxOptions) {
+		c.clientOption = append(c.clientOption, func(c *ClientSqlx) {
+			c.executor = executor
+		})
+	}
+}
+
+func ClientSqlxWithExecutorErrorChecks(checks ...exec.ErrorChecker) ClientSqlxOption {
```
ajscholl, line 503: And this WithErrorCheck?

pkg/exec/executor_backoff.go
```diff
@@ -9,24 +9,40 @@ import (
...
+
+	BackoffExecutorOption func(*BackoffExecutor)
+)
 
 func NewBackoffExecutor(logger log.Logger, res *ExecutableResource, settings *BackoffSettings, checks []ErrorChecker, notifier ...Notify) *BackoffExecutor {
-	return &BackoffExecutor{
+	return NewBackoffExecutorWithOptions(logger, res, settings, checks, notifier)
+}
+
+func NewBackoffExecutorWithOptions(logger log.Logger, res *ExecutableResource, settings *BackoffSettings, checks []ErrorChecker, notifier []Notify, options ...BackoffExecutorOption) *BackoffExecutor {
```
ajscholl, line 31: Why is this passing checks and options? You can already set an option to add a check, so passing checks as a separate argument is redundant. Which makes the whole function redundant and I don't think we need any changes to this file

pkg/db/executor.go
```diff
@@ -12,40 +12,50 @@ import (
 )
 
 func NewExecutor(config cfg.Config, logger log.Logger, name string, backoffType string, notifier ...exec.Notify) exec.Executor {
+	return NewExecutorWithOptions(config, logger, name, backoffType, notifier)
+}
+
+func NewExecutorWithOptions(config cfg.Config, logger log.Logger, name string, backoffType string, notifier []exec.Notify, options ...exec.BackoffExecutorOption) exec.Executor {
```
ajscholl, line 18: Why is a function to create some kind of executor getting backoff executor options? it should just be executor options passed here

ajscholl (changes requested):

pkg/db/client.go
bt-justtrack: Don't know, it was already public and making it private now would potentially break smth

pkg/db/client.go
bt-justtrack: since we have multiple types in this package that could have options down the line, adding the Prefix for which type the Options are prevents collisions and helps with autocomplete. Since i am naming it ClientOption will change this to ClientWithExecutor

pkg/exec/executor_backoff.go
bt-justtrack: If we are not cleanly following the options pattern here and don't allow passing BackoffExecutorOptions you need to create special handling for adding custom checks everywhere where you are also passing default options. 

That would mean in db.NewExecutor you would need 
```go
	type clientSqlxOptions struct {
		additionalErrorCheckers []exec.ErrorChecker
		clientOption    []func(*ClientSqlx)
	}
```
and set these through an option in the db package. That achieves that specific use case, but necessitates the same implementation in every location where you want to attach custom errorcheckers to default ones. In addition once there are also more options eg you also want to apply them to the settings you need to again in every location support that dedicated logic. Following the option pattern cleanly means you dont need that for every check.
I've also slightly changed the way the db.ClientOption that passes the checks is implemented which makes this nicer imo.

pkg/db/client.go
ajscholl: If you leave the type public, we could just return a concrete type... or we could duplicate the contents of `NewClientWithInterfaces` here (not so nice) or change it so that it calls `newClientWithInterfaces` (which can return a concrete type just fine)

bt-justtrack: Changed it so it uses a concrete type

ajscholl (approved):

pkg/db/client.go
```diff
@@ -271,32 +270,19 @@ func (c *ClientSqlx) ExecMultiInTx(ctx context.Context, sqlers ...Sqler) (result
...
 		}
 
-		results = append(results, res)
-	}
-
-	if err = tx.Commit(); err != nil {
-		return nil, fmt.Errorf("can not commit transaction: %w", err)
+		return nil
+	})
+	if err != nil {
```
ajscholl, line 284: You could also just return the result of c.WithTx directly

pkg/db/client.go
```diff
@@ -445,36 +431,40 @@ func (c *ClientSqlx) BeginTx(ctx context.Context, ops *sql.TxOptions) (*sqlx.Tx,
...
 		if err != nil {
-			errRollback := tx.Rollback()
-			if errRollback != nil {
-				err = multierror.Append(err, fmt.Errorf("can not rollback tx: %w", errRollback))
+			return nil, err
+		}
 
-				return
+		defer func() {
+			if err != nil {
```
ajscholl, line 443: What happens if err is nil at this place because `do` did panic? I would expect the function to catch the panic and return an error for it, right?

pkg/db/client.go
bt-justtrack: That would mean we could return partial results which suggest there could be changed rows even though we rolled back the transaction.

PR #1229, opened 2025-03-28
log: introduce channel based log levels;

yicooo:
We identified a use case where we need finer control over log levels for specific channels to prevent excessive logging while maintaining observability. Previously, the logger configuration allowed defining a set of channels per handler but enforced a single log level for all of them.

This PR introduces per-channel log level configuration

Previously
```yaml
log:
  handlers:
    main:
      type: iowriter
      level: debug
      channels: [metrics]
      formatter: json
      timestamp_format: 2006-01-02T15:04:05Z07:00
      writer: stdout
``` 

With the change
```yaml
log:
  handlers:
    main:
      type: iowriter
      level: debug
      channels:
        - name: metrics
          level: error
      formatter: json
      timestamp_format: 2006-01-02T15:04:05Z07:00
      writer: stdout
```

With this change, the metrics channel can be configured to log only error level messages, even if the handler’s default level is debug. 

Important note: This is a breaking change, we could come up with a solution/dirty hack to make this a non-breaking change too if needed.

### DIFF
--- a/docs/docs/reference/src/log/main.go
+++ b/docs/docs/reference/src/log/main.go
@@ -15,8 +15,8 @@ func main() {
 	handler := log.NewHandlerIoWriter(
 		// the min log level to write (trace, debug, info, warn, error)
 		log.LevelDebug,
-		// a list of channels to filter for, if empty nothing is filtered. []string{"http"} would write logs from http channel only
-		[]string{},
+		// configuration of channels to filter for, if empty handler default will be used
+		log.Channels{},
 		// how to format the message. this will format in a console friendly way. log.FormatterJson would format log message as json
 		log.FormatterConsole,
 		// how to format the log time. uses the structure of the `time` package
--- a/pkg/log/handler_iowriter.go
+++ b/pkg/log/handler_iowriter.go
@@ -13,11 +13,15 @@ func init() {
 }
 
 type HandlerIoWriterSettings struct {
-	Level           string   `cfg:"level" default:"info"`
-	Channels        []string `cfg:"channels"`
-	Formatter       string   `cfg:"formatter" default:"console"`
-	TimestampFormat string   `cfg:"timestamp_format" default:"15:04:05.000"`
-	Writer          string   `cfg:"writer" default:"stdout"`
+	Level           string                    `cfg:"level" default:"info"`
+	Channels        map[string]ChannelSetting `cfg:"channels"`
+	Formatter       string                    `cfg:"formatter" default:"console"`
+	TimestampFormat string                    `cfg:"timestamp_format" default:"15:04:05.000"`
+	Writer          string                    `cfg:"writer" default:"stdout"`
+}
+
+type ChannelSetting struct {
+	Level string `cfg:"level"`
 }
 
 func handlerIoWriterFactory(config cfg.Config, name string) (Handler, error) {
@@ -43,18 +47,23 @@ func handlerIoWriterFactory(config cfg.Config, name string) (Handler, error) {
 		return nil, fmt.Errorf("io writer formatter of type %s not available", settings.Formatter)
 	}
 
-	return NewHandlerIoWriter(settings.Level, settings.Channels, formatter, settings.TimestampFormat, writer), nil
+	channels := make(Channels, len(settings.Channels))
+	for channel, level := range settings.Channels {
+		channels[channel] = LevelPriority(level.Level)
+	}
+
+	return NewHandlerIoWriter(settings.Level, channels, formatter, settings.TimestampFormat, writer), nil
 }
 
 type handlerIoWriter struct {
 	level           int
-	channels        []string
+	channels        Channels
 	formatter       Formatter
 	timestampFormat string
 	writer          io.Writer
 }
 
-func NewHandlerIoWriter(level string, channels []string, formatter Formatter, timestampFormat string, writer io.Writer) *handlerIoWriter {
+func NewHandlerIoWriter(level string, channels Channels, formatter Formatter, timestampFormat string, writer io.Writer) Handler {
 	return &handlerIoWriter{
 		level:           LevelPriority(level),
 		channels:        channels,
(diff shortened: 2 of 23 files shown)


docs/docs/reference/src/log/main.go
```diff
@@ -16,7 +16,7 @@ func main() {
 		// the min log level to write (trace, debug, info, warn, error)
 		log.LevelDebug,
 		// a list of channels to filter for, if empty nothing is filtered. []string{"http"} would write logs from http channel only
```
ajscholl, line 18: The comment is now outdated, can you please update it?

pkg/log/handler_iowriter.go
```diff
@@ -13,11 +13,17 @@ func init() {
...
+	Channels        []Channel `cfg:"channels"`
+	Formatter       string    `cfg:"formatter" default:"console"`
+	TimestampFormat string    `cfg:"timestamp_format" default:"15:04:05.000"`
+	Writer          string    `cfg:"writer" default:"stdout"`
+}
+
+type Channel struct {
+	Name     string `cfg:"name"`
+	Level    string `cfg:"level" default:"info"`
+	Disabled bool   `cfg:"disabled" default:"false"`
```
ajscholl, line 26: Hmm, this completely changes the way the channel filter works. Previously one would simply specify the channels to filter for, so not specifying any channel would filter it out. Now we always log all channels unless a channel is explicitly disabled, so there is no way to filter out all channels unless one (as there are no regex or glob patterns one could use).

And if we go with a flag, I would prefer an enabled flag instead I think

pkg/log/logger.go
```diff
@@ -194,32 +187,14 @@ func (l *gosoLogger) log(level int, msg string, args []interface{}, loggedErr er
...
-		if !strings.HasPrefix(channel, "!") {
-			continue
-		}
-
-		channel = strings.TrimPrefix(channel, "!")
-
-		if channel == current {
-			return false
+func (l *gosoLogger) shouldLog(current string, level int, h Handler) bool {
+	for _, channel := range h.Channels() {
```
ajscholl, line 191: I guess some kind of map to look up the channel name would be useful here, otherwise you have to search all channels every time you log something. I think that would also work well directly in the config like this:
```yml
channels:
    metrics:
        level: info
        enabled: true
    kernel:
        level: warning
```

j4k4:
Hi!
First of all, thank you for your contribution! Very much appreciated! It's a valid use case and it would be great to have this functionality included.
We had meanwhile some internal discussions about the implementation itself and would like to propose a different approach:

We were thinking about adding a new log level `none` which is basically a max int and would disable the logging completely for the whole handler (if set on handler level) or for a specific channel. This would make the `enabled` or `disabled` flag obsolete. After setting `none` on handler level, one could enable logging for specific channels by setting a dedicated log level. The other way around one could set `info` on handler level and disable specific channels by setting the level `none`? Configuration could be done via map:
```yaml
log:
  handlers:
    main:
      type: iowriter
      level: debug
      channels:
        metrics:
          level: error
      formatter: json
      timestamp_format: 2006-01-02T15:04:05Z07:00
      writer: stdout
```

Would this meet your requirements and would you be up to change your implementation? :slightly_smiling_face:

yicooo:
> Hi! First of all, thank you for your contribution! Very much appreciated! It's a valid use case and it would be great to have this functionality included. We had meanwhile some internal discussions about the implementation itself and would like to propose a different approach:
> 
> We were thinking about adding a new log level `none` which is basically a max int and would disable the logging completely for the whole handler (if set on handler level) or for a specific channel. This would make the `enabled` or `disabled` flag obsolete. After setting `none` on handler level, one could enable logging for specific channels by setting a dedicated log level. The other way around one could set `info` on handler level and disable specific channels by setting the level `none`? Configuration could be done via map:
> 
> ```yaml
> log:
>   handlers:
>     main:
>       type: iowriter
>       level: debug
>       channels:
>         metrics:
>           level: error
>       formatter: json
>       timestamp_format: 2006-01-02T15:04:05Z07:00
>       writer: stdout
> ```
> 
> Would this meet your requirements and would you be up to change your implementation? 🙂

Hey @j4k4 @ajscholl , thanks for the comment/review and sorry for the late reply.

I am fine with the changes you proposed. I am changing the implementation accordingly. Initially we had the same conversations and I decided to go with enabled/disabled flag since it is more verbose and frankly none is not a level; and list is simply faster compared to maps when there are few elements.

pkg/log/logger.go
```diff
@@ -14,12 +14,16 @@ const (
...
+	LevelNone  = "none"
 
 	PriorityTrace = 0
 	PriorityDebug = 1
 	PriorityInfo  = 2
 	PriorityWarn  = 3
 	PriorityError = 4
+	// PriorityNone is used to indicate that no logging should be done.
+	// Value is set to the maximum int value to ensure that it is always greater than any other priority.
+	PriorityNone = int(^uint(0) >> 1)
```
netsi, line 26: You could use `math.MaxInt`

yicooo: done

pkg/log/handler_iowriter.go
```diff
@@ -14,12 +14,14 @@ func init() {
...
 type HandlerIoWriterSettings struct {
 	Level           string   `cfg:"level" default:"info"`
-	Channels        []string `cfg:"channels"`
+	Channels        Channels `cfg:"channels"`
 	Formatter       string   `cfg:"formatter" default:"console"`
 	TimestampFormat string   `cfg:"timestamp_format" default:"15:04:05.000"`
 	Writer          string   `cfg:"writer" default:"stdout"`
 }
 
+type Channels map[string]int
```
ajscholl, line 23: Do I read this correctly that we require a number to be specified in the config.yml with this? Am I missing something which turns `level: info` into `2`?

yicooo: Done

yicooo:
@j4k4 @ajscholl could someone run the workflows please?

yicooo:
> @j4k4 @ajscholl could someone run the workflows please?

I need this again since I squashed the commits locally.

j4k4:
Nice notes, thank you.
Merged and released in v0.42.0.

PR #1232, opened 2025-04-11
httpserver: graceful server shutdown

j4k4:
The current httpserver gets closed once the kernel gets stopped which can result in connection issues if there are still open connections or if the app still gets new connections.  

This release enables more graceful shutdown behaviour for the httpserver by implementing 2 additional timeouts:
```yaml
httpserver:
  default:
    timeout:
      drain: 30s
      shutdown: 30s
```
On kernel stop, the httpserver module will get unhealthy but won't close the server until the configured *drain* timeout runs out. In this time frame the orchestrator can adjust the traffic flow, so that the app won't get any new connections. After draining, the http server shuts down and will wait for at most the configured *shutdown* timeout to handle all remaining open connections.

### DIFF
--- a/pkg/httpserver/server.go
+++ b/pkg/httpserver/server.go
@@ -9,12 +9,14 @@ import (
 	"slices"
 	"strconv"
 	"strings"
+	"sync/atomic"
 	"time"
 
 	"github.com/gin-contrib/location"
 	"github.com/gin-gonic/gin"
 	"github.com/justtrackio/gosoline/pkg/appctx"
 	"github.com/justtrackio/gosoline/pkg/cfg"
+	"github.com/justtrackio/gosoline/pkg/clock"
 	"github.com/justtrackio/gosoline/pkg/kernel"
 	"github.com/justtrackio/gosoline/pkg/log"
 	"github.com/justtrackio/gosoline/pkg/tracing"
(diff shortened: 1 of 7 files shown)


pkg/httpserver/server.go
```diff
@@ -155,51 +158,57 @@ func NewWithInterfaces(logger log.Logger, router *gin.Engine, tracer tracing.Ins
...
-func (a *HttpServer) GetPort() (*int, error) {
-	if a == nil {
+func (s *HttpServer) GetPort() (*int, error) {
+	if s == nil {
 		return nil, errors.New("apiServer is nil, module is not yet booted")
 	}
 
-	if a.listener == nil {
+	if s.listener == nil {
 		return nil, errors.New("could not get port. module is not yet booted")
```
ajscholl, line 208: feels also old

```suggestion
		return nil, errors.New("could not get port. module is not yet running")
```

pkg/httpserver/server.go
```diff
@@ -155,51 +158,57 @@ func NewWithInterfaces(logger log.Logger, router *gin.Engine, tracer tracing.Ins
...
+	if err := s.server.Shutdown(shutdownCtx); err != nil {
+		s.logger.Error("Server Close: %w", err)
+	}
 }
 
-func (a *HttpServer) GetPort() (*int, error) {
-	if a == nil {
+func (s *HttpServer) GetPort() (*int, error) {
+	if s == nil {
 		return nil, errors.New("apiServer is nil, module is not yet booted")
```
ajscholl, line 204: this log message looks really old, not only the mention of apiServer, but also the module being not yet booted

```suggestion
		return nil, errors.New("httpServer is nil, module is not yet running")
```

bt-justtrack:
We should also use graceful shutdown for the health_check kernel module, it is also just closed atm.

bt-justtrack (approved):

PR #1235, opened 2025-04-24
funk: added more methods to the Set datatype

ajscholl:


pkg/funk/set.go
```diff
@@ -3,17 +3,107 @@ package funk
...
+		result = true
+	}
+
+	return result
+}
+
+// AddSet adds all elements of a Set to the Set.
+//
+// It returns true if any item was not contained in this Set.
+func (s Set[T]) AddSet(items Set[T]) bool {
```
bt-justtrack, line 34: I would suggest Union as the usual name for this in set theory.

bt-justtrack (approved):

PR #1237, opened 2025-04-29
Consumer improvements

ajscholl:
This MR adds healthchecks to consumers and changes the way the kinsumer writes its metrics.

### DIFF
--- a/pkg/cloud/aws/kinesis/shard_reader.go
+++ b/pkg/cloud/aws/kinesis/shard_reader.go
@@ -46,9 +46,12 @@ type shardReader struct {
 	metricWriter       metric.Writer
 	metadataRepository MetadataRepository
 	kinesisClient      Client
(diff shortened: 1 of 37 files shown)


pkg/cloud/aws/kinesis/shard_reader.go
```diff
@@ -532,6 +581,7 @@ func getShardReaderDefaultMetrics(stream Stream) metric.Data {
 			},
 			Unit:  metric.UnitCount,
 			Value: 0.0,
+			Kind:  metric.KindTotal,
```
bt-justtrack, line 584: In addition to having the 3 total metrics here i would suggest that we also write defaults for the shardid metrics, else with prometheus we would not have any metrics existing as the kindtotal metrics get filtered by the prometheus writer.
``` go
		{
			Priority:   metric.PriorityHigh,
			MetricName: metricNameReadCount,
			Dimensions: metric.Dimensions{
				"StreamName": string(stream),
				"ShardId":   metric.DimensionDefault,
			},
			Value: 0.0,
			Unit:  metric.UnitCount,
		},
```

bt-justtrack (changes requested):

bt-justtrack (approved):

PR #1244, opened 2025-06-16
kafka: fix possible endless loop when shutting down on empty batch

erdac:


pkg/stream/input_kafka.go
```diff
@@ -46,13 +46,13 @@ func (i *KafkaInput) Run(ctx context.Context) error {
 
 	for {
```
ajscholl, line 47: I think I would simplify this to:

```go
	defer close(i.data)
	
	for msg := range i.consumer.Data() {
			i.data <- KafkaToGosoMessage(msg)
	}
	
	return nil // or ctx.Err() if you want
```

pkg/kafka/consumer/consumer.go
```diff
@@ -97,11 +98,20 @@ func (c *Consumer) run(ctx context.Context) error {
...
-				return ctx.Err()
+		batch := c.manager.Batch(ctx)
+
+		select {
+		case <-ctx.Done():
+			for _, msg := range batch {
+				c.backlog <- msg
+			}
+
+			close(c.backlog)
```
ajscholl, line 109: I would just defer this... and maybe restructure the method a bit:

```go
    defer close(c.backlog)

	for {
		select {
		case <-ctx.Done():
			return ctx.Err()
		default:
		}

		batch := c.manager.Batch(ctx)
		for _, msg := range batch {
			c.backlog <- msg
		}
	}
```

ajscholl (approved):

PR #1254, opened 2025-07-07
test/suite: make fixtures available in subscriber test cases and added options when asserting subscriber results

ajscholl:
The subscriber test cases were created before initializing the app running them in the past. This makes it impossible to access fixtures from a call like `WithFixtureSetFactory` as the factory hasn't been run at that point. We now first run the fixture set factories before constructing the test case to allow for this.

Additionally, we now support test cases which assert that a specific model was not not persisted (for example, if you don't want to store all events). You can now pass `suite.WithExpectFound` or `suite.WithExpectNotFound` to the fetcher passed to the `Assert` method as optional arguments.

### DIFF
--- a/pkg/test/suite/testcase_subscriber.go
+++ b/pkg/test/suite/testcase_subscriber.go
@@ -24,14 +24,14 @@ func init() {
 type SubscriberTestCase interface {
 	GetName() string
 	GetModelId() mdl.ModelId
-	GetInput() interface{}
+	GetInput() any
 	GetVersion() int
 }
 
 type subscriberTestCase struct {
 	Name    string
 	ModelId mdl.ModelId
-	Input   interface{}
+	Input   any
 	Output  mdlsub.Model
 	Version int
 }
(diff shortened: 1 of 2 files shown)


pkg/test/suite/testcase_subscriber.go
```diff
@@ -360,9 +413,21 @@ type KvstoreSubscriberFetcher struct {
...
+		return
+	}
+
+	assert.Truef(f.t, ok, "model for subscription %s and key %v should be available in the store", f.name, key)
+
+	if opts.expectFound {
+		assert.Truef(f.t, ok, "unexpected missing item with key %v when fetching kvstore subscription %s", key, f.name)
+	} else {
+		assert.Falsef(f.t, ok, "unexpected found item with key %v when fetching kvstore subscription %s", key, f.name)
+	}
```
bt-justtrack, line 432: asserting True on ok and then asserting False does not work

bt-justtrack (changes requested):

PR #1255, opened 2025-07-08
feat: add traffic redistribution middleware;

bt-justtrack:


pkg/httpserver/settings.go
```diff
@@ -0,0 +1,77 @@
...
+		// Shutdown timeout is the maximum amount of time to wait for serving active requests before stopping the server
+		Shutdown time.Duration `cfg:"shutdown" default:"60s" validate:"min=1000000000"`
+	}
+
+	// TrafficDistributorSettings configures the traffic distributor middleware controlling maximum life of client connections.
+	TrafficDistributorSettings struct {
+		Enabled                   bool          `cfg:"enabled"                   default:"false"`
+		MaxConnectionAge          time.Duration `cfg:"maxConnectionAge"          default:"1m"`
+		MaxConnectionRequestCount int           `cfg:"maxConnectionRequestCount" default:"0"`
+	}
```
ajscholl, line 76: I don't see them in our settings struct, which seems wrong. I am also not sure about the defaults - why isn't it enabled by default? I would also give them a different name, maybe ConnectionLimitSettings? Next, please use snake_case for setting names. Finally, I would shorten the names to maximum_age or max_age and maximum_requests or max_requests, as connection_limits already establish the context of a connection

pkg/httpserver/traffic_redistribution_middleware.go
```diff
@@ -0,0 +1,152 @@
...
+	"github.com/justtrackio/gosoline/pkg/appctx"
+	"github.com/justtrackio/gosoline/pkg/cache"
+	"github.com/justtrackio/gosoline/pkg/cfg"
+	"github.com/justtrackio/gosoline/pkg/clock"
+	"github.com/justtrackio/gosoline/pkg/log"
+)
+
+const (
+	distributorMaxSize   int64  = 10000
+	distributorPruneSize uint32 = 1000
```
ajscholl, line 18: You could leave out the types here I guess

pkg/httpserver/traffic_redistribution_middleware.go
```diff
@@ -0,0 +1,152 @@
...
+)
+
+//go:generate mockery --name TrafficDistributor
+type (
+	TrafficDistributor interface {
```
ajscholl, line 23: I dislike the name because it doesn't distribute traffic, it closes connections, so it is more a connection manager (but manager itself is an empty word) or connection limiter or connection closer than a traffic distributor

pkg/httpserver/traffic_redistribution_middleware.go
```diff
@@ -0,0 +1,152 @@
...
+		// ShouldCloseConnection checks whether the connection to the remote address should be closed (after handling any currently in flight requests).
```
ajscholl, line 24:
```suggestion
		// ShouldCloseConnection checks whether the connection to the remote address should be closed (after handling the next request).
```

pkg/httpserver/traffic_redistribution_middleware.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+
+	trafficDistributionKey string
+)
+
+// ProvideTrafficDistributor returns a TrafficDistributor.
+// This is a component that tracks the lifecycle of client connections
+// based on configurable policies. On that basis it provides recommendations on whether a connection
+// should be closed or not. This is useful to prevent clients from keeping connections open, eg in k8s
+// where load balancing only happens on new connections.
```
ajscholl, line 48:
```suggestion
// ProvideTrafficDistributor returns a TrafficDistributor.
// This is a component that tracks the lifecycle of client connections based on configurable policies. 
// On that basis it provides recommendations on whether a connection should be closed or not. 
// This is useful to prevent clients from keeping connections open for an extended period of time, 
// for example in kubernetes where load balancing only happens on new connections.
```

pkg/httpserver/traffic_redistribution_middleware.go
```diff
@@ -0,0 +1,152 @@
...
+func ProvideTrafficDistributor(ctx context.Context, config cfg.Config, _ log.Logger) (TrafficDistributor, error) {
+	return appctx.Provide(ctx, trafficDistributionKey("TrafficDistributor"), func() (TrafficDistributor, error) {
+		return NewTrafficDistributor(config)
+	})
+}
+
+func NewTrafficDistributor(config cfg.Config) (TrafficDistributor, error) {
+	settings := &TrafficDistributorSettings{}
+	config.UnmarshalKey("trafficDistributor", settings)
```
ajscholl, line 57: So all HTTP servers share the same settings? Feels wrong

ajscholl (changes requested):

PR #1272, opened 2025-07-23
feat: Load ipread db directly through memory

bt-justtrack:


pkg/ipread/provider_maxmind.go
```diff
@@ -109,46 +106,47 @@ func (p *maxmindProvider) Refresh(ctx context.Context) (err error) {
...
 
 	p.lck.Lock()
+	readerToClose := p.reader
 	p.reader = ipreader
+	if readerToClose != nil {
+		if err := readerToClose.Close(); err != nil {
+			return fmt.Errorf("can not close existing reader: %w", err)
+		}
+	}
 	p.lck.Unlock()
```
ajscholl, line 131: This must be deferred, or you never unlock this in case of an error

ajscholl (approved):

applike-ss:
That will be 120Mi for each instance of an application on top of the regular memory usage (~60-80Mi). Does it really make sense to have it in-memory? I'd suggest to at least have an option to have it on-disk and use reader implementations (1 for on-disk, 1 for in-memory) if you really want to have the option to have it in-memory.

PR #1298, opened 2025-08-27
currency: extend currency handling by using fx rates api;

irinaJT:


pkg/currency/provider_fxratesapi.go
```diff
@@ -0,0 +1,109 @@
...
+		logger.Info("using fxratesapi as currency provider")
+
+		return &fxRatesApiProvider{logger, http, &settings}, nil
+	}
+}
+
+type fxRatesApiProvider struct {
+	logger   log.Logger
+	http     http.Client
+	settings *ProviderSettings
```
ajscholl, line 38: Why are the settings a pointer? I would make them not a pointer so we know they are never nil

pkg/currency/provider_fxratesapi.go
```diff
@@ -0,0 +1,109 @@
...
+		rates = append(rates, Rate{Currency: currency, Rate: rate})
+	}
+
+	return rates, nil
+}
+
+func (f *fxRatesApiProvider) FetchHistoricalRates(ctx context.Context, dates []time.Time) (map[time.Time][]Rate, error) {
+	result := make(map[time.Time][]Rate)
+	for _, d := range dates {
+		url := ExchangeRateFxRatesUrl + "historical?access_key=" + f.settings.ApiKey + "&base=EUR&date=" + d.Format("2006-01-02")
```
ajscholl, line 75: We see `d.Format("2006-01-02")` so many times, maybe we should just put that in a variable?

pkg/currency/provider_fxratesapi.go
```diff
@@ -0,0 +1,109 @@
...
+
+		request := f.http.NewRequest().WithUrl(url)
+
+		response, err := f.http.Get(ctx, request)
+		if err != nil {
+			f.logger.Warn("error requesting fxratesapi historical rates for %s: %v", d.Format("2006-01-02"), err)
+
+			continue
```
ajscholl, line 83: I think I would turn all these log and continue calls into an error return... something is not right and we should dig into that, not silently ignore it

pkg/currency/providers.go
```diff
@@ -0,0 +1,37 @@
...
+
+//go:generate go run github.com/vektra/mockery/v2 --name Provider
+type Provider interface {
+	Name() string
+	GetPriority() int
+	FetchLatestRates(ctx context.Context) ([]Rate, error)
+	FetchHistoricalRates(ctx context.Context, dates []time.Time) (map[time.Time][]Rate, error)
+}
+
+var ProviderRegistry = map[string]ProviderOptions{
```
ajscholl, line 25: This should not be exported

pkg/currency/providers.go
```diff
@@ -0,0 +1,37 @@
...
+	ECBProviderName:        ECBProviderOption(),
```
ajscholl, line 26: I would not export the functions creating these providers, it is not required. I would also remove the call and directly pass the function (then they can also be slightly simplified) and just call them `newWhateverProvider`

pkg/currency/providers.go
```diff
@@ -0,0 +1,37 @@
...
+	FxRatesApiProviderName: FxRatesApiProviderOption(),
+}
+
+type ProviderOptions func(ctx context.Context, logger log.Logger, http http.Client, settings ProviderSettings) (Provider, error)
```
ajscholl, line 30: I would call this `ProviderFactory`

pkg/currency/providers.go
```diff
@@ -0,0 +1,37 @@
...
+
+// RegisterProvider registers a new currency provider with the given name and options.
```
ajscholl, line 32: What options? You are passing a factory, not options to the provider

pkg/currency/providers.go
```diff
@@ -0,0 +1,37 @@
...
+// If a provider with the same name already exists, it will be overwritten.
+// Providers should be registered before currency module is initialized (e.g. in init function).
+func RegisterProvider(name string, option ProviderOptions) {
+	ProviderRegistry[name] = option
```
ajscholl, line 36: I think we should panic if something gets overwritten, most of these functions follow that logic

pkg/currency/updater.go
```diff
@@ -2,43 +2,55 @@ package currency
...
 const (
 	ExchangeRateRefresh           = 8 * time.Hour
-	ExchangeRateUrl               = "https://www.ecb.europa.eu/stats/eurofxref/eurofxref-daily.xml"
-	HistoricalExchangeRateUrl     = "https://www.ecb.europa.eu/stats/eurofxref/eurofxref-hist.xml"
+	ExchangeRateECBUrl            = "https://www.ecb.europa.eu/stats/eurofxref/eurofxref-daily.xml"
+	HistoricalExchangeRateECBUrl  = "https://www.ecb.europa.eu/stats/eurofxref/eurofxref-hist.xml"
 	ExchangeRateDateKey           = "currency_exchange_last_refresh"
 	HistoricalExchangeRateDateKey = "currency_exchange_historical_last_refresh"
+	ExchangeRateFxRatesUrl        = "https://api.fxratesapi.com/"
 )
```
ajscholl, line 25: I would move the provider specific stuff into their files

pkg/currency/updater.go
```diff
@@ -2,43 +2,55 @@ package currency
...
 
-const YMDLayout = "2006-01-02"
+const (
+	YMDLayout = "2006-01-02"
```
ajscholl, line 28: Apparently that format is provided by `time.DateOnly`

pkg/currency/updater.go
```diff
@@ -59,16 +71,27 @@ func NewUpdater(ctx context.Context, config cfg.Config, logger log.Logger) (Upda
 		return nil, fmt.Errorf("failed to unmarshal currency updater settings: %w", err)
 	}
 
-	return NewUpdaterWithInterfaces(logger, store, httpClient, clock.Provider, settings), nil
+	providersConfig := make(ProvidersConfig)
+	if err := config.UnmarshalKey("currency_service.providers", &providersConfig); err != nil {
```
ajscholl, line 75: This doesn't work if I configure my provider completely via env variables. You have to loop over all registered providers and try to get their config manually

pkg/currency/updater.go
```diff
@@ -143,20 +167,24 @@ func (s *updaterService) needsRefresh(ctx context.Context) bool {
 }
 
 func (s *updaterService) getCurrencyRates(ctx context.Context) ([]Rate, error) {
-	request := s.http.NewRequest().WithUrl(ExchangeRateUrl)
+	allRates := make([]Rate, 0, RatesAmount)
+	rateMap := make(map[string]float64)
```
ajscholl, line 171: As you are not using the values from this map, you could replace this with `funk.Set[string]` for a nicer API

PR #1314, opened 2025-09-30
httpserver: made validation error log warning instead of error;

### LINKED ISSUE #1313: avoid error logging for httpserver validation errors
Since validation are usually triggered by faulty user input, we don't want validation errors to write error metrics, as this can lead to overly trigger-happy alarms. This PR makes the validation errors log a warning instead of a full error.

nikplx:
Closes #1313

### DIFF
--- a/pkg/funk/map.go
+++ b/pkg/funk/map.go
@@ -1,5 +1,12 @@
 package funk
 
+import (
+	"cmp"
+	"iter"
+	"maps"
+	"slices"
+)
+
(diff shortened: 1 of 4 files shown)


nikplx:
> Do we need to make this configurable, as there are arguments for both approaches? For an internal http server I might want to have this as an error? 🤔

I don't really know if we need to configure this actually. I mean if you need it to write the error metric you can just return a normal error instead of an validation error. Imo validation errors should be used when the user gives us malformed input and I don't really think this constitutes an "application-error"

PR #1318, opened 2025-10-07
metric: ignore parent ctx cancellations on shutdown

nikplx:
There seems to be a race condition in the shutdown process of the metrics package:
- In shutdown the application context gets cancelled
- The metrics writeout works with the bare context
- It gets cancelled too
- Sometimes it can push out some metrics, sometimes it dies too soon

We should fix this by giving it its own context with a time based cancellation, so that it has at least some amount of time to push out all the metrics.


pkg/metric/writer_cw.go
```diff
@@ -98,11 +116,32 @@ func (w *cloudwatchWriter) WriteOne(ctx context.Context, data *Datum) {
...
 }
 
-func (w *cloudwatchWriter) Write(ctx context.Context, batch Data) {
+func (w *cloudwatchWriter) Write(applicationCtx context.Context, batch Data) {
 	if len(batch) == 0 {
 		return
 	}
 
+	timeout := w.writeTimeout
+	if timeout <= 0 {
```
ajscholl, line 125: If you have such logic, put it in the constructor, not the actual code. If we even keep it at all

pkg/metric/writer_cw.go
```diff
@@ -98,11 +116,32 @@ func (w *cloudwatchWriter) WriteOne(ctx context.Context, data *Datum) {
...
+		timeout = defaultWriteTimeout
+	}
+
+	manualCtx, cancel := exec.WithDelayedCancelContext(applicationCtx, timeout)
+
+	cfn, manualCtx := coffin.WithContext(manualCtx)
```
ajscholl, line 131: What is the use of the coffin here? It will never fail before the write method returns or the delayed context is cancelled, so why not use that directly?

pkg/metric/writer_cw.go
```diff
@@ -98,11 +116,32 @@ func (w *cloudwatchWriter) WriteOne(ctx context.Context, data *Datum) {
...
 	if len(batch) == 0 {
 		return
 	}
 
+	timeout := w.writeTimeout
+	if timeout <= 0 {
+		timeout = defaultWriteTimeout
+	}
+
+	manualCtx, cancel := exec.WithDelayedCancelContext(applicationCtx, timeout)
```
ajscholl, line 129: There is nothing manual about the context, it should be called `delayedCtx` instead

pkg/metric/writer_cw.go
ajscholl, line 172: Wouldn't we want to return a (multi) error from this method now instead of always nil?

pkg/metric/writer_cw.go
```diff
@@ -31,12 +33,15 @@ const (
...
+	defaultWriteTimeout = 10 * time.Second
 )
 
 type (
 	CloudWatchSettings struct {
-		Naming    CloudwatchNamingSettings `cfg:"naming"`
-		Aggregate bool                     `cfg:"aggregate" default:"true"`
+		Naming       CloudwatchNamingSettings `cfg:"naming"`
+		Aggregate    bool                     `cfg:"aggregate" default:"true"`
+		WriteTimeout time.Duration            `cfg:"write_timeout" default:"10s"`
```
ajscholl, line 44: It is not a timeout, if a write takes 1 minute but the application is not shutting down, we will still do it, even if the timeout is set to 10s. It is a grace time.

pkg/metric/writer_cw.go
```diff
@@ -31,12 +33,15 @@ const (
 	chunkSizeCloudWatch = 20
 	minusOneWeek        = -1 * 7 * 24 * time.Hour
 	plusOneHour         = 1 * time.Hour
+
+	defaultWriteTimeout = 10 * time.Second
```
ajscholl, line 37: I am not sure I like the double default, we should be able to rely on the config parser setting the right defaults

PR #1340, opened 2025-11-28
Feat batch fixture loading

bt-justtrack:
Adds batched loading of fixtures


bt-justtrack:
closing as this has limited impact (only applicable for types with non complex gorm relationship tags) and adding quite some complexity

PR #1341, opened 2025-11-28
Fix mysql test foreign key

bt-justtrack:
The intermittent MySQL foreign key constraint errors were caused by a metadata caching issue in MySQL/InnoDB when using TRUNCATE.

**Root Cause**

Gosoline's LifeCyclePurger uses TRUNCATE TABLE with FOREIGN_KEY_CHECKS=0 on a separate connection. In InnoDB, TRUNCATE is implemented as DROP+CREATE, which changes the internal table_id. Other connections (like the handler's connection pool) may have cached FK constraint metadata referencing the old table_id, causing FK validation to fail intermittently even when the referenced rows exist.

**Solution**

Disabled gosoline's default TRUNCATE-based purge for the database by adding to config.test.yml:
resource_lifecycles:
  purge:
    enabled: true
    excludes:
      - "^db/.*"

Implemented a DELETE-based purge in table_flush.go that:
- Queries all tables in the database
- Deletes all rows using DELETE instead of TRUNCATE
- Uses FK_CHECKS=0 within a single transaction to handle circular references

**Why This Works**

DELETE doesn't change the table structure or internal table_id, so FK constraint metadata remains valid across all connections. The trade-off is that DELETE is slower than TRUNCATE for large tables, but for test fixtures this is acceptable.

### DIFF
--- a/pkg/test/env/container_runner.go
+++ b/pkg/test/env/container_runner.go
@@ -43,6 +43,7 @@ func (r ContainerRequest) id() string {
 type ContainerConfig struct {
 	RunnerType   string
 	Hostname     string
+	ExternalHost string // Host address for external containers (used by external runner)
 	Auth         authSettings
 	Repository   string
 	Tmpfs        []TmpfsSettings
--- a/pkg/test/env/container_runner_external.go
+++ b/pkg/test/env/container_runner_external.go
@@ -1,0 +1,63 @@
+package env
+
+import (
+	"context"
+	"strconv"
+
+	"github.com/justtrackio/gosoline/pkg/cfg"
+	"github.com/justtrackio/gosoline/pkg/log"
+)
+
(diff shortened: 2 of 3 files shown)


pkg/db/lifecycle_purger.go
```diff
@@ -68,63 +57,80 @@ func (p LifeCyclePurger) Purge(ctx context.Context) (err error) {
...
 
-	if err = rows.Close(); err != nil {
-		return fmt.Errorf("could not close rows: %w", err)
+	if err := p.deleteTables(ctx, tables); err != nil {
+		return err
 	}
 
-	if err = rows.Err(); err != nil {
-		return fmt.Errorf("could not iterate over rows: %w", err)
+	return nil
```
erdac:
```suggestion
	return p.deleteTables(ctx, tables)
```

PR #1349, opened 2026-01-13
smpl: new sampling package;

j4k4:
This release adds advanced sampling capabilities and fingers-crossed logging to the Gosoline framework.
- **New sampling package (`pkg/smpl`)** with pluggable strategies.
- **Fingers-crossed logging (`pkg/log`)** to buffer logs until a trigger.
- **Integration glue** between sampling decisions and log emission.

### Sampling (`pkg/smpl`)
- Introduced a new sampling package centered around a `Decider`.
- Added strategies:
  - **tracing** to align with distributed tracing systems.
  - **Probabilistic sampling**.
  - **HTTP header-based sampling**.

### Logging + fingers-crossed (`pkg/log`)
- Added **fingers-crossed handler** which buffers log records and only flushes them when a condition is met (e.g. an error-level event).
- Added sampling-aware logging integration to reduce log volume while keeping “interesting” traces.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -28,7 +28,7 @@ Operate in this repository as the maintainer of the **gosoline** application fra
    - `golangci-lint run --build-tags integration,fixtures ./...`
 6. Check for missing godoc parts for exported types/functions you added or changed.
 7. Update AGENTS.md files if your changes affect package structure, APIs, or workflows documented there. Check both the root `AGENTS.md` and any package-specific `AGENTS.md` (e.g., `pkg/<package>/AGENTS.md`).
-7. Summarize work with requirement coverage, commands executed, and pending follow-ups. Never stage or commit; CI and reviewers expect clean diffs only.
+8. Summarize work with requirement coverage, commands executed, and pending follow-ups. Never stage or commit; CI and reviewers expect clean diffs only.
 
 ## GitHub MCP server workflow
 - **Repository:** `justtrackio/gosoline`. Pass owner `justtrackio` and name `gosoline` to GitHub MCP tools.
--- a/.github/workflows/docusaurus-deploy.yml
+++ b/.github/workflows/docusaurus-deploy.yml
@@ -12,34 +12,88 @@ defaults:
 
 jobs:
   deploy:
-    environment:
-      name: github-pages
-      url: ${{ steps.deployment.outputs.page_url }}
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+
       - name: Set up Node.js
         uses: actions/setup-node@v4
         with:
-          node-version: 18
-          cache: npm
-          cache-dependency-path: '**/package-lock.json'
+          node-version: 20
+          cache: yarn
+          cache-dependency-path: '**/yarn.lock'
+
       - name: Install dependencies
-        run: npm ci
+        run: yarn install --frozen-lockfile
+
       - name: Build
-        run: npm run build
-      - name: Setup Pages
-        uses: actions/configure-pages@v5
-      - name: Upload artifact
-        uses: actions/upload-pages-artifact@v3
+        run: yarn build
+
+      - name: Deploy to gh-pages branch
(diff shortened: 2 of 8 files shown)


pkg/log/logger.go
```diff
@@ -199,7 +200,11 @@ func (l *gosoLogger) log(ctx context.Context, level int, msg string, args []any,
 	}
 }
 
-func (l *gosoLogger) shouldLog(current string, level int, h Handler) (bool, error) {
+func (l *gosoLogger) shouldLog(ctx context.Context, current string, level int, h Handler) (bool, error) {
+	if smplctx.IsSampled(ctx) {
```
ajscholl: If we sample something, then we are using this sample from the whole. So I think the condition here (and maybe the rest of the implementation) is backwards

docs/docs/how-to/sampling/src/use-sampling/main.go
```diff
@@ -0,0 +1,58 @@
...
+)
+
+func main() {
+	ctx := context.Background()
+	config := cfg.New()
+	logger := log.NewLogger()
+
+	// Register a custom sampling strategy that can be referenced in config.
+	//
+	// This is process-global and should be done during application startup, before a decider is created.
```
ajscholl: Should we maybe move it to an `init` block and declare that as the right place?

pkg/smpl/strategies.go
```diff
@@ -0,0 +1,45 @@
...
+
+	return true, trace.Sampled, nil
+}
+
+// DecideByAlways is a strategy that always decides to sample.
+func DecideByAlways(ctx context.Context) (isApplied bool, isSampled bool, err error) {
+	return true, true, nil
+}
+
+// DecideByNever is a strategy that always decides not to sample.
```
ajscholl: might be clearer to read that way:

```suggestion
// DecideByNever is a strategy that never decides to sample.
```

AGENTS.md
```diff
@@ -26,7 +26,8 @@ Operate in this repository as the maintainer of the **gosoline** application fra
    - `go build -tags fixtures,integration ./...`
    - `go test -tags fixtures,integration ./...`
    - `golangci-lint run --build-tags integration,fixtures ./...`
-6. Update AGENTS.md files if your changes affect package structure, APIs, or workflows documented there. Check both the root `AGENTS.md` and any package-specific `AGENTS.md` (e.g., `pkg/<package>/AGENTS.md`).
+6. Check for missing godoc parts for exported types/functions you added or changed.
+7. Update AGENTS.md files if your changes affect package structure, APIs, or workflows documented there. Check both the root `AGENTS.md` and any package-specific `AGENTS.md` (e.g., `pkg/<package>/AGENTS.md`).
 7. Summarize work with requirement coverage, commands executed, and pending follow-ups. Never stage or commit; CI and reviewers expect clean diffs only.
```
ajscholl:
```suggestion
8. Summarize work with requirement coverage, commands executed, and pending follow-ups. Never stage or commit; CI and reviewers expect clean diffs only.
```

docs/docs/how-to/sampling/src/use-sampling/main.go
j4k4: yeah, changed it and made clear, that it should be done before app start, not during..

pkg/smpl/decider.go
```diff
@@ -0,0 +1,136 @@
...
+// NewDecider creates a new Decider based on the provided configuration.
+// It parses the settings from the "sampling" configuration key.
+func NewDecider(ctx context.Context, config cfg.Config) (Decider, error) {
+	settings := &Settings{}
+	if err := config.UnmarshalKey("sampling", settings); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal settings: %w", err)
+	}
+
+	var ok bool
+	var factory StragegyFactory
```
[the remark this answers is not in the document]
j4k4: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/justtrackio/gosoline/pull/1349#discussion_r2690532611)

PR #1389, opened 2026-04-17
httpserver/auth: remove token value from Google auth cache-miss log entry

applike-ss:
The informational log message emitted when a Google ID token is not in the cache no longer includes the token value as a structured field. The message itself is preserved.


applike-ss:
The logging change in this PR is fully contained within #1394 which rewrites the same function. Closing to avoid a merge conflict — the improvement lands with #1394.

PR #1392, opened 2026-04-17
cfg: mask sensitive values in DebugConfig log output

applike-ss:
DebugConfig now redacts the logged value for any config key whose name contains password, secret, token, key, dsn, credential, or access_key, replacing it with ***. The config fingerprint is still computed from the original values so that it accurately reflects all configuration changes.

### DIFF
--- a/pkg/cfg/debug.go
+++ b/pkg/cfg/debug.go
@@ -11,6 +11,21 @@ import (
 	"github.com/justtrackio/gosoline/pkg/funk"
 )
 
+var debugSensitivePatterns = []string{
+	"password", "secret", "token", "key", "dsn", "credential",
+}
+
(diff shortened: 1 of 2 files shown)


pkg/cfg/debug.go
```diff
@@ -11,6 +11,21 @@ import (
 	"github.com/justtrackio/gosoline/pkg/funk"
 )
 
+var debugSensitivePatterns = []string{
+	"password", "secret", "token", "key", "dsn", "credential", "access_key",
```
irinaJT: does it make sense to have here "access_key" if we already have "key"? should be covered for both cases.

applike-ss: yap, fixed. Thanks!

irinaJT (approved):

PR #1398, opened 2026-04-17
httpserver: return generic message for 5xx responses

applike-ss:
The default JSON error handler now returns the static string "internal server error" for 5xx status codes instead of the raw error message, which may contain internal implementation details. The health-check endpoint similarly returns "unhealthy" for all failed module checks.


pkg/httpserver/health_check.go
irinaJT: shouldn't this be %w?

applike-ss: yap, fixed. Thanks!

irinaJT (approved):

PR #1400, opened 2026-04-21
cloud: add sagemakerruntime package;

irinaJT:


pkg/cloud/aws/sagemakerruntime/naming.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+func (s EndpointNameSettings) GetClientName() string {
+	return s.ClientName
+}
+
+type EndpointNamingSettings struct {
+	EndpointPattern    string `cfg:"endpoint_pattern,nodecode" default:"{app.env}-{group}-{name}"`
+	EndpointDelimiter  string `cfg:"endpoint_delimiter" default:"-"`
+	OrganizationalUnit string `cfg:"organizational_unit" default:""`
```
erdac: it doesn't look like the organizational unit is actually used?

pkg/cloud/aws/sagemakerruntime/naming.go
```diff
@@ -0,0 +1,63 @@
...
+	Group              string `cfg:"group"`
```
erdac: Does this actually need to be here?

You are basically making a group required for the endpoint name, which feels too restrictive to me since we don't do that for the naming of any other service and the endpoint could be anything .

IMO, unless there is some reason why this needs to be required the endpoint pattern should be `{app.namespace}-{name}` like for all the other services and then we can set the group as an optional tag if we need it. We do the same for kafka and kinesis for example.

pkg/cloud/aws/AGENTS.md
```diff
@@ -54,6 +55,8 @@ AWS services (SQS, SNS, Kinesis) use `cfg.Identity.Format()` with pattern-based
 
 **Note:** DynamoDB table naming uses `ModelId` (from `pkg/ddb`), not `cfg.Identity.Format()` directly.
 
+**Note:** SageMaker Runtime endpoint naming differs from other services: only `{app.env}` comes from the calling service's identity. The `{group}` and `{name}` placeholders are configured separately under `cloud.aws.sagemakerruntime.clients.<name>.naming.group` and `.name`, since the endpoint belongs to a different service and cannot be derived from the caller's metadata.
```
erdac: I don't really see a reason why this should be different from the other services. See comment above.

pkg/cloud/aws/sagemakerruntime/naming.go
irinaJT: it is needed if we want to use it in a custom pattern. Actually I can set it as a default for pattern

erdac (approved):

pkg/cloud/aws/sagemakerruntime/client.go
```diff
@@ -0,0 +1,116 @@
...
+)
+
+//go:generate go run github.com/vektra/mockery/v2 --name Client
+type Client interface {
+	InvokeEndpoint(ctx context.Context, body []byte) ([]byte, error)
+}
+
+type ClientSettings struct {
+	gosoAws.ClientSettings
+	Identity    cfg.Identity `cfg:"identity"`
```
erdac: I think the correct way to do this is to use cfg.ResourceIdentifier instead and then use settings.ToIdentity() late like we do here for example: `pkg/stream/output_kinesis.go:22`

PR #1404, opened 2026-05-22
httpserver: implemented concurrent connection and request limits;

ajscholl:
This commit adds these settings:

httpserver.<name>.concurrency.max_requests (default: 0 = disabled)
httpserver.<name>.concurrency.max_connections (default: 0 = disabled)
httpserver.<name>.concurrency.overload_status_code (default: 503)
httpserver.<name>.concurrency.retry_after (default: 0 = header omitted)

If you configure a connection limit, we will not accept new connections until less than the limit of connections are active. Idle connections are closed to make room for new connections if the limit is reached.

If you configure a request limit, additional request will be immediately be rejected with status 503 (unless a different status is configured) and a Retry-After header is sent if configured.


pkg/http/client_test.go
```diff
@@ -76,6 +77,37 @@ func getClient(t *testing.T, retries int, timeout time.Duration) http.Client {
 	return client
 }
 
+func getRetryAfterClient(t *testing.T, retries int, timeout time.Duration, defaultWaitTime time.Duration) http.Client {
```
selm0, line 80: Do we need a test for the WithRequestTimeout option?

pkg/http/retry_after.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+func newRetryAfterFunc(settings Settings) resty.RetryAfterFunc {
+	return func(_ *resty.Client, response *resty.Response) (time.Duration, error) {
+		if response == nil || response.Request == nil {
+			return 0, nil
+		}
+
+		retryAfterHeader := strings.TrimSpace(response.Header().Get(httpHeaders.RetryAfter))
+		if retryAfterHeader == "" && response.StatusCode() != stdHttp.StatusServiceUnavailable {
```
selm0, line 31: So in case of a HTTP 200 we're returning a retry duration?

pkg/http/retry_after.go
```diff
@@ -0,0 +1,124 @@
...
+	if !ok {
+		return nil
+	}
+
+	remaining := time.Until(deadline)
+	if retryAfter <= remaining {
+		return nil
+	}
+
+	return fmt.Errorf("retry wait duration %s exceeds remaining request time %s", retryAfter, remaining)
```
selm0, line 116: Is it preferred to error out here? Assuming we already had some retries and we're suddenly at 5mins retry where the deadline is only 1min. Won't this log a lot of errors?
On the other hand, I'm not sure if it's better to then set to the remaining instead, that can also be cut short, i.e. if you have two consecutive requests

pkg/httpserver/AGENTS.md
```diff
@@ -40,6 +40,10 @@ httpserver.default.mode: release  # or debug
 httpserver.default.timeout.read: 60s
 httpserver.default.timeout.write: 60s
 httpserver.default.compression.level: default
+httpserver.default.concurrency.max_requests: 0       # disabled; reject excess active requests when > 0
+httpserver.default.concurrency.max_connections: 0    # disabled; evict idle keep-alive connections when > 0
+httpserver.default.concurrency.overload_status_code: 503
```
selm0, line 45: Should this be 429? Or is the technical overload more important, as it's not a rate-limit per se

pkg/httpserver/connection_limit_listener.go
```diff
@@ -0,0 +1,116 @@
...
+		release: l.release,
+	}, nil
+}
+
+func (l *connectionLimitListener) acquire() error {
+	select {
+	case l.limit <- struct{}{}:
+		return nil
+	case <-l.closed:
+		return net.ErrClosed
```
selm0, line 72: Do we need to consider the application context here, and in case of shutdown reject any incoming connections?

pkg/httpserver/connection_limit_test.go
```diff
@@ -0,0 +1,185 @@
...
+	"github.com/justtrackio/gosoline/pkg/httpserver"
+	httpserverMocks "github.com/justtrackio/gosoline/pkg/httpserver/mocks"
+	logMocks "github.com/justtrackio/gosoline/pkg/log/mocks"
+	"github.com/justtrackio/gosoline/pkg/test/matcher"
+	tracingMocks "github.com/justtrackio/gosoline/pkg/tracing/mocks"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestConnectionLimit_DelaysAcceptUntilConnectionSlotIsReleased(t *testing.T) {
```
selm0: Does it help to test the connectionLimitListener directly with mocks?

pkg/httpserver/connection_pressure_manager.go
```diff
@@ -0,0 +1,147 @@
...
+	idle        *list.List
+	idleNotify  chan struct{}
+}
+
+type connectionState struct {
+	idleSince time.Time
+	idleEntry *list.Element
+}
+
+func newConnectionPressureTracker(ctx context.Context, metrics ServerMetricRecorder) *connectionPressureTracker {
```
selm0: Do you want to align the names? tracker <-> manager

pkg/httpserver/connection_pressure_manager.go
```diff
@@ -0,0 +1,147 @@
...
+
+	switch state {
+	case http.StateNew:
+		m.markActive(conn)
+	case http.StateActive:
+		m.markActive(conn)
+	case http.StateIdle:
+		m.markIdle(conn)
+	case http.StateHijacked, http.StateClosed:
+		m.remove(conn)
```
selm0, line 68: Do we need to handle default here?

pkg/httpserver/connection_pressure_manager.go
```diff
@@ -0,0 +1,147 @@
...
+func (m *connectionPressureTracker) CloseOldestIdleConnection() (bool, error) {
+	m.mu.Lock()
+	oldestIdleConn := m.popOldestIdleConnection()
+	m.mu.Unlock()
+
+	if oldestIdleConn == nil {
+		return false, nil
+	}
+
+	if err := oldestIdleConn.Close(); err != nil {
```
selm0, line 81: What happens when this errors, is the connection still closed or is it in a zombie state then?

pkg/httpserver/connection_pressure_manager_test.go
```diff
@@ -0,0 +1,110 @@
...
+	"github.com/gin-gonic/gin"
+	"github.com/justtrackio/gosoline/pkg/httpserver"
+	httpserverMocks "github.com/justtrackio/gosoline/pkg/httpserver/mocks"
+	logMocks "github.com/justtrackio/gosoline/pkg/log/mocks"
+	"github.com/justtrackio/gosoline/pkg/test/matcher"
+	tracingMocks "github.com/justtrackio/gosoline/pkg/tracing/mocks"
+	"github.com/stretchr/testify/require"
+)
+
+func TestConnectionPressureTracker_TracksConnectionOpenedAndClosed(t *testing.T) {
```
selm0: Also here, is the test easier if we can directly instantiate with mocks? The recorder approach is a bit bigger module tested than just the server

pkg/httpserver/middleware_concurrency_test.go
```diff
@@ -0,0 +1,107 @@
...
+	close(release)
+
+	assert.Equal(t, http.StatusNoContent, <-result)
+	assert.Equal(t, http.StatusNoContent, <-result)
+}
+
+func TestConcurrentRequestLimitMiddleware_RejectsWhenLimitReached(t *testing.T) {
+	entered := make(chan struct{})
+	release := make(chan struct{})
+	router := newConcurrentRequestLimitRouter(httpserver.ConcurrencySettings{
```
selm0, line 64: s.a. I also noticed that there are no direct integration tests, maybe it's worth having both the unit tests and the recorder tests?

pkg/http/retry_after.go
ajscholl: Only if you need a retry for a 200. You need to add this manually by calling `httpClient.AddRetryCondition`. If your function requires a retry on a HTTP 200, then yes.

pkg/httpserver/AGENTS.md
ajscholl: 429 would require the client to send more than one request to be viable. But if you are unlucky, you can get a 503 on the first request to an overloaded server, so 503 seems correct to me in this case.

pkg/httpserver/connection_pressure_manager.go
ajscholl: Not 100% sure. But I think it is generally unwise to touch the connection in any way after we tried to close it - if it refers to a file descriptor internally, and that fd got allocated to some other file, closing it again (or doing anything at all with it) might cause some "interesting" things for that other file using the same fd.

pkg/http/retry_after.go
ajscholl: If the server says it is back in 5 min, then we should not try to send another request before that. If the caller then has to deal with an error, that is something I would prefer instead of keeping the server overloaded. Yes, we might see some errors in the logs, but the main aim should be to get the server restored to capacity automatically without any developer intervention and sending more requests to the server seems to counteract that

pkg/httpserver/connection_limit_listener.go
ajscholl: In case of a shutdown we should close the socket and thus stop accepting new connections. They will then be handled like before during a shutdown, I don't think we need to explicitly handle this here.

pkg/httpserver/connection_pressure_manager.go
ajscholl: yes

pkg/httpserver/connection_pressure_manager.go
ajscholl: It handles all possible cases from net/http. How would you handle the default?

pkg/httpserver/middleware_concurrency_test.go
ajscholl: fixed

pkg/httpserver/connection_pressure_manager_test.go
ajscholl: refactored

pkg/http/client_test.go
ajscholl: added

selm0 (approved):

PR #1405, opened 2026-06-03
feat(kafka): add Consumer/Producer abstraction with metrics and lifecycle managers

applike-ss:
## Summary

Introduce higher-level `kafka.Consumer` and `kafka.Producer` abstractions in `pkg/kafka` that mirror the kinesis Kinsumer/RecordWriter pattern. The stream package now uses these abstractions instead of the kafka package directly.

### New abstractions (`pkg/kafka/consumer` and `pkg/kafka/producer`)

- **`kafka.Consumer`** — encapsulates polling, rebalancing, health checks, executor retries, and partition management (logic moved from `pkg/stream/input_kafka.go`)
- **`kafka.Producer`** — wraps the existing Writer with metrics tracking

### Lifecycle managers (metadata registration)

Both register lifecycle managers exposing metadata:
- `kafka.consumers`: `{bootstrap_servers, name, topic}`
- `kafka.producers`: `{bootstrap_servers, name, topic}`

### Metrics (standard Kafka terminology)

| Consumer | Producer |
|----------|----------|
| RecordsConsumed | RecordsSent |
| RecordsConsumedFailed | RecordsSentFailed |
| PollCount | ProduceBatchSize |
| PollDuration | |

### Stream package refactoring

- `stream/input_kafka.go` — thin wrapper around `kafka.Consumer` (~80 lines, down from ~230)
- `stream/output_kafka.go` — thin wrapper around `kafka.Producer`

### Removed

- `pkg/kafka/lifecycle_manager.go` — replaced by consumer/producer-specific lifecycle managers

## Metadata server output

```json
{
  "kafka": {
    "consumers": [{"bootstrap_servers": ["127.0.0.1:9092"], "name": "default", "topic": "foo"}],
    "producers": [{"bootstrap_servers": ["127.0.0.1:9092"], "name": "default", "topic": "bar"}]
  },
  "stream": {
    "consumers": [{"name": "default"}],
    "producers": [{"name": "default"}]
  }
}
```

## Testing

- All unit tests pass (`go test ./pkg/kafka/... ./pkg/stream/...`)
- Tests use testify/mockery mocks
- Full build passes with `-tags fixtures,integration`
- Zero golangci-lint issues


pkg/kafka/producer/lifecycle.go
```diff
@@ -0,0 +1,74 @@
...
+func NewLifecycleManagerProducer(name string, connectionName string, topic string, brokers []string) reslife.LifeCycleerFactory {
+	return func(ctx context.Context, config cfg.Config, logger log.Logger) (reslife.LifeCycleer, error) {
+		conn, err := connection.ParseSettings(config, connectionName)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse kafka connection settings for connection name %q: %w", connectionName, err)
+		}
+
+		service, err := admin.NewService(ctx, logger, topic, conn.Brokers)
+		if err != nil {
+			return nil, fmt.Errorf("could not create kafka producer lifecycle manager: %w", err)
```
erdac:
```suggestion
			return nil, fmt.Errorf("could not create kafka admin service: %w", err)
```

pkg/kafka/consumer/lifecycle.go
```diff
@@ -0,0 +1,74 @@
...
+func NewLifecycleManagerConsumer(name string, connectionName string, topic string, brokers []string) reslife.LifeCycleerFactory {
+	return func(ctx context.Context, config cfg.Config, logger log.Logger) (reslife.LifeCycleer, error) {
+		conn, err := connection.ParseSettings(config, connectionName)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse kafka connection settings for connection name %q: %w", connectionName, err)
+		}
+
+		service, err := admin.NewService(ctx, logger, topic, conn.Brokers)
+		if err != nil {
+			return nil, fmt.Errorf("could not create kafka consumer lifecycle manager: %w", err)
```
erdac:
```suggestion
			return nil, fmt.Errorf("could not create kafka admin service: %w", err)
```

pkg/kafka/consumer/partition_manager.go
```diff
@@ -86,6 +101,8 @@ func (p *PartitionManager) OnPartitionsLostOrRevoked(ctx context.Context, _ *kgo
 	defer wg.Wait()
 
 	for topic, partitions := range lost {
+		p.metricWriter.WriteOne(ctx, metric.NewMetricDatum(metricNameRebalanceCount, metric.Dimensions{kafka.DimensionConsumer: p.name, kafka.DimensionTopic: topic}, 1.0, metric.UnitCount, metric.PriorityHigh))
```
erdac, line 102: Aren't we counting each rebalance twice this way, once before the rebalance when the partitions are revoked and then again after the rebalance when we get assigned new ones?

Should we be maybe only do this on revoke?

pkg/kafka/consumer/partition_consumer.go
```diff
@@ -47,19 +67,54 @@ func (c PartitionConsumer) Consume(ctx context.Context) error {
...
+			waitStart = c.clock.Now()
 		}
 	}
 }
 
-func (c PartitionConsumer) Stop() {
+func (c *PartitionConsumer) handleWithRecovery(records []*kgo.Record) (failed bool) {
+	defer func() {
+		if err := coffin.ResolveRecovery(recover()); err != nil {
+			c.logger.Error(context.Background(), "panic in message handler for partition %d of topic %s: %w", c.partition, c.topic, err)
```
erdac: why not use the context from `Consume` here?

pkg/kafka/consumer/lifecycle.go
```diff
@@ -0,0 +1,74 @@
...
+	service        *admin.Service
+	name           string
+	connectionName string
+	topic          string
+	brokers        []string
+}
+
+var _ LifecycleManagerConsumer = &lifecycleManagerConsumer{}
+
+func NewLifecycleManagerConsumer(name string, connectionName string, topic string, brokers []string) reslife.LifeCycleerFactory {
```
erdac: I think the consumer can just provide the already parsed connection here instead of connection name and brokers. Then you should have all you need and don't need to parse that here again.

pkg/kafka/producer/lifecycle.go
```diff
@@ -0,0 +1,74 @@
...
+	service        *admin.Service
+	name           string
+	connectionName string
+	topic          string
+	brokers        []string
+}
+
+var _ LifecycleManagerProducer = &lifecycleManagerProducer{}
+
+func NewLifecycleManagerProducer(name string, connectionName string, topic string, brokers []string) reslife.LifeCycleerFactory {
```
erdac: Same here. You can just provide the already parsed connection.

pkg/kafka/producer/lifecycle.go
```diff
@@ -0,0 +1,74 @@
...
+	return func(ctx context.Context, config cfg.Config, logger log.Logger) (reslife.LifeCycleer, error) {
+		conn, err := connection.ParseSettings(config, connectionName)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse kafka connection settings for connection name %q: %w", connectionName, err)
+		}
+
+		service, err := admin.NewService(ctx, logger, topic, conn.Brokers)
```
erdac: It was already like this before but no idea why I added the topic as an argument to the constructor.  
It probably makes more sense to provide that via `service.CreateTopic(ctx, topic)` instead.

pkg/kafka/consumer/partition_manager.go
applike-ss: i've removed it from the partition assigned loop

pkg/kafka/consumer/lifecycle.go
applike-ss: due to the admin service now involved i chose a slightly different approach, but let me know if this is not to your liking

pkg/kafka/producer/lifecycle.go
applike-ss: due to the admin service now involved i chose a slightly different approach, but let me know if this is not to your liking

erdac (approved):

PR #1406, opened 2026-06-03
mdlsub: fix output db causes panic on update;

erdac:
pkg/mdlsub/transformer.go currently defines TypedTransformer.Transform as returning *M, then untypedTransformer.transform dereferences it before persistence. 

OutputDb.Persist then calls gorm.Save with a non-pointer model which causes gorm to panic on an update in the case where no rows are affected.

So, this change ensure that the model we pass to gorm.Save is addressable.

### DIFF
--- a/pkg/mdlsub/output_db.go
+++ b/pkg/mdlsub/output_db.go
@@ -3,6 +3,7 @@ package mdlsub
 import (
 	"context"
 	"fmt"
+	"reflect"
 
 	"github.com/jinzhu/gorm"
 	"github.com/justtrackio/gosoline/pkg/cfg"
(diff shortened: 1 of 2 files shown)


selm0:
> pkg/mdlsub/transformer.go currently defines TypedTransformer.Transform as returning *M, then untypedTransformer.transform dereferences it before persistence.
> 
> OutputDb.Persist then calls gorm.Save with a non-pointer model which causes gorm to panic on an update in the case where no rows are affected.
> 
> So, this change ensure that the model we pass to gorm.Save is addressable.

What's the default behaviour for other subscriber targets? Do they all take Ms, or is *M the default?
Having the output_db handle both pointer and non pointer variants is only the second best choice, as this means that the input is not consistent

erdac:
> > pkg/mdlsub/transformer.go currently defines TypedTransformer.Transform as returning *M, then untypedTransformer.transform dereferences it before persistence.
> > OutputDb.Persist then calls gorm.Save with a non-pointer model which causes gorm to panic on an update in the case where no rows are affected.
> > So, this change ensure that the model we pass to gorm.Save is addressable.
> 
> What's the default behaviour for other subscriber targets? Do they all take Ms, or is *M the default? Having the output_db handle both pointer and non pointer variants is only the second best choice, as this means that the input is not consistent

They all take the dereferenced M.  
The input is only a pointer in the case when someone returns a double pointer in their transformer.

selm0:
> > > pkg/mdlsub/transformer.go currently defines TypedTransformer.Transform as returning *M, then untypedTransformer.transform dereferences it before persistence.
> > > OutputDb.Persist then calls gorm.Save with a non-pointer model which causes gorm to panic on an update in the case where no rows are affected.
> > > So, this change ensure that the model we pass to gorm.Save is addressable.
> > 
> > 
> > What's the default behaviour for other subscriber targets? Do they all take Ms, or is *M the default? Having the output_db handle both pointer and non pointer variants is only the second best choice, as this means that the input is not consistent
> 
> They all take the dereferenced M. The input is only a pointer in the case when someone returns a double pointer in their transformer.

Is there a way to prevent than a double pointer of entering? Then we can always have the non-pointer inside the output_db

erdac:
> > > > pkg/mdlsub/transformer.go currently defines TypedTransformer.Transform as returning *M, then untypedTransformer.transform dereferences it before persistence.
> > > > OutputDb.Persist then calls gorm.Save with a non-pointer model which causes gorm to panic on an update in the case where no rows are affected.
> > > > So, this change ensure that the model we pass to gorm.Save is addressable.
> > > 
> > > 
> > > What's the default behaviour for other subscriber targets? Do they all take Ms, or is *M the default? Having the output_db handle both pointer and non pointer variants is only the second best choice, as this means that the input is not consistent
> > 
> > 
> > They all take the dereferenced M. The input is only a pointer in the case when someone returns a double pointer in their transformer.
> 
> Is there a way to prevent than a double pointer of entering? Then we can always have the non-pointer inside the output_db

but how does that help?
we still need to make that addressable because the issue is that gorm can't handle non-pointers in that one edge case.

selm0 (approved):

PR #1413, opened 2026-06-15
test/suite: isolate per-run configuration

### LINKED ISSUE #1411: test/suite: isolate per-test SuiteConfiguration state
## Summary

`pkg/test/suite` appears to leak mutable suite state between test cases. This can make isolated test environments accidentally reuse app context, app module wrappers, and cached clients from earlier test cases.

We hit this in `terminal-user-events` while running:

```bash
go test -tags integration,fixtures ./test/consumer-s2s/... -run Deduplication
```

A later stream test failed while creating a DynamoDB-backed resource against a LocalStack endpoint from a previous test case:

```text
Post "http://172.17.0.1:<old-port>/": dial tcp 172.17.0.1:<old-port>: connect: connection refused
```

The failure pattern was:

1. The suite used isolated environments, so each test case got its own LocalStack container.
2. TiKV-only tests had a base-test signature, skipped in Redis mode, and were run as Gosoline base application tests.
3. `buildTestCaseBase` wrote `suiteConf.appCtx = environment.Context()` into the shared suite configuration.
4. Later stream tests reused that mutated `appCtx`, including app-context cached AWS/DynamoDB clients from an environment whose LocalStack container had already stopped.

## Issue 1: `buildTestCaseBase` leaks `suiteConf.appCtx`

`pkg/test/suite/testcase_base.go` mutates the shared suite configuration:

```go
suiteConf.appCtx = environment.Context()
```

That mutation persists after the base test has finished. With isolated environments, this can make a later test use a context from a stopped environment.

### Expected behavior

Base tests can use `environment.Context()` for lifecycle registrations, but only for that specific application run.

### Suggested fix

Pass a per-run app context into `RunTestCaseApplication`, clone the suite configuration for the invocation, or restore the previous `suiteConf.appCtx` with `defer`.

## Issue 2: `RunTestCaseApplication` mutates `suiteConf.appModules`

`pkg/test/suite/testcase_application.go` wraps module factories in-place:

```go
for k, factory := range suiteConf.appModules {
    suiteConf.appModules[k] = newEssentialModuleFactory(factory)
}
```

This makes `RunTestCaseApplication` non-idempotent and lets one test case alter how later test cases are built. Repeated application runs can also accumulate wrappers.

### Expected behavior

Application runs should not modify `suiteConf.appModules`.

### Suggested fix

Build a local `map[string]kernel.ModuleFactory` containing wrapped factories and leave `suiteConf.appModules` unchanged.

## Issue 3: base-test app context should be scoped, not shared

The base-test runner uses `environment.Context()` because lifecycle managers registered in `SetupTest` need to be visible to the application lifecycle. That requirement is valid, but using shared mutable config to propagate this context causes cross-test contamination.

### Expected behavior

Base-test app context should be scoped to the base test application run only.

### Suggested fix

Introduce an explicit per-run option or internal helper for `RunTestCaseApplication` that accepts an app context. The base-test runner can pass `environment.Context()` through that scoped path without mutating global suite configuration.

## Issue 4: missing regression coverage for suite isolation

There should be tests proving that suite configuration remains isolated between test cases.

Suggested coverage:

- A skipped/no-op base test does not affect a later stream or application test.
- Isolated-environment suites do not reuse app context or app-context cached resources from stopped environments.
- Multiple `RunTestCaseApplication` invocations do not mutate or repeatedly wrap `suiteConf.appModules`.
- Base tests still have lifecycle registrations visible when using `environment.Context()`.

## Why this matters

The current behavior is surprising because a skipped test with base-test signature can change the runtime context of later tests. This is especially hard to diagnose with components like LocalStack, Redis, or DynamoDB, where stale app-context cached clients keep pointing at stopped containers.

Making `SuiteConfiguration` immutable per test run, or copying/restoring mutable fields around each invocation, would keep isolated environments truly isolated while preserving the base-test lifecycle behavior.

j4k4:
## Summary
- clone SuiteConfiguration for each application run so per-test options cannot leak
- scope base-test app context to the base application run while preserving lifecycle registrations
- avoid mutating httpserver/subscriber/base runner modules and app options globally
- add regression coverage for suite state isolation

Closes #1411

## Tests
- go test -tags fixtures,integration ./pkg/test/suite

### DIFF
--- a/pkg/test/suite/suite_configuration.go
+++ b/pkg/test/suite/suite_configuration.go
@@ -2,6 +2,7 @@ package suite
 
 import (
 	"context"
+	"maps"
 	"slices"
 
 	"github.com/justtrackio/gosoline/pkg/application"
@@ -25,6 +26,24 @@ type SuiteConfiguration struct {
 
 	testCaseWhitelist   []string
 	testCaseRepeatCount int
+}
+
+func (s *SuiteConfiguration) clone() *SuiteConfiguration {
+	clone := &SuiteConfiguration{
+		envOptions:                       append([]env.Option(nil), s.envOptions...),
+		envSetup:                         append([]func() error(nil), s.envSetup...),
+		envIsShared:                      s.envIsShared,
+		fixtureSetFactories:              append([]fixtures.FixtureSetsFactory(nil), s.fixtureSetFactories...),
+		fixtureSetPostProcessorFactories: append([]fixtures.PostProcessorFactory(nil), s.fixtureSetPostProcessorFactories...),
+		appCtx:                           s.appCtx,
+		appOptions:                       append([]application.Option(nil), s.appOptions...),
+		appModules:                       maps.Clone(s.appModules),
+		appFactories:                     append([]kernel.ModuleMultiFactory(nil), s.appFactories...),
+		testCaseWhitelist:                append([]string(nil), s.testCaseWhitelist...),
+		testCaseRepeatCount:              s.testCaseRepeatCount,
+	}
+
+	return clone
 }
 
 func newSuiteConfiguration(options []Option) *SuiteConfiguration {
--- a/pkg/test/suite/suite_options.go
+++ b/pkg/test/suite/suite_options.go
@@ -1,6 +1,7 @@
 package suite
 
 import (
+	"context"
 	"fmt"
 	"time"
 
@@ -17,6 +18,12 @@ import (
 )
 
 type Option func(s *SuiteConfiguration)
+
+func withAppCtx(ctx context.Context) Option {
+	return func(s *SuiteConfiguration) {
+		s.appCtx = ctx
+	}
+}
 
 func WithClockProvider(clk clock.Clock) Option {
 	return func(s *SuiteConfiguration) {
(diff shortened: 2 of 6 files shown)


pkg/test/suite/suite_configuration.go
```diff
@@ -27,6 +27,28 @@ type SuiteConfiguration struct {
...
+		appOptions:                       append([]application.Option(nil), s.appOptions...),
+		appModules:                       make(map[string]kernel.ModuleFactory, len(s.appModules)),
+		appFactories:                     append([]kernel.ModuleMultiFactory(nil), s.appFactories...),
+		testCaseWhitelist:                append([]string(nil), s.testCaseWhitelist...),
+		testCaseRepeatCount:              s.testCaseRepeatCount,
+	}
+
+	for name, module := range s.appModules {
+		clone.appModules[name] = module
+	}
```
ajscholl:
```suggestion
		appModules:                       maps.Clone(s.appModules),
		appFactories:                     append([]kernel.ModuleMultiFactory(nil), s.appFactories...),
		testCaseWhitelist:                append([]string(nil), s.testCaseWhitelist...),
		testCaseRepeatCount:              s.testCaseRepeatCount,
	}
```

ajscholl (approved):

PR #1422, opened 2026-08-04
kafka: add delayed consumption

j4k4:
## Summary

Adds an optional Kafka consumer `consume_delay` setting that postpones batch handling until the newest record in the batch reaches the configured age. This supports consumers that intentionally process records after a fixed delay. The default remains `0`, preserving existing behavior.

## Implementation

- Add `consume_delay` to `consumer.Settings`, validated as non-negative.
- Delay each partition batch relative to its newest record timestamp; batches already older than the configured delay proceed immediately.
- For consumer groups with a delay enabled, pause fetching the assigned partition while a batch is queued or waiting, then resume it after the batch completes, is canceled, or encounters a commit failure. This prevents more batches from accumulating for that partition during the delay.
- Keep delayed non-grouped/read-only processing healthy by periodically marking the consumer health check healthy; grouped consumers intentionally do not do this while a partition is delayed.
- Split elapsed timing into existing wait, new `SleepDuration`, process, and commit metrics.
- Propagate partition-consumer failures through the owning consumer instead of logging them a second time during shutdown.
- Treat expected generation/rebalance commit errors as graceful shutdown when the partition was already revoked or lost, while retaining normal error reporting for active partitions.

## Tests

Adds focused coverage for:

- disabled, elapsed, canceled, and batch-timestamp-based delays
- health-check behavior during delayed grouped and read-only consumption
- pausing and resuming fetches across completion, revocation, cancellation, and commit failure
- avoiding blocked assignments when a delayed consumer has stopped
- rebalance-time commit errors versus active-partition commit failures
- wait and sleep duration metric emission

## Documentation

Documents `consume_delay` for Kafka stream inputs, including its per-batch behavior.

## Operational note

Enabling delayed consumption for grouped Kafka consumers trades throughput for delayed handling: franz-go discards buffered records for paused partitions and refetches them after resuming, which adds broker round trips and removes per-partition fetch pipelining.


pkg/kafka/consumer/partition_consumer.go
```diff
@@ -62,44 +100,159 @@ func (c *PartitionConsumer) Consume(ctx context.Context) error {
...
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-c.stop:
+			return nil
+		case records := <-c.assignedBatch:
+			if len(records) == 0 {
+				continue
+			}
 
+			keepConsuming, err := c.consumeBatch(ctx, records, waitStart)
```
bt-justtrack, line 142: **High — graceful-drain deadline can be bypassed.** `consumePending` checks `ctx.Done()` only when selecting the next batch, then calls `consumeBatch` without rechecking it. If the delayed shutdown context expires while this batch is queued, `consumeBatch` can still invoke the handler and commit after `ConsumeGraceTime`. Please recheck cancellation immediately before processing and committing each pending batch, including the fast path.

pkg/kafka/consumer/partition_manager.go
```diff
@@ -132,37 +165,110 @@ func (p *PartitionManager) OnPartitionsLostOrRevoked(ctx context.Context, _ *kgo
...
 	p.stopping.Store(true)
 
 	p.lck.Lock()
+	partitionCount := len(p.consumers)
+
+	p.logger.Info(ctx, "draining %d Kafka partitions for consumer %s", partitionCount, p.name)
+
 	for assignment, consumer := range p.consumers {
-		consumer.Stop()
+		consumer.Drain()
```
bt-justtrack, line 257: **High — graceful shutdown can still block indefinitely.** `Drain()` causes the worker to run `consumeBatch`, whose `messageHandler.Handle` is synchronous. `PartitionManager.Stop` then waits for the worker to die, but the handler has no context or deadline (the stream Kafka handler sends to an unbuffered channel), so `ConsumeGraceTime` cannot bound a blocked handler. Please make delivery cancellation-aware or decouple/timeout the drain before waiting.

pkg/kafka/consumer/partition_manager.go
```diff
@@ -132,37 +165,110 @@ func (p *PartitionManager) OnPartitionsLostOrRevoked(ctx context.Context, _ *kgo
...
+		healthTicker = p.clock.NewTicker(healthInterval)
+		healthTickerChan = healthTicker.Chan()
+		defer healthTicker.Stop()
+	}
+
+	for {
+		select {
+		case <-ctx.Done():
+			return false
+		case <-stopped:
```
bt-justtrack, line 238: **High — read-only delayed batches can be dropped on shutdown.** In the read-only path, `handleWithoutCommit` runs synchronously and passes `c.stopped` here. `Consumer.Stop` closes that channel, so a batch currently waiting in `ConsumeDelay` returns false and is discarded; there is no partition worker or offset commit to replay it. Please preserve or drain read-only batches during graceful stop, or explicitly define the at-most-once behavior.

pkg/kafka/consumer/partition_manager.go
```diff
@@ -28,15 +30,28 @@ type PartitionManager struct {
...
+	healthTimeout  time.Duration
 }
 
 type assignment struct {
 	topic     string
 	partition int32
 }
 
-func NewPartitionManager(logger log.Logger, clk clock.Clock, metricWriter metric.Writer, messageHandler KafkaMessageHandler, name string) *PartitionManager {
+func newPartitionManager(
```
bt-justtrack, line 45: **Medium — this removes a previously exported API.** `NewPartitionManager` was public on the base branch and is replaced here by `newPartitionManager`; the same change also removes `NewPartitionConsumer` and `HandleWithoutCommit`. External packages using `pkg/kafka/consumer` will fail to compile even though this repository has no remaining callers. Please keep compatibility wrappers or make the breaking change explicit.

pkg/kafka/logging/logger.go
```diff
@@ -10,7 +10,7 @@ import (
 )
 
 const (
-	KafkaLoggingChannel     = "stream.kafka"
+	KafkaLoggingChannel     = "kafka"
```
bt-justtrack, line 13: **Medium — renaming the channel breaks existing log configuration.** `handlerBase.ChannelLevel` looks up channel settings by exact name, so deployments configuring `stream.kafka` will no longer apply their per-channel level after this change to `kafka`. Please preserve `stream.kafka` as an alias or provide a migration path.

pkg/kafka/consumer/partition_manager.go
```diff
@@ -132,37 +165,110 @@ func (p *PartitionManager) OnPartitionsLostOrRevoked(ctx context.Context, _ *kgo
...
 	}
 	p.lck.Unlock()
 
 	close(p.done)
 
-	if err := p.cfn.Wait(); err != nil {
-		p.logger.Error(ctx, "failed to stop partition consumers: %w", err)
+	// Partition errors are escalated through p.errors and logged by the module owner.
+	// Waiting here must not report the coffin's copy of the same error again.
+	<-p.cfn.Dead()
```
bt-justtrack, line 266: **Medium — commit failures during the graceful drain can be hidden.** A drained worker can return a non-context commit error through `p.errors`, but this code only waits on `p.cfn.Dead()` and never consumes that error. Because `Run` can already have selected `c.stopped` and return nil, shutdown can be reported as successful after handling a batch whose offset was not committed. Please surface drain errors before reporting success.

pkg/kafka/logging/logger.go
j4k4: nothing should brake here..

